PR #84, opened 2014-11-17
Use godep command to find godep dependencies

anguslees:
Using godep "correctly" allows for two alternate methods to find
dependencies:
1. Use the regular $GOPATH project tree like regular go projects
2. Use the godep wrapper and use hermetic/vendored libraries in Godeps/

This change cleans up the import paths and modifies the ./build script
to use option (2).


eyakubovich:
I'm closing this for now as we want vendoring to work consistently between CoreOS projects.

PR #85, opened 2014-11-17
output flannel vars as docker command line options

eparis:
Add two new options:

-docker-fmt=false
-docker-fmt-prefix=DOCKER_NETWORK_OPTIONS

These will switch from

    FLANNEL_SUBNET=1.2.3.4/8
    FLANNEL_MTU=1450

to:

    DOCKER_NETWORK_OPTIONS="--bip=1.2.3.4/8 --mtu=1450"

So the docker unit file can import the env file and use it as docker
command line options.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,8 @@ type CmdLineOpts struct {
 	ipMasq        bool
 	subnetFile    string
 	iface         string
+	dockerFmt     bool
+	dockerFmtPrefix string
 }
 
 var opts CmdLineOpts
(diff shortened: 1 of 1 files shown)


eparis:
ExecStartPost is not necessarily finished before system considers a service 'ready'.

I tried to fix it cleanly using only systemd.  uggh.  I still think this pr is the better solution.

http://pkgs.fedoraproject.org/cgit/flannel.git/commit/?id=8d0e6c275d6eb3c779d4f3f5e29b84dd865796b2

Tell me that's not horrific!  I need a new 2 new unit files and a generator...

eyakubovich:
@eparis I recently retested ExecStartPost (with a long sleep) and it waited for it to complete before marking the service ready. At command line, `systemctl start foo.service` would not return until ExecStartPost finishes. Of course, there's a timeout on the whole service starting but that's pretty long.

eparis:
I just retested and it worked as well. It was lennart who told me they weren't tied together. Liar   :)

eyakubovich:
Replaced with https://github.com/coreos/flannel/issues/93

PR #90, opened 2014-11-26
Add new "host-gw" backend.

anguslees:
Add a new "routed" backend that forwards traffic using regular IP
routes.  Useful on private layer2 networks (or dedicated virtual
networks).


eyakubovich:
@anguslees yeah, something along `subnet` lines. `flat` or `l2flat`? Or maybe emphasize that the host is used the gateway: `host-gw`?

/cc @philips you're good with names.

kelseyhightower:
I'm +1 on host-gw. I think we should roll with that name for now and mark this backend as a tech preview until it matures a bit.

PR #108, opened 2014-12-28
VXLAN static forwarding

titanous:
This removes the complicated monitoring code and replaces it with three network configuration entries for each subnet lease that cause the packets to end up in the right place:

- A route pointing all traffic to the subnet via a gateway that is the IP of the flannel interface on the target host.
- A forwarding database entry that maps the network-routable IP address of the target host to the MAC address of the flannel interface on the same host.
- An ARP table entry mapping the IP address of the flannel interface on the target host to the MAC address of the same interface.


eyakubovich:
@titanous we'd like to merge this commit but need to make sure the stale entries are removed on start. Do you mind if we take these commits, rebase them and add the required functionality? Your commit (with you as an author) will be retained.

titanous:
Go for it. We've been using this in Flynn since I opened the PR with no trouble.

eyakubovich:
Thanks. Great to hear that it works well.

eyakubovich:
Obsoleted by https://github.com/coreos/flannel/pull/191

PR #116, opened 2015-01-23
build: add script to build flannel ACI

alban:
This is the first step for making an .aci from flannel.

Discussion about this happened on:
https://github.com/coreos/rocket/issues/389

### DIFF
--- a/build
+++ b/build
@@ -15,7 +15,8 @@ eval $(go env)
 
 if [ ${GOOS} = "linux" ]; then
 	echo "Building flanneld..."
-	go build -o ${GOBIN}/flanneld ${REPO_PATH}
+	# Static compilation is useful when flannel is run in a container
+        go build --ldflags '-extldflags "-static"' -o ${GOBIN}/flanneld ${REPO_PATH}
 else
 	echo "Not on Linux - skipping flanneld build"
 fi
--- a/build-aci
+++ b/build-aci
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+# A non-installed actool can be used, for example:
+# ACTOOL=../../appc/spec/bin/actool
+ACTOOL=${ACTOOL:-actool}
+
+IMAGEDIR=${IMAGEDIR:-bin/image-aci}
+
+VERSION=$1
+
+if [ ! -x "$ACTOOL" ] ; then
+    echo "actool ($ACTOOL) is not executable"
+    exit 1
+fi
+
+if [ ! -x bin/flanneld ] ; then
+    echo "bin/flanneld not found. Is it compiled?"
+    exit 1
+fi
+
+if [ -z "$VERSION" ] ; then
+    echo "Usage: build-aci VERSION"
+    exit 1
+fi
+
+mkdir -p $IMAGEDIR/rootfs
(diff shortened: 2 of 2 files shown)


build
```diff
@@ -15,7 +15,8 @@ eval $(go env)
 
 if [ ${GOOS} = "linux" ]; then
 	echo "Building flanneld..."
-	go build -o ${GOBIN}/flanneld ${REPO_PATH}
+	# Static compilation is useful when flannel is run in a container
+        go build --ldflags '-extldflags "-static"' -o ${GOBIN}/flanneld ${REPO_PATH}
```
jonboulle: I think your indentation is a little off here

build-aci
```diff
@@ -0,0 +1,86 @@
...
+
+mkdir -p $IMAGEDIR/rootfs
+if [ ! -d $IMAGEDIR/rootfs -o ! -x $IMAGEDIR/rootfs ]; then
+    echo "$IMAGEDIR/rootfs is not a writeable directory"
+    exit 1
+fi
+
+
+cp bin/flanneld  $IMAGEDIR/rootfs/
+cp README.md            $IMAGEDIR/rootfs/
```
jonboulle: I guess you're using tabs ;-)

build-aci
```diff
@@ -0,0 +1,86 @@
...
+mkdir -p $IMAGEDIR/rootfs/dev/net
+
+cat <<DF > $IMAGEDIR/manifest
+{
+    "acVersion": "0.1.1",
```
jonboulle: This should probably be bumped

jonboulle:
lgtm other than those small things, but I'll defer to @eyakubovich

eyakubovich:
@alban do you get warnings with static linkage? I remember doing static compilation but abandoned due to nasty warnings. I think there was also an issue with resolver.

If flannel is started with `--ip-masq`, it calls out to `iptables` to install the rules. Therefore iptables also needs to be bundled. Also, since `iptables` is itself dynamically linked, it was just easier to have flannel dynamically linked.

alban:
@eyakubovich : yes, I get the warnings I mentioned in the commit message:

Maybe the script should take the dynamic libraries and iptables from another base image rather than to ship whatever is on the build machine...

eyakubovich:
@alban For docker, we base the image on flannelbox -- a small distro I built using buildroot. You're right, it would be good to fix the versions of iptables and libraries to known good ones. Even better if we use the same versions for Docker and ACI builds.

jonboulle:
@eyakubovich so what needs to be done here? do we need to codify the build process for flannelbox in this repo?

eyakubovich:
@jonboulle The config file form flannelbox is checked in: https://github.com/coreos/flannel/blob/master/dist/flannelbox.config

I guess we could also create a Dockerfile that will assemble buildroot env and build flannelbox in a container. That will produce a rootfs to which we can drop-in flanneld and then build an ACI out of that. Of course this should just produce the same output as docker2aci so it maybe easier to use that.

It'd be great if we could produce a minimal ACI with just the required bits and also have an ability to use well known versions of dependencies and not the ones on the build machine. But I'm not sure how to do so in an automated fashion except for a long bash script.

eyakubovich:
Closing in favor of https://github.com/coreos/flannel/pull/143

PR #118, opened 2015-01-29
Factor out installing IP masquerading out of backends

### LINKED ISSUE #101: Setup IP Masquerade in VXLAN case
Right now VXLAN does not honor ipMasq parameter. It needs to or that code needs to move out of the backends.

eyakubovich:
This is an attempt to have the same IP masquerading rules
across all backends. They work purely on address ranges and
without interface names.

Practically, VXLAN now has IP masquerade support.

Fixes #101

### DIFF
--- a/backend/alloc/alloc.go
+++ b/backend/alloc/alloc.go
@@ -23,7 +23,7 @@ func New(sm *subnet.SubnetManager) backend.Backend {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


vcaputo:
lgtm, but I'm no iptables guru.  The fact that you felt it was necessary to originally make the masquerading per-backend suggests there's some doubt/murkiness.  /cc @marineam wanna take a look?

PR #121, opened 2015-01-31
Use rename(2) so that subnet file becomes atomically visible

titanous:
Before this patch the subnet file could become visible in a partially written state. Instead of writing to the destination, we write to a temporary file in the same directory and then `rename(2)` the file into place so that it will become visible in a consistent state.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,7 +7,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
-	"path"
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


eyakubovich:
O_TMPFILE is a perfect candidate here but alas, it's too new and it'd be great to be compatible with older kernels.

eyakubovich:
LGTM

PR #131, opened 2015-02-25
flannel: add static configuration and subnet lease support

kelseyhightower:
flannel can be configured with a static configuration and subnet lease.

    flanneld -config '{"Network": "10.0.0.0/8", "SubnetLen": 20, "SubnetMin": "10.10.0.4", "SubnetMax": "10.99.0.0"}' -subnet="10.10.0.0-24"

The result of this change means flannel can start without etcd. Once etcd
is online, flannel will register its lease.

### DIFF
--- a/Documentation/static-configuration.md
+++ b/Documentation/static-configuration.md
@@ -1,0 +1,39 @@
+# Static Configuration Guide
+
+This guide will demonstrate how to run flannel with a static configuration and subnet lease.
+
+Why would you want to do this?
+
+Running both flannel and etcd from docker containers presents a bootstrapping problem if you want etcd to reside within the flannel overlay network.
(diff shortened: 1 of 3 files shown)


eyakubovich:
@kelseyhightower I'm not sure I understand how the whole thing is going to work. This will make it possible for etcd to run over flannel network but how will other nodes join in? They would have the same bootstrapping problem. How would they reach etcd to get their lease and setup a routes/overlay to etcd node?

eyakubovich:
@kelseyhightower I understand how this works in alloc mode + GCE where something else configures the network. But I fail to understand how it can work with overlays? How will other nodes contact etcd over overlay if it's not setup for them yet?

eyakubovich:
This will be handled by reservations soon.

PR #135, opened 2015-03-05
Remove workaround which leaks sockets

lemenkov:
Hello All.
This workaround over the strange issue with 300s timeouts followed by the responses with an empty body leads to a clearly visible socket leackage. For the additional details about the underlying issue see these links:

* https://github.com/golang/go/issues/8946
* https://github.com/golang/go/issues/8648

I see that this (empty body in response) no longer causes flannel to stuck so it should be better removed.

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>

### DIFF
--- a/subnet/registry.go
+++ b/subnet/registry.go
@@ -1,7 +1,6 @@
 package subnet
 
 import (
-	"fmt"
 	"path"
 	"sync"
 	"time"
@@ -104,9 +103,8 @@ func (esr *etcdSubnetRegistry) watchSubnets(since uint64, stop chan bool) (*etcd
 		}
 
 		if len(resp.Body) == 0 {
-			// etcd timed out, go back but recreate the client as the underlying
-			// http transport gets hosed (http://code.google.com/p/go/issues/detail?id=8648)
-			esr.resetClient()
+			// etcd timed out, and the underlying http transport gets hosed
+			// http://code.google.com/p/go/issues/detail?id=8648
 			continue
 		}
 
@@ -118,17 +116,6 @@ func (esr *etcdSubnetRegistry) client() *etcd.Client {
 	esr.mux.Lock()
 	defer esr.mux.Unlock()
 	return esr.cli
-}
-
-func (esr *etcdSubnetRegistry) resetClient() {
-	esr.mux.Lock()
-	defer esr.mux.Unlock()
-
-	var err error
-	esr.cli, err = newEtcdClient(esr.etcdCfg)
-	if err != nil {
-		panic(fmt.Errorf("resetClient: error recreating etcd client: %v", err))
-	}
 }
 
 func ensureExpiration(resp *etcd.Response, ttl uint64) {


eyakubovich:
@lemenkov Does the current version leak sockets? And this patch fixes that by keeping the connections open longer?

Do you know if those golang bugs got fixed?

lemenkov:
@eyakubovich yes, the current version (f3c0734881ce62f0a8eabba57ba7ea6da3f18479) still leaks sockets. It opens ~1 new socket per 5 minutes exacly as requested here:

https://github.com/coreos/flannel/blob/f3c0734881ce62f0a8eabba57ba7ea6da3f18479/subnet/registry.go#L128

There is definitely an issue somewhere within etcd because it should reply faster, but still flannel must handle it properly, without eating all the available sockets.

Regarding golang - not sure if it was fixed completely (I'm using 1.3.3) but it seems that flannel works now.

eyakubovich:
@lemenkov I tested this PR with etcd 2.0 and it seems to work well. But with 0.4.6 I still get those nasty messages after 5 mins. At the same time I was not aware of leaked sockets. Yes, it destroys and re-creates the client but the old socket gets closed. On my machine I'm not seeing leaked sockets. What distro do you see this on? CoreOS?

I also re-ran the golang/go#8946 test cases and they still seem to fail (I'm using Go 1.4.1).

lvlv:
+1 for this PR

I'm also experiencing this issue, my etcd server is v2.0.0-rc1 and my client is built from master. I have flannel running on Ubuntu 14.10 and CentOS 7, all of them leak sockets. Let me know how can I help to investigate this more.
my 2 cents: even if etcd (server) closes the connection with zero length messages (by bug), the client should not leak sockets. Thanks for your further help in advance!

$ netstat -alpn | grep flannel
tcp        0      0 172.16.3.242:49143      172.16.1.42:4001        ESTABLISHED 1308/flannel
tcp        0      0 172.16.3.242:48031      172.16.1.42:4001        ESTABLISHED 1308/flannel
tcp        0      0 172.16.3.242:48552      172.16.1.42:4001        ESTABLISHED 1308/flannel
tcp        0      0 172.16.3.242:47476      172.16.1.42:4001        ESTABLISHED 1308/flannel
tcp        0      0 172.16.3.242:50595      172.16.1.42:4001        ESTABLISHED 1308/flannel
tcp        0      0 172.16.3.242:49920      172.16.1.42:4001        ESTABLISHED 1308/flannel

eyakubovich:
@lvlv I completely agree that we shouldn't leak sockets. It's just tracking it down is hard. AFAIK, etcd started setting better timeouts in etcd v2.0.3 so I would try a newer version first. If possible can you run flanneld under strace: `strace -t -ff -e trace=socket,close bin/flanneld`. This should show how often the sockets get created and closed.

lvlv:
My previous experiments show that the sockets are created exactly every 5
minutes. I'll get more detailed logs from strace if you prefer.

On Sat, Mar 28, 2015 at 2:00 AM, Eugene Yakubovich <notifications@github.com

lvlv:
Attached is my logs after run flannel for 10 minutes, also netstat shows 3
sockets are opened.
(I added some more debug info in logs inside
coreos/go-etcd/etcd/requests.go to see if Body.close is always called)

$ sudo netstat -alpn | grep flannel
tcp        0      0 127.0.0.1:50290         127.0.0.1:4001
 ESTABLISHED 28592/flannel
tcp        0      0 172.16.3.242:51877      172.16.1.42:4001
 ESTABLISHED 11792/flanneld
tcp        0      0 127.0.0.1:50266         127.0.0.1:4001
 ESTABLISHED 28592/flannel
tcp        0      0 172.16.3.242:39578      172.16.1.42:4001
 ESTABLISHED 11792/flanneld
tcp        0      0 127.0.0.1:50274         127.0.0.1:4001
 ESTABLISHED 28592/flannel

lvlv:
I just sent out PR#144 <https://github.com/coreos/flannel/pull/144>, it
demonstrate the place where connections are leaked.

lemenkov:
@lvlv like your approach! It's much better than mine so I'm going to close this PR in favour or yours one.

Everyone please go to #144.

PR #136, opened 2015-03-05
Retry VXLAN interface creation if it fails

linvjw:
Otherwise, the process can exit prematurely if flanneld is started
before the host networking configuration is not yet completed.

### DIFF
--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -79,7 +79,8 @@ func (vb *VXLANBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
 		if err == nil {
 			break
 		} else {
-			log.Warning("Failed to create VXLAN interface, retrying...")
+			log.Error("VXLAN init: ", err)
+			log.Info("Retrying in 1 second...")
 
 			// wait 1 sec before retrying
 			time.Sleep(1*time.Second)
--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"time"
 
 	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
 	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/vishvananda/netlink"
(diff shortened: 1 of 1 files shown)


linvjw:
Eric Paris had objections to using network-online.target.  Apparently its use is discouraged by the systemd folks.

cgwalters:
I think `network-online.target` is better than a polling loop personally.

So for the real fix -the reason the VXLAN device creation is failing is because the specified external interface isn't up, right?  We could watch that via netlink, or alternatively via DBus to NetworkManager/systemd-networkd.

(I have zero patches in flannel, so feel free to ignore me as well)

linvjw:
Honestly, I'd rather use network-online.target as well...

linvjw:
That said, I don't really see the harm in retrying once per second until the devices comes up.

eparis:
It follows the same pattern as what flannel does if etcd is down.  It just keeps looping....

eparis:
Then again, systemd stopped telling people to not use network-online.  so maybe i'm fine with it now....

eyakubovich:
We don't have a way to get notified when etcd is up. Also, etcd can go down and we need to retry to connect back to it. The situation here is similar with the exception that if vxlan device gets hosed sometime into operation, there's nothing to recreate it. So we don't currently don't have the same sort of resiliency built in.

I like this PR but do you know if it's possible to differentiate the errors and only loop on "master device not found" type of error? If not, can you print the error from `newVXLANDevice` in the warning (otherwise real errors will be hard to diagnose).

linvjw:
I'm not sure of all the possible error returns.  I'll research it...

As for printing the error, it would be easy to add something like this:

--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -79,6 +79,7 @@ func (vb *VXLANBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
                if err == nil {
                        break
                } else {
+                      log.Warning(err)
                        log.Warning("Failed to create VXLAN interface, retrying...")
 
                        // wait 1 sec before retrying

I'm open to suggestions on the actual log message wording, of course.  The err value comes from the ensureLink function in backend/vxlan/device.go and already has fairly verbose wording.

linvjw:
Clearly I don't know the magic screen formatting bits...the "log.Warning(err)" part would be the addition...

linvjw:
Huzzah...

```
--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -79,6 +79,7 @@ func (vb *VXLANBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
                if err == nil {
                        break
                } else {
+                       log.Warning(err)
                        log.Warning("Failed to create VXLAN interface, retrying...")
 
                        // wait 1 sec before retrying
```

eyakubovich:
@linvjw Let's change it to an error (I think of a warning of something that you march past but here we're stopped in our tracks). It's not a fatal error but still.

Maybe something like:
```
log.Error("VXLAN init: ", err)  // since like you say, err is already verbose
log.Info("Retrying in 1 second...")
```

linvjw:
Sure, I just pushed a patch with the wording as you suggested.

eyakubovich:
LGTM. Thank you for the contribution.

PR #137, opened 2015-03-06
allow changing the default interface via the environment

AntonioMeireles:
simplifies drop-ins a lot when one needs to set custom interfaces as
there isn't the need anymore to dupe the whole ExecStart thing

closes coreos/bugs#228

hi, please review...

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,12 @@ MTU that it supports.
 -v=0: log level for V logs. Set to 1 to see messages related to data path
 ```
 
+## Environment Variables
+Besides using the `-iface` command line option one can also change the the default interface to use
+(IP or name) for inter-host communication via the the `FLANNELD_INTERFACE` environment variable.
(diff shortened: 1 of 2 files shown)


eyakubovich:
@AntonioMeireles all command like flags are also exposed via env vars with `FLANNELD_` prefix followed by env var name in caps and hyphens changed to underscores: https://github.com/coreos/flannel/blob/master/main.go#L62

However, if you're using CoreOS, it's best to use cloud-config and set the `coreos.flannel.interface` field: see https://coreos.com/docs/cluster-management/setup/cloudinit-cloud-config/#configuration-parameters The docs there are slightly outdated in that the cloud-config now creates an environment file that's fed into `docker run --env-file=...` instead creating vars in drop-in. Unfortunately env vars in drop-in don't propagate into the container.

AntonioMeireles:
oh, ok [walked the code too fast :smile:], got the picture. thanks! closing.

PR #140, opened 2015-03-10
add script to generate Rocket network conf file

eyakubovich:
After flannel has started and written out the subnet env
file, run mk-rkt-net.sh to generate network conf file
in /etc/rkt/net.d. For example:

$ ./mk-rkt-net.sh 10-mynet.conf

will create /etc/rkt/net.d/10-mynet.conf with "mynet" as
the name of the network.

### DIFF
--- a/dist/mk-rkt-net.sh
+++ b/dist/mk-rkt-net.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+usage() {
+	echo "$0 [-f FLANNEL-ENV-FILE] [-n NET-NAME] [-r ROUTES] RKT-NETCONF-FILE
+
+Generate Rocket network file based on flannel env file
+OPTIONS:
+	-f	Path to flannel env file. Defaults to /run/flannel/subnet.env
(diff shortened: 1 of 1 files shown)


philips:
Would it make sense to have a "flannel" plugin type that read this directly? Or just have flannel output this configuration directly?

eyakubovich:
For now I think there's value in having flannel logically separate from Rocket. We do the same thing with Docker and mk-docker-opts.sh. The glue script is ok when it's all automated by the distro via unit files or whatever. It's uglier when the user has to do that manually.

An advantage of keeping flannel and rkt decoupled is that it makes it possible to use flannel+veth or flannel+bridge (and flannel+ovs someday). The rkt plugins would be veth, bridge, ovs. Otherwise there's a combinatorial blow up.

Having said that, I think we need to have a broader discussion on how to reduce the number of networking layers with plugins between flannel, rkt, k8s, etc.

eyakubovich:
Closing in favor of https://github.com/appc/cni/pull/20

PR #149, opened 2015-04-06
updated to try both new and old etcd ports

eyakubovich:


jonboulle:
lgtm assuming it works :-)

On Mon, Apr 6, 2015 at 12:37 PM, Eugene Yakubovich <notifications@github.com

PR #155, opened 2015-04-18
backend: Add Amazon VPC backend

rohansingh:
The Amazon VPC backend adds routes to a VPC route table. Routing is then
handled by Amazon's infrastructure.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -20,6 +20,18 @@
 			"Rev": "d1c4472bf2efd3826f2b5bdcc02d8416798d678c"
 		},
 		{
+			"ImportPath": "github.com/mitchellh/goamz/aws",
+			"Rev": "caaaea8b30ee15616494ee68abd5d8ebbbef05cf"
+		},
+		{
+			"ImportPath": "github.com/mitchellh/goamz/ec2",
+			"Rev": "caaaea8b30ee15616494ee68abd5d8ebbbef05cf"
+		},
+		{
+			"ImportPath": "github.com/vaughan0/go-ini",
+			"Rev": "a98ad7ee00ec53921f08832bc06ecf7fd600e6a1"
+		},
+		{
 			"ImportPath": "github.com/vishvananda/netlink",
 			"Rev": "2187ba67a244c1c32f53bf88876e766bbbbcd5e6"
 		}
--- a/Godeps/_workspace/src/github.com/mitchellh/goamz/aws/attempt.go
+++ b/Godeps/_workspace/src/github.com/mitchellh/goamz/aws/attempt.go
@@ -1,0 +1,74 @@
+package aws
+
+import (
+	"time"
+)
+
+// AttemptStrategy represents a strategy for waiting for an action
+// to complete successfully. This is an internal type used by the
+// implementation of other goamz packages.
+type AttemptStrategy struct {
+	Total time.Duration // total duration of attempt.
+	Delay time.Duration // interval between each try in the burst.
+	Min   int           // minimum number of retries; overrides Total
+}
+
+type Attempt struct {
+	strategy AttemptStrategy
+	last     time.Time
+	end      time.Time
+	force    bool
+	count    int
+}
+
+// Start begins a new sequence of attempts for the given strategy.
+func (s AttemptStrategy) Start() *Attempt {
+	now := time.Now()
+	return &Attempt{
+		strategy: s,
+		last:     now,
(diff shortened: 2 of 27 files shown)


rohansingh:
Apologies for the out-of-the-blue code drop. It's pretty straightforward though, and the usage is outlined in `README.md`. Let me know if you see any issues, or if this isn't something that you think fits with the direction of flannel.

backend/vpc/vpc.go
```diff
@@ -0,0 +1,130 @@
...
+
+type VpcBackend struct {
+	sm     *subnet.SubnetManager
+	rawCfg json.RawMessage
+	cfg    struct {
+		Region       string
+		RouteTableID string
+	}
+	ec2        *ec2.EC2
+	instanceID string
```
eyakubovich: Does it need to be a member? I think it can be local var in Init.

backend/vpc/vpc.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type VpcBackend struct {
+	sm     *subnet.SubnetManager
+	rawCfg json.RawMessage
+	cfg    struct {
+		Region       string
+		RouteTableID string
+	}
+	ec2        *ec2.EC2
```
eyakubovich: Does it need to be a member? I think it can be local var in Init.

rohansingh: Oops. Left over from a draft version.

eyakubovich:
This is a great addition. I made some small nits. Also, let's call the backend `aws-vpc` to make it clear that it's for AWS. We've also planned to add the same thing for GCE, so it will help to disambiguate.

rohansingh:
@eyakubovich Thanks for the review. Just pushed a version with the following changes:

1. Fixed scoping of the vars you pointed out.
2. Renamed to `aws-vpc`.
3. Auto-detect the region from AWS metadata, so you just specify the route table ID in the config.

eyakubovich:
LGTM. Thank you for contributing.

Also, do you know what happens when the traffic is sent in a single VPC subnet? Does it still hop through the router (i.e. extra latency) or is does it stay within the subnet? I understand that this is all virtual so it may all look very different under the hood but it'd be interesting to understand.

benmccann:
Nice!  We were just talking about this over at kubernetes where I had some questions about their AWS docs (https://github.com/GoogleCloudPlatform/kubernetes/issues/7036) and Flannel. @brendanburns said "AWS VPC Advanced networking is also capable of subnet per host". I just wanted to clarify that means all AWS VPCs? And not AWS VPCs with "Enhanced Networking"?

This is pretty cool. If the same thing is added for GCE then Flannel could be a pretty generic way of handling this on any cloud and perhaps integrated as a default component of Kubernetes...

rohansingh:
@eyakubovich Don't have access to my test rig at the moment, but I'd think traffic within the same subnet wouldn't go through the router. The router would just send packets back to the host on which they originated, after all.

@benmccann The AWS [Enhanced Networking](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html) thing is something else altogether. It's not required for this.

What @bredanburns was referring to is exactly what this backend does, except that the Kube setup scripts do it using the AWS CLI:
https://github.com/GoogleCloudPlatform/kubernetes/blob/master/cluster/aws/util.sh#L493

By the way, what that setup script is doing with creating a VPC subnet and automatically choosing & associating the correct route table is more sophisticated than what this backend is doing. It might be something to look into emulating.

benmccann:
Cool. Thanks for the clarifications!

bakins:
Very cool!

Thoughts on using https://github.com/awslabs/aws-sdk-go rather than goamz?  I noticed that Hashicorp has started to use it rather than mitchell's version.  Not 100% sure it supports all the API we need for routes.

benmccann:
@justinsb was switching Kubernetes to use awslabs rather than goamz

rohansingh:
Looks like it's got support: http://godoc.org/github.com/awslabs/aws-sdk-go/service/ec2#EC2.CreateRoute

justinsb:
On switching to the official AWS SDK, I think it is going to happen.  I filed this issue: https://github.com/GoogleCloudPlatform/kubernetes/issues/6162 and @jsdir took up the cause!

Sounds like we're just waiting for a few final things to be supported in the AWS SDK: https://github.com/GoogleCloudPlatform/kubernetes/issues/6162#issuecomment-93910186

PR #159, opened 2015-04-23
cycle through etcd endpoints when connecting

eyakubovich:


philips:
huh, what is this for?

eyakubovich:
@philips if etcd2 is configured to listen just on the new IANA ports, this allows flannel to work without having the user specify --etcd-endpoints. Initially I thought that if you just pass a list of URL's to go-etcd, it'll cycle through them but it doesn't work.

crawford:
Clever. So the etcd client library only attempts to connect to the first address?

eyakubovich:
That was the observed behavior. However, it has some complicated logic that I didn't follow on what to do with multiple endpoints.

crawford:
LGTM

yichengq:
@eyakubovich @crawford @philips 
go-etcd is designed to go through all endpoints in the etcd cluster. There was some bug on the cycle, and it has been fixed in master. I just backport it to 0.4 branch.

Moreover, we have added basic backward compatibility in go-etcd 2.0 recently, and I think we can move to use go-etcd master version in flannel.

eyakubovich:
@yichengq That's great news. I'll give it a try.

PR #163, opened 2015-04-29
fix the bug of lost route information when restart network

windwizard:
hostgw: fix the bug of lost route infomation when restart network

when we're using flannel with hostgw backend, if we unplug the network cable and then plug the network cable immediately, we find the route infomation will be lost and can never be recovered.  and of course we can not be routed to containers in other minor with k8s.  Our fix periodly check the routes information, and will fix it when route information lost.

### DIFF
--- a/backend/hostgw/hostgw.go
+++ b/backend/hostgw/hostgw.go
@@ -87,6 +87,12 @@ func (rb *HostgwBackend) Run() {
 
 	defer rb.wg.Wait()
 
+	rb.wg.Add(1)
+	go func() {
+		rb.sm.ListLeases(evts, rb.stop)
+		rb.wg.Done()
+	}()
+
 	for {
 		select {
 		case evtBatch := <-evts:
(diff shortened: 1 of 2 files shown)


subnet/subnet.go
```diff
@@ -394,6 +396,36 @@ func (sm *SubnetManager) WatchLeases(receiver chan EventBatch, cancel chan bool)
 	}
 }
 
+func (sm *SubnetManager) ListLeases(receiver chan EventBatch, cancel chan bool) {
```
eyakubovich: Instead of adding functionality to periodically list leases, it's better to keep the list in the backend itself. You can then also keep netlink.Route objects instead of Leases so they're ready to be pushed down into the kernel.

backend/hostgw/hostgw.go
```diff
@@ -145,8 +151,62 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+				}
+			}
+
 		default:
 			log.Error("Internal error: unknown event type: ", int(evt.Type))
 		}
 	}
 }
+
+func checkSubnetExistInRouts(lease subnet.SubnetLease) bool {
```
eyakubovich: Typo: `checkSubnetExistInRouts` -> `checkSubnetExistInRoutes`.

backend/hostgw/hostgw.go
```diff
@@ -145,8 +151,62 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+	routeList, err := netlink.RouteList(nil, netlink.FAMILY_V4)
```
eyakubovich: This ends up getting called multiple times every second (as many as there are subnets). It should be called once outside the loop.

backend/hostgw/hostgw.go
```diff
@@ -145,8 +151,62 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+	if err != nil {
+		return true
+	} else {
+		ipNet := lease.Network.ToIPNet()
+		gateWay := lease.Attrs.PublicIP.ToIP()
+		for _, route := range routeList {
+			if route.Dst == nil {
+				continue
+			}
+			if route.Dst.IP.Equal(ipNet.IP) && route.Gw.Equal(gateWay) && bytesEqual(route.Dst.Mask, ipNet.Mask) {
```
eyakubovich: Can you define a helper, `routeEqual`

backend/hostgw/hostgw.go
```diff
@@ -144,9 +162,76 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+	for {
+		select {
+		case <-cancel:
+			return
+		default:
+			rb.checkSubnetExistInRoutes()
+		}
+		time.Sleep(routeCheckRetries * time.Second)
+	}
+}
```
eyakubovich: I think it's best to use `time.After` here:

```
for {
    select {
        case <-cancel:
            return
        case time.After(routeCheckRetries * time.Second):
            rb.checkSubnetExistInRoutes()
    }
}

File: backend/hostgw/hostgw.go [hunk 6]
```diff
@@ -28,6 +29,15 @@ import (
 	"github.com/coreos/flannel/subnet"
 )
 
+const (
+	routeCheckBegins  = 120
+	routeCheckRetries = 10
+)
+
+var (
+	routeList []netlink.Route
```
eyakubovich: Better to avoid global vars and put it into `HostgwBackend`.

backend/hostgw/hostgw.go
```diff
@@ -144,9 +162,76 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+}
+
+func routeEqual(x, y netlink.Route) bool {
+	if x.Dst.IP.Equal(y.Dst.IP) && x.Gw.Equal(y.Gw) && bytesEqual(x.Dst.Mask, y.Dst.Mask) {
+		return true
+	}
+	return false
+}
+
+func bytesEqual(x, y []byte) bool {
```
eyakubovich: There's `bytes.Equal` that does this.

backend/hostgw/hostgw.go
```diff
@@ -144,9 +162,76 @@ func (rb *HostgwBackend) handleSubnetEvents(batch subnet.EventBatch) {
...
+				if r.Dst == nil {
+					continue
+				}
+				if routeEqual(r, route) {
+					exist = true
+					break
+				}
+			}
+			if !exist {
+				netlink.RouteAdd(&route)
```
eyakubovich: Check for errors and log them.

backend/hostgw/hostgw.go
```diff
@@ -28,6 +29,15 @@ import (
 	"github.com/coreos/flannel/subnet"
 )
 
+const (
+	routeCheckBegins  = 120
```
eyakubovich: Having this feels like a race. Is there a way to avoid it?

eyakubovich:
@windwizard Few more comments but it's shaping up.

eyakubovich:
LGTM. Thank you for your contribution.

PR #169, opened 2015-05-08
Multi network

eyakubovich:
This is the first of two PR's to get mutli-network + client/server option added to flannel.
This PR refactors code to abstract subnet manager into an interface that can be implemented by both a "local" and "remote" implementations. It also introduces `network` arg throughout to enable running multiple networks at the same time.


rajatchopra:
Perfunctory review, I will spend some more time later. Looks good I think, though I feel we will have to re-organize the code at some point.

 - Split out subnet package in three parts?
     - api+interface (what is there in subnet.go)
     - implementation of the interface above (what is there in etcd.go)
     - the actual registry that reads/writes/watches etcd (registry.go/watch.go)
 - The documentation needs to be updated with multiple network usage. I believe a follow-up PR can do that.
 - What @mrunalp suggested: we should Godep the dependencies, but compile it by modifying GOPATH. Using Godeps/.. in import path seems unclean.
 - More of an architectural thing:
       - Should flannel just obtain the networks established in the config? Why do we need to supply the networks at start? Soon we will have to deal with on-demand acquire a network's lease kind of thing. 
       - Do we want to keep retrying on errors in network.Init()?

None of the above are a blocker though. Will get back with another round.

eyakubovich:
FWIW, I hate the idea of godep patching source files to point to Godep dir. However, it's what every CoreOS project uses. I'm open to idea of using GOPATH but want to check-in with @philips, @bcwaldon, and @jonboulle as they have experimented with godep a lot more.

eyakubovich:
I did the cmd line arg to pick networks to join as a first cut. Totally agree that it'll need to change. One option is that once we have `flannelctl` to use that to instruct flannel to join or leave networks.

eyakubovich:
Retrying on errors may be a good or a bad idea. Seems like little hard and we had requests/PRs to make that happen at the backends.

rajatchopra:
I think beyond the re-organization of code mentioned above, this looks good to me as first cut.
Definitely a TODO after the client-server PR.

jonboulle:
I have a mild preference to stick with rewrites unless we come up with a compelling alternative to godep we can move to more generally.

backend/alloc/alloc.go
```diff
@@ -28,9 +34,9 @@ func (m *AllocBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Sub
 		PublicIP: ip.FromIP(extIP),
 	}
 
-	sn, err := m.sm.AcquireLease(&attrs, m.stop)
+	l, err := m.sm.AcquireLease(m.ctx, m.network, &attrs)
 	if err != nil {
```
jonboulle: yuck, use a switch statement!

subnet/renew.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
+)
+
+const (
+	renewMargin = time.Hour
+)
+
+func LeaseRenewer(ctx context.Context, m Manager, network string, lease *Lease) {
+	for {
+		dur := lease.Expiration.Sub(time.Now()) - renewMargin
```
jonboulle: this isn't right, should be outside the loop

PR #170, opened 2015-05-09
backend/awsvpc: allow RBAC for instances

grepory:
using getAuth instead of getEnv allows goamz to acquire authentication
material from instance metadata as well as the environment.


eyakubovich:
@grepory Code LGTM. Can you quickly mention this in the README.md.

grepory:
How's that?

eyakubovich:
LGTM. That's great, thank you for the PR.

PR #171, opened 2015-05-13
Use netlink to get interface addresses instead of golang's lib

eyakubovich:
golang incorrectly returns peer addr for non-point-to-point interfaces that have a peer addr defined.


eyakubovich:
Fixes https://github.com/coreos/flannel/issues/168

backend/vxlan/device.go
```diff
@@ -161,7 +161,7 @@ func (dev *vxlanDevice) MonitorMisses(misses chan *netlink.Neigh) {
 	}
 
 	for {
-		msgs, err := nlsock.Recieve()
+		msgs, err := nlsock.Receive()
```
vcaputo: hah

eyakubovich: I noticed that misspelling a long time ago but never submitted a PR to fix it because I didn't want to break the API. But someone else fixed it.

vcaputo:
lgtm

PR #173, opened 2015-05-14
turn on go-etcd logging

eyakubovich:


yichengq:
the code lgtm
AFAIK, the log of go-etcd prints out details of each request. Will it be too many logs in stderr then?

eyakubovich:
@yichengq it's definitely not ideal but i figure it's better than being completely silent when trying to connect.

yichengq:
k. lgtm then.

PR #177, opened 2015-05-15
Client / server

eyakubovich:


remote/client.go
```diff
@@ -0,0 +1,190 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"path"
+
+	//log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
```
bobbyrullo: remove unused code

remote/client.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/coreos/flannel/subnet"
+)
+
+type httpRespErr struct {
+	resp *http.Response
+	err  error
+}
+
+func httpDo(ctx context.Context, req *http.Request) (*http.Response, error) {
+	// Run the HTTP request in a goroutine and pass the response to f.
```
bobbyrullo: What is "f"? Also, why run this in a goroutine? Is this just so that it exits out quickly if the requester cancels?

bobbyrullo:
a nit, and a question, otherwise LGTM.

remote/client.go
eyakubovich: They comment is out of date. Yes, it's in a goroutine to be able to interrupt it (cancel request). Unfortunately that's really the only sane way to do interruption in Go.

remote/client.go
```diff
@@ -0,0 +1,190 @@
...
+	if network == "" {
+		network = "/_"
+	}
+	if network[0] != '/' {
+		network = "/" + network
+	}
+	return m.host + path.Join(append([]string{network}, parts...)...)
+}
+
+func (m *RemoteManager) GetNetworkConfig(ctx context.Context, network string) (*subnet.Config, error) {
```
bakins: the `ctx` isn't actually being used in most of these functions.  Is the thought to use in the future without changing the function signatures?

eyakubovich: RemoteManager implements subnet.Manager interface which has `ctx` as the first arg for consistency. Not all methods/implementations (there're two now) do anything with it yet. New etcd client is ctx based so once we move over to it, it'll be used more.

remote/server.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	c, err := sm.GetNetworkConfig(ctx, network)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
bakins: Need to set content-type header. Perhaps create a helper for the json responses rather than repeating?
```go
func JSON(w http.ResponseWriter, code int, v interface{}) {
	w.Header().Set("Content-Type", "application/json; charset=utf-8")
	w.WriteHeader(code)
        if err:= json.NewEncoder(w).Encode(v); err != nil {
           // log something
       }
}
```

eyakubovich: Great catch!

remote/client.go
bakins: Got it. :+1:

bakins:
LGMT other than the content-type.

rajatchopra:
Ran the setup. Works fine. LGTM.

rajatchopra:
TLS on the server remains as a TODO.

PR #181, opened 2015-05-18
fix handling of failed watches that result in snapshot

eyakubovich:
This also fixes the cursor not being used in some cases.

### DIFF
--- a/subnet/watch.go
+++ b/subnet/watch.go
@@ -21,11 +21,16 @@ import (
 	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
 )
 
+// WatchLeases performs a long term watch of the given network's subnet leases
+// and communicates addition/deletion events on receiver channel. It takes care
+// of handling "fall-behind" logic where the history window has advanced too far
+// and it needs to diff the latest snapshot with its saved state and generate events
 func WatchLeases(ctx context.Context, sm Manager, network string, receiver chan []Event) {
 	lw := &leaseWatcher{}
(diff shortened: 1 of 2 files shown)


subnet/watch.go
```diff
@@ -23,9 +23,10 @@ import (
 
 func WatchLeases(ctx context.Context, sm Manager, network string, receiver chan []Event) {
 	lw := &leaseWatcher{}
+	var cursor interface{}
```
jonboulle: Gross, can't you use -1 instead of an empty interface?

eyakubovich: The idea is that the cursor is opaque to these "helper methods" and at the interface level. Each subnet.Manager impl is free to implement the cursor as it sees fit.

subnet/watch.go
```diff
@@ -21,11 +21,16 @@ import (
 	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
 )
 
+// WatchLeases performs a long term watch of the given network's subnet leases
+// and communicates addition/deletion events on receiver channel. It takes care
+// of handling "fall-behind" logic where the history window has advanced to far
```
jonboulle: too far

subnet/watch.go
```diff
@@ -21,11 +21,16 @@ import (
...
+// and it needs diff the latest snapshot with its saved state and generate events
```
jonboulle: to diff

jonboulle:
lgtm

PR #189, opened 2015-05-27
properly handle nil cursor and use Etcd vs Modified index

eyakubovich:
This fixes two bugs:
1. watch was not getting a snapshot with nil cursor.
2. watch was not advancing properly due to wrong index used.


MohdAhmad:
LGTM

PR #201, opened 2015-06-04
docs: add test and build procedure for centos

greenpau:
docs: add test and build procedure for centos

this procedure guides a user how to build and test
flanneld

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,62 @@ overlay network:
 * Step 2: Git clone the flannel repo: ```https://github.com/coreos/flannel.git```
 * Step 3: Run the build script: ```cd flannel; ./build```
 
+If you are building `flanneld` on CentOS 7, you may need to proceed with the following steps:
+
+1. Install Golang: 
+
+    ```
+    yum install -y git golang golang-vim golang-src
+    ```
+    
+2. If necessary, add proxy server addresses:
+
+    ```
+    export http_proxy="http://proxy.domain.com:9001"
+    export https_proxy="http://proxy.domain.com:9002"
+    ```
+
+3. Create a development directory and clone `flannel` project:
+
+    ```
+    cd ~/dev && git clone https://github.com/coreos/flannel.git && cd flannel
+    ```
+    
+4. Add `GOPATH` environment variable pointing to `flannel`'s workspace:
+
+
+    ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,62 @@ overlay network:
...
+    export https_proxy="http://proxy.domain.com:9002"
+    ```
+
+3. Create a development directory and clone `flannel` project:
+
+    ```
+    cd ~/dev && git clone https://github.com/coreos/flannel.git && cd flannel
+    ```
+    
+4. Add `GOPATH` environment variable pointing to `flannel`'s workspace:
```
eyakubovich: This step should not be necessary. `build` script sets `GOPATH` inside and `test` sources it in.

greenpau: @eyakubovich , I was running `./test` first (before `build`). The `test` required `Cover`, but the `test` failed without `$GOPATH`. Once I had it, I was able to get `Cover` and run `test`.

greenpau: Perhaps it is better to add the following code to `test`:

```
cd `dirname $0` || exit 1;

if [ -z "${GOPATH+x}" ]; then
        export GOPATH=${PWD}/gopath
fi

if [ -z "${GOBIN+x}" ]; then
        export GOBIN=${PWD}/bin
fi

go get golang.org/x/tools/cmd/cover
```

This way ${PWD} points to source code directory.

eyakubovich: It's fine to require the user to download the tooling (compiler, cover, etc) and place it wherever they want and set GOROOT and/or GOBIN as they see fit. But GOPATH we set in the scripts so that vendoring works. I think your original PR was fine, just without step 4.

philips:
hey @greenpau it would be nice to put this into a hacking doc something like we have for rkt: https://github.com/coreos/rkt/blob/master/Documentation/hacking.md

Could you do that and create a CentOS section?

greenpau:
> it would be nice to put this into a hacking doc something like we have for rkt: 

> Could you do that and create a CentOS section?

@philips , in which file do you want the CentOS section? Are you asking me to create `Documentation/hacking.md` in this repo?

I have more materials for the doc. I just need to dig them out.

tomdee:
There is a dockerized build process in master now there should be OS agnostic. Closing rather than leaving a needs/rework PR open for months.

PR #202, opened 2015-06-05
Automatically detect routeTableID

MohdAhmad:


backend/awsvpc/awsvpc.go
```diff
@@ -103,6 +102,58 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
 	}
 	ec2c := ec2.New(auth, region)
 
+	if m.cfg.RouteTableID == "" {
+		log.Infof("routeTableID not passed as config parameter, attempting to detect")
```
eyakubovich: Stylistic nit: lets put the contents of this if block into a separate function.

backend/awsvpc/awsvpc.go
```diff
@@ -103,6 +102,58 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
...
+		metaPath := "network/interfaces/macs/"
+		macs, err := aws.GetMetaData(metaPath)
```
eyakubovich: We already have `instanceID`. Can we use `ec2.Instances([]string{instanceID}, nil)` to get `InstancesResp`. Then it should just be `resp.Reservations[0].Instances[0].SubnetId`. I haven't tried it but I wonder if it'll work and maybe cleaner than going through the mac addresses.

backend/awsvpc/awsvpc.go
```diff
@@ -103,6 +102,34 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
...
+
+	resp, err := ec2c.DescribeRouteTables([]string{m.cfg.RouteTableID}, filter)
+	if err != nil {
+		return nil, fmt.Errorf("error describing route tables: %v", err)
+	}
+
+	for _, routeTable := range resp.RouteTables {
+		for _, route := range routeTable.Routes {
+			if l.Subnet.String() == route.DestinationCidrBlock && route.State == "active" {
+				log.Errorf("matching *active* entry to: %s that will be deleted: %s, %s \n", l.Subnet.String(), route.DestinationCidrBlock, route.GatewayId)
```
eyakubovich: Small nit: Error objects start with lowercase but log entries start with a capital.

backend/awsvpc/awsvpc.go
```diff
@@ -103,6 +102,34 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
...
+		m.cfg.RouteTableID = routeTableID
+	}
+
+	filter := ec2.NewFilter()
+	filter.Add("route.destination-cidr-block", l.Subnet.String())
+	filter.Add("route.state", "active")
+
+	resp, err := ec2c.DescribeRouteTables([]string{m.cfg.RouteTableID}, filter)
+	if err != nil {
+		return nil, fmt.Errorf("error describing route tables: %v", err)
```
eyakubovich: Since we now know that special (wildcard) permissions are required for Describe methods, what do you think about not failing here but continuing with a warning if the error is Unauthorized? Just include in the warning some text on what permission is required.

eyakubovich:
LGTM

PR #207, opened 2015-06-10
Dynamically handles L3 misses with vxlan backend

### LINKED ISSUE #203: Does flannel with vxlan require manual entry for docker container IPs in ARP?
Hi,

I'm running flanneld on 2 hosts which use vxlan device. When I try to ping from hostA docker container to the hostB docker container, it fails. The reason being, missing ARP entry for the hostB container IP or even hostB docker0 interface. I manually add the ARP entry for hostB container IP with the device address of hostB flannel.1 in hostA and the ping worked. Does flannel expect us to do this or should flannel daemon be taking care of ARP entries of docker containers?

arp -n
Address                  HWtype  HWaddress           Flags Mask            Iface
192.168.160.0            ether   42:63:3d:36:49:04   CM                    flannel.1
192.168.208.0            ether   ea:7e:ef:56:77:80   CM                    flannel.1
192.168.208.2            ether   02:42:c0:a8:d0:02   C                     docker0

As you can see from the above, there is only an entry for flannel.1 interface of hostB but not it's container IP which is 192.168.160.2 or it's docker0 interface which is 192.168.160.1

I build flannel locally and here is how I run it:
sudo bin/flanneld -etcd-endpoints="http://localhost:4001" -etcd-prefix="/coreos.com/network" -subnet-file="/run/flannel/subnet.env" -v=2 >flannel.log 2>&1 &

Please let me know if I'm missing something here. 

Thanks!

MohdAhmad:
Since L2 misses are still handled statically DOVE extensions are
no longer required.routes.go maintains the mappings. devices.go now monitors for L3 misses and vxlan is
updated to handleL3Misses.

Fixes #203

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -19,9 +19,11 @@ import (
 	"net"
 	"os"
 	"syscall"
+	"time"
 
 	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
 	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/vishvananda/netlink"
+	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/vishvananda/netlink/nl"
 
 	"github.com/coreos/flannel/pkg/ip"
 )
@@ -65,6 +67,9 @@ func newVXLANDevice(devAttrs *vxlanDeviceAttrs) (*vxlanDevice, error) {
 	if err != nil {
 		return nil, err
 	}
+	// this enables ARP requests being sent to userspace via netlink
+	sysctlPath := fmt.Sprintf("/proc/sys/net/ipv4/neigh/%s/app_solicit", devAttrs.name)
+	sysctlSet(sysctlPath, "3")
 
 	return &vxlanDevice{
 		link: link,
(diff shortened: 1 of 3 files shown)


backend/vxlan/device.go
```diff
@@ -198,13 +198,63 @@ func (dev *vxlanDevice) DelL3(n neigh) error {
...
+		log.Error("Failed to subscribe to netlink RTNLGRP_NEIGH messages")
+		return
+	}
+
+	for {
+		msgs, err := nlsock.Receive()
+		if err != nil {
+			log.Errorf("Failed to receive from netlink: %v ", err)
+
+			// wait 1 sec before retrying but honor the cancel channel
```
eyakubovich: This comment is a relic, please remove it.

backend/vxlan/routes.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"bytes"
+	"net"
+
+	"github.com/coreos/flannel/pkg/ip"
+)
+
+type route struct {
+	network ip.IP4Net
+	vtepIP  net.IP
```
eyakubovich: This is not used anymore b/c we don't process L2 misses anymore. Please remove.

eyakubovich:
LGTM

PR #214, opened 2015-06-16
GCE backend

MohdAhmad:


eyakubovich:
Can you include in `gce.go` the message from https://github.com/MohdAhmad/flannel/blob/36314ccfdb5fb771f7aef370abad51d0c35d1776/contrib/flannel-route-manager/LICENSE at the top and mention that this portion is derivative work. It's ok to re-license but we need to preserve the notice and give credit.

eyakubovich:
Please update README.md to document the backend.

backend/gce/gce.go
```diff
@@ -0,0 +1,249 @@
...
+
+		if i%5 == 0 {
+			log.Infof("Delete operation status: %v waiting for completion...", operation.Status)
+		}
+
+		if operation.Status == "DONE" {
+			break
+		}
+
+		if i > 100 {
```
eyakubovich: I would put the predicate into the `for` statement and then change
```
if operation.Status == "DONE" {
    return false, nil
}
```

and last return to `return false, fmt.Errorf("timeout waiting for delete operation to finish")`

backend/gce/gce.go
```diff
@@ -0,0 +1,249 @@
...
+	log.Infof("Inserting route for subnet: %v", subnet)
+	route := &compute.Route{
+		Name:            formatRouteName(g.gceNetwork.Name, subnet),
+		DestRange:       subnet,
+		Network:         g.gceNetwork.SelfLink,
+		NextHopInstance: g.gceInstance.SelfLink,
+		Priority:        1000,
+		Tags:            []string{},
+	}
+	return g.computeService.Routes.Insert(g.project, route).Do()
```
eyakubovich: Poll for DONE and check operation error.

PR #215, opened 2015-06-17
Update README to detail aws requirements

MohdAhmad:
Adds:
1. Permissions required by EC2 running flannel.
2. Disabling of the source/destination checks.
3. Mentions the auto route table id detection feature.


eyakubovich:
LGTM. Thank you, it will really help our users out.

PR #220, opened 2015-06-22
Updates docs to include resource limits

MohdAhmad:
Also updates the naming to routes to not include the network name.


eyakubovich:
LGTM

Issue #216, opened 2015-06-21
How to determine when flannelized IPs can't see one another? 


jayunit100:
I've setup **flannel 0.2** between some machines using a vagrant+virtualbox recipe with etcd, and **docker 1.6**.     All indicators which I can think of seem to be green, **but** containers aren't able to ping/reach one another.  What other items shall I check for ? 

So far I've verified:

1) The following checklist items seem to be working:

- can start flannel as the default networking backdrop on two independent VMs in virtualbox, 
- using one etcd server, which is working perfectly, 
- and each of the VMs is getting a non 172 default address  (i.e. 10.x.y.z)
- the flannel network interface is indeed showing up (so is the docker0) in `ip a` both appear to be the same subnet.

2) Additionally, the **flanneld logs** are not indicating any failures.

machine1: 
```
Jun 21 12:19:09 kube0.ha systemd[1]: Starting Flanneld overlay address etcd agent...
Jun 21 12:19:09 kube0.ha flanneld[19445]: I0621 12:19:09.810365   19445 main.go:247] Installing signal handlers
Jun 21 12:19:09 kube0.ha flanneld[19445]: I0621 12:19:09.810399   19445 main.go:118] Determining IP address of default interface
Jun 21 12:19:09 kube0.ha flanneld[19445]: I0621 12:19:09.810994   19445 main.go:205] Using 10.0.2.15 as external interface
Jun 21 12:19:12 kube0.ha flanneld[19445]: I0621 12:19:12.199864   19445 subnet.go:83] Subnet lease acquired: 10.0.24.0/24
Jun 21 12:19:12 kube0.ha flanneld[19445]: I0621 12:19:12.207320   19445 main.go:215] VXLAN mode initialized
Jun 21 12:19:12 kube0.ha flanneld[19445]: I0621 12:19:12.207750   19445 vxlan.go:115] Watching for L2/L3 misses
Jun 21 12:19:12 kube0.ha flanneld[19445]: I0621 12:19:12.208097   19445 vxlan.go:121] Watching for new subnet leases
Jun 21 12:19:12 kube0.ha systemd[1]: Started Flanneld overlay address etcd agent.
```
machine 2:
```
[root@kube0 vagrant]# 
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Jun 21 12:18:33 kube1.ha systemd[1]: Starting Flanneld overlay address etcd agent...
[...]

PR #223, opened 2015-06-23
Don't allow extra args

jayunit100:
Related to #216  ... If you mistakenly forget to add the right arg names to  flannel , then it just silently ignores them.  that causes alot of confusion .  Lets fail if there are wrong/bad/unknown/ignored args sent in, and fail fast.


eyakubovich:
LGTM. Thank you for fixing it.

PR #225, opened 2015-06-24
prevent failed builds from being shadowed by old working builds

jayunit100:
Minor put important improvement to prevent binary failed builds shadowed by old builds

### DIFF
--- a/build
+++ b/build
@@ -13,9 +13,16 @@ export GOPATH=${PWD}/gopath
 
 eval $(go env)
 
-if [ ${GOOS} = "linux" ]; then
-	echo "Building flanneld..."
-	go build -o ${GOBIN}/flanneld ${REPO_PATH}
+if [ -f $GOBIN/flanneld ]; then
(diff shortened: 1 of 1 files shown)


eyakubovich:
A failed build will be vocal so you know it has failed. It is very typical to not delete a binary before starting a build. Manual deletion also impedes edit-compile-run cycle by adding an extra step.

Can you speak to how this has been a problem for you?

jayunit100:
testing th e binary , after build - leads to testing an old artifact.  so iits quite misleading imo... but i see your point to.

eyakubovich:
We can add a Makefile with `clean` target (and others that forward to our bash scripts) but it maybe simpler to just add a simple `clean` script to wipe `./bin/flanneld`.

jayunit100:
Yup Makefile is probably the safest bet.  I can look at it this wk.  good feedback thanks ;

philips:
Hello @jayunit100 I am going to close this out. Please re-open if you have a chance to work on it.

PR #227, opened 2015-06-25
Update readme with client/server and multi-network modes

eyakubovich:
- Added client/server section
- Added multi-network section
- Added environment variables section

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,51 @@ Additionally it will monitor etcd for new members of the network and adjust the 
 
 After flannel has acquired the subnet and configured backend, it will write out an environment variable file (`/run/flannel/subnet.env` by default) with subnet address and MTU that it supports.
 
+## Client/Server mode (EXPERIMENTAL)
+
+While flannel has been designed to work without a need for a central controller, utilizing etcd for coordination, it can be configured to run in a client/server mode.
+Such setup offers the advantange of having only a single (or a handful) of server having access to etcd with flannel daemons accessing etcd via the server.
+The server is completely stateless and does not assume that it has exclusive access to the etcd keyspace.
+In the future this will be exploited to provide failover.
+Currently though, the clients accept only a single endpoint to which to connect to.
+It is important to note that the server itself does not join the flannel network (it won't assign itself a subnet) -- it just satisfies requests from the clients.
+
+To run a server on a host with 10.0.0.3 IP address:
+```
+$ flanneld --listen=0.0.0.0:8888
+```
+
+To run flannel daemon in client mode:
+```
+$ flanneld --remote=10.0.0.3
+```
+
+## Multi-network mode (EXPERIMENTAL)
+
+Multi-network mode allows a single flannel daemon to join multiple networks.
+Each network is independent from each other and has its own configuration, IP space, interfaces.
+To configure three networks, `blue`, `green`, and `red`, start by publishing their configurations to etcd in different directories:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,6 +132,51 @@ Additionally it will monitor etcd for new members of the network and adjust the
...
+Multi-network mode allows a single flannel daemon to join multiple networks.
+Each network is independent from each other and has its own configuration, IP space, interfaces.
+To configure three networks, `blue`, `green`, and `red`, start by publishing their configurations to etcd in different directories:
+```
+$ etcdctl set /coreos.com/network/blue/config  '{ "Network": "10.1.0.0/16", "Backend": { "Type": "vxlan", "VNI": 1 } }'
+$ etcdctl set /coreos.com/network/green/config '{ "Network": "10.2.0.0/16", "Backend": { "Type": "vxlan", "VNI": 2 } }'
+$ etcdctl set /coreos.com/network/red/config   '{ "Network": "10.3.0.0/16", "Backend": { "Type": "vxlan", "VNI": 3 } }'
+```
+
+Next, start the flannel daemon, specifying the etcd prefix (we used one different from default) and the networks to join:
```
jonboulle: > (we used one different from default)

@eyakubovich what does this mean? etcd prefix isn't specified in the example unless you just mean implicitly because --networks is specified

eyakubovich: It's a bug. I used a different prefix in the example initially but then decided against it to keep things simpler.

README.md
```diff
@@ -132,6 +132,51 @@ Additionally it will monitor etcd for new members of the network and adjust the
...
+```
+$ flanneld --networks=blue,green,red
```
jonboulle: what if I'm mixing client/server mode and --networks, does that work? if so what does it look like?

README.md
```diff
@@ -132,6 +132,51 @@ Additionally it will monitor etcd for new members of the network and adjust the
 
 After flannel has acquired the subnet and configured backend, it will write out an environment variable file (`/run/flannel/subnet.env` by default) with subnet address and MTU that it supports.
 
+## Client/Server mode (EXPERIMENTAL)
```
jonboulle: If I have a client/server set up can I also run a normal flannel daemon in the same overlay?

README.md
eyakubovich: It does work. I considered having a separate section on that but decided that they're orthogonal enough that it should just make sense.

README.md
eyakubovich: You can (since the server is stateless) and we should document that.

README.md
eyakubovich: But since you asked, at least a note should be added.

jonboulle:
@eyakubovich updated for your consideration, could do with the additional notes we discussed up there ^

eyakubovich:
@jonboulle thank you for the edits. PTAL.

README.md
```diff
@@ -132,6 +132,65 @@ Additionally it will monitor etcd for new members of the network and adjust the
...
+
+Next, start the flannel daemon, specifying the networks to join:
+```
+$ flanneld --networks=blue,green,red
+```
+
+Instead of writing out a single `/run/flannel/subnet.env` file with flannel parameters, it will create a .env file for each network in the directory `/run/flannel/networks`:
+```
+$ ls /run/flannel/networks/
+bar.env  baz.env  foo.env
```
MohdAhmad: These correspond to blue, green and red, right?

eyakubovich: Haha, yes. I had them named bar, baz, and foo on my machine. Good catch!

MohdAhmad:
LGTM. Looking forward to trying this out myself.

PR #235, opened 2015-06-30
Disable srcDest check

### LINKED ISSUE #230: aws-vpc backend: flannel should be able to disable SrcDestCheck
Flannel requires `SrcDestCheck` to be disabled on each instance when `aws-vpc` backend is used (see how http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck).

There is no easy way to disable `SrcDestCheck` when CoreOS is managed by CloudFormation in an ASG. So flannel seems like a good candidate to disable instance `SrcDestCheck` attribute when it starts?

IAM policy action that is required to do so is `ec2:ModifyInstanceAttribute`.

MohdAhmad:
Permissions required ec2:ModifyInstanceAttribute. Returns a warning if
it fails. Updates readme to reflect change.

Fixes #230

### DIFF
--- a/backend/awsvpc/awsvpc.go
+++ b/backend/awsvpc/awsvpc.go
@@ -102,6 +102,10 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
 	}
 	ec2c := ec2.New(auth, region)
 
+	if _, err = m.DisableSrcDestCheck(instanceID, ec2c); err != nil {
+		log.Info("Warning- disabling source destination check falied!: %v", err)
+	}
+
 	if m.cfg.RouteTableID == "" {
 		log.Infof("RouteTableID not passed as config parameter, attempting to detect")
 		routeTableID, err := m.DetectRouteTableID(instanceID, ec2c)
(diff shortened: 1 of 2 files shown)


backend/awsvpc/awsvpc.go
```diff
@@ -162,6 +166,15 @@ func (m *AwsVpcBackend) Init(extIface *net.Interface, extIP net.IP) (*backend.Su
 	}, nil
 }
 
+func (m *AwsVpcBackend) DisableSrcDestCheck(instanceID string, ec2c *ec2.EC2) (*ec2.ModifyInstanceResp, error) {
+	modifyAttributes := &ec2.ModifyInstance{
+		SourceDestCheck:    false,
+		SetSourceDestCheck: true,
+	}
+
+	return ec2c.ModifyInstance(instanceID, modifyAttributes)
```
eyakubovich: I'm assuming this won't fail if the source/dest check is already disabled, correct?

eyakubovich:
LGTM other than the assumption above.

backend/awsvpc/awsvpc.go
MohdAhmad: Yes, this will return an unauthorized operation error which we log.

PR #237, opened 2015-07-01
Reuses route entry if match found

### LINKED ISSUE #226: Restarting flannel with aws-vpc backend will complain about active route being deleted
If flannel is restarted (or instance simply rebooted), flannel will print something like:
`Matching *active* entry to: 10.1.59.0/24 that will be deleted: 10.1.59.0/24`

This is benign as it will delete and re-add the same route but a good optimization would be to avoid deleting a route that's about to be added.

MohdAhmad:
If a matching route entry is present in the route table, flannel will reuse it instead of deleting it and inserting a new one. In addition, the checking for matching routes and creating routes is moved into a separate functions.

Fixes #226


eyakubovich:
LGTM

Issue #229, opened 2015-06-25
Add TLS support to server mode


eyakubovich:

eyakubovich:
Fixed by #244

PR #244, opened 2015-07-09
Add TLS support for client/server communication

### LINKED ISSUE #178: Figure out auth for client/server mode
Once https://github.com/coreos/flannel/pull/177 lands, need to figure out the auth and the whole security piece.

eyakubovich:
Analogous to etcd, optionally use TLS to encrypt communication
between the client and the server. Also, as an option, client side
certificates can be used to authenticate connecting clients.

Fixes #178 and #229

### DIFF
--- a/Documentation/client-server.md
+++ b/Documentation/client-server.md
@@ -1,0 +1,103 @@
+## Client/Server mode (EXPERIMENTAL)
+
+### Getting Started
+
+By default flannel runs without a central controller, utilizing etcd for coordination.
+However, it can also be configured to run in client/server mode, where a special instance of the flannel daemon (the server) is the only one that communicates with etcd.
+This setup offers the advantange of having only a single server directly connecting to etcd, with the rest of the flannel daemons (clients) accessing etcd via the server.
+The server is completely stateless and does not assume that it has exclusive access to the etcd keyspace.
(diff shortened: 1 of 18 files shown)


Documentation/client-server.md
```diff
@@ -0,0 +1,103 @@
...
+```
+$ flanneld --listen=fd://5
+```
+
+### Use of SSL/TLS to secure client/server communication
+
+By default, the communication between the client and server is unencrypted (uses HTTP).
+Just like the link between flannel and etcd can be secured via SSL/TLS, so too can the link between the client and the server be encrypted using SSL/TLS.
+You will need a CA certificate and also a private key, certificate pair for the server.
+The server certificate must be signed the corresponding CA.
```
MohdAhmad: signed *by the

remote/client.go
```diff
@@ -165,11 +188,11 @@ type httpRespErr struct {
 	err  error
 }
 
-func httpDo(ctx context.Context, req *http.Request) (*http.Response, error) {
+func (m *RemoteManager) httpDo(ctx context.Context, req *http.Request) (*http.Response, error) {
 	// Run the HTTP request in a goroutine (so it can be canceled) and pass
 	// the result via the channel c
 	tr := &http.Transport{}
```
MohdAhmad: Please delete.

MohdAhmad:
Apart from the two comments, LGTM.

Issue #254, opened 2015-07-16
duplicate route added for host-gw backend


bhouse:
Hi!

We're having an issue with flannel in our kubernetes cluster running with the `host-gw` backend. We're running on AWS with flannel v0.5.1.

We are running a replication controller and service with a simple container that just responds with `true` to http requests on port 80. Everything works fine setting up the pod/rc/service, but when we query the service we get a timeout with `no route to host`.

We traced this all the way down to the request making it to the `kube-proxy` process, and then eventually timing out trying to contact the pod ip:port. On the minion itself there are duplicate routes for the flannel subnet that has been assigned to the minion, where it first routes traffic for the subnet to the network interface, and below that routes traffic to the docker bridge interface.

Our theory is this is resulting in a local routing loop, everytime we query the packet continuously gets routed to the nic instead of to docker until it times out. When we remove the specific route for the nic then the service starts working.

In this example, the minion ip address is `10.0.239.99`, and it gets `10.0.231.0/24` as a subnet for pod addresses from flannel.

```
# Flanneld log output
 
Jul 16 17:54:29 bhouse-minion1.aws1.zdsystest.com sdnotify-proxy[623]: I0716 17:54:29.911888 00001 etcd.go:212] Picking subnet in range 10.0.225.0 ... 10.0.239.0
Jul 16 17:54:29 bhouse-minion1.aws1.zdsystest.com sdnotify-proxy[623]: go-etcd2015/07/16 17:54:29 DEBUG: put /coreos.com/network/subnets/10.0.231.0-24, {"PublicIP":"10.0.239.99","BackendType":"host-gw"}, ttl: 86400, [http://10.0.239.77:2379]
Jul 16 17:54:29 bhouse-minion1.aws1.zdsystest.com sdnotify-proxy[623]: go-etcd2015/07/16 17:54:29 DEBUG: Connecting to etcd: attempt 1 for keys/coreos.com/network/subnets/10.0.231.0-24?prevExist=false
Jul 16 17:54:29 bhouse-minion1.aws1.zdsystest.com sdnotify-proxy[623]: go-etcd2015/07/16 17:54:29 DEBUG: send.request.to http://10.0.239.77:2379/v2/keys/coreos.com/network/subnets/10.0.231.0-24?prevExist=false | method PUT
Jul 16 17:54:30 bhouse-minion1.aws1.zdsystest.com sdnotify-proxy[623]: go-etcd2015/07/16 17:54:30 DEBUG: recv.response.from
[...]

PR #255, opened 2015-07-17
Ignore route to local ext IP in hostgw mode

lvlv:
Route to local ext IP may get added when local docker0 is not yet setup, see:

[root@localhost ~]# ip route
default via 10.0.2.2 dev enp0s3
10.0.2.0/24 dev enp0s3  proto kernel  scope link  src 10.0.2.15
**10.10.176.0/20 via 172.17.8.101 dev enp0s8**
**10.10.176.0/20 dev docker0  proto kernel  scope link  src 10.10.176.1**
10.10.192.0/20 via 172.17.8.102 dev enp0s8
10.15.240.0/20 via 172.17.8.100 dev enp0s8
169.254.0.0/16 dev enp0s3  scope link  metric 1002
169.254.0.0/16 dev enp0s8  scope link  metric 1003
172.17.8.0/24 dev enp0s8  proto kernel  scope link  src 172.17.8.101

### DIFF
--- a/backend/hostgw/hostgw.go
+++ b/backend/hostgw/hostgw.go
@@ -145,6 +145,9 @@ func (rb *HostgwBackend) handleSubnetEvents(batch []subnet.Event) {
 				Gw:        evt.Lease.Attrs.PublicIP.ToIP(),
 				LinkIndex: rb.extIface.Index,
 			}
+			if rb.extIP.Equal(route.Gw) {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


eyakubovich:
LGTM. This is a good short term fix but we need a general mechanism for ignoring our own leases.

MohdAhmad:
This fixes #254, right?

eyakubovich:
@MohdAhmad correct.

bhouse:
thanks!

PR #269, opened 2015-07-31
Actually process -m flag in mk-docker-opts.sh

### LINKED ISSUE #268: mk-docker-opts.sh do not handle -m parameter at all
In code `flannel/dist/mk-docker-opts.sh`:

**1) There is no "m" in getopts, so -m parameter is ignored**

```
while getopts "f:d:ick:" opt; do
    case $opt in
        m)
            ipmasq=false
            ;;
done
```

User would always get an error when use -m parameter:

```
$ sudo flannel-0.5.2/mk-docker-opts.sh -m
flannel-0.5.2/mk-docker-opts.sh: illegal option -- m
flannel-0.5.2/mk-docker-opts.sh [-f FLANNEL-ENV-FILE] [-d DOCKER-ENV-FILE] [-i] [-c] [-m] [-k COMBINED-KEY]

Generate Docker daemon options based on flannel env file
OPTIONS:
	-f	Path to flannel env file. Defaults to /run/flannel/subnet.env
	-d	Path to Docker env file to write to. Defaults to /run/docker_opts.env
	-i	Output each Docker option as individual var. e.g. DOCKER_OPT_MTU=1500
	-c	Output combined Docker options into DOCKER_OPTS var
	-k	Set the combined options key to this value (default DOCKER_OPTS=)
	-m	Do not output --ip-masq (useful for older Docker version)
```

**2) variable FLANNEL_IPMASQ never have a value**

Even I add "m" to getops list, the -m parameter still not works as expected. Below result shows no matter -m is used or not, DOCKER_OPT_IPMASQ is not created at all.

```
$ sudo flannel-0.5.2/mk-docker-opts.sh -d /run/flannel_docker_opts.env -i -m
$ cat /run/flannel_docker_opts.env
DOCKER_OPT_BIP="--bip=172.17.78.1/24"
DOCKER_OPT_MTU="--mtu=8973"
$ sudo flannel-0.5.2/mk-docker-opts.sh -d /run/flannel_docker_opts.env -i
$ cat /run/flannel_docker_opts.env
DOCKER_OPT_BIP="--bip=172.17.78.1/24"
DOCKER_OPT_MTU="--mtu=8973"
```

From the code I found the root cause is the judgement variable "FLANNEL_IPMASQ" never defined, so it won't equal true ever.

```
if [ "$FLANNEL_IPMASQ" = true ] && [ $ipmasq = true ] ; then
        DOCKER_OPT_IPMASQ="--ip-masq=false"
fi
```

Actually just remove the "FLANNEL_IPMASQ" judgement will fix the issue. Like below:

```
if [ $ipmasq = true ]; then
        DOCKER_OPT_IPMASQ="--ip-masq=false"
fi
```

**3) Also there is a small problem with default Docker env file**

Current the default Docker env file is `/run/docker_opts.env`, however in CoreOS the default env file for Docker is `/run/flannel_docker_opts.env`. It's a bit inconvenient.

I do wondering how no one ever find 'DOCKER_OPT_IPMASQ' missing from their env file generated...

eyakubovich:
- Process -m flag
- If requested, output DOCKER_OPT_IPMASQ var regardless if it's true or
  false

Fixes #268

### DIFF
--- a/dist/mk-docker-opts.sh
+++ b/dist/mk-docker-opts.sh
@@ -23,7 +23,7 @@ indiv_opts=false
 combined_opts=false
 ipmasq=true
 
-while getopts "f:d:ick:" opt; do
+while getopts "f:d:icmk:" opt; do
 	case $opt in
 		f)
 			flannel_env=$OPTARG
@@ -66,8 +66,15 @@ if [ -n "$FLANNEL_MTU" ]; then
 	DOCKER_OPT_MTU="--mtu=$FLANNEL_MTU"
 fi
 
-if [ "$FLANNEL_IPMASQ" = true ] && [ $ipmasq = true ] ; then
-	DOCKER_OPT_IPMASQ="--ip-masq=false"
+if [ -n "$FLANNEL_IPMASQ" ] && [ $ipmasq = true ] ; then
+	if [ "$FLANNEL_IPMASQ" = true ] ; then
+		DOCKER_OPT_IPMASQ="--ip-masq=false"
+	elif [ "$FLANNEL_IPMASQ" = false ] ; then
+		DOCKER_OPT_IPMASQ="--ip-masq=true"
+	else
+		echo "Invalid value of FLANNEL_IPMASQ: $FLANNEL_IPMASQ" > /dev/stderr
+		exit 1
+	fi
 fi
 
 eval docker_opts="\$${combined_opts_key}"


linfan:
Hi Eugene,
I had a look in the new script, it seems that FLANNEL_IPMASQ variable still undefined.
So DOCKER_OPT_IPMASQ still never output.

```
$ sudo flannel-0.5.2/mk-docker-opts.sh -i -m        <-- with 'm' parameter
$ cat /run/docker_opts.env
DOCKER_OPT_BIP="--bip=172.17.78.1/24"
DOCKER_OPT_MTU="--mtu=8973"

$ sudo flannel-0.5.2/mk-docker-opts.sh -i        <-- without 'm' parameter
$ cat /run/docker_opts.env
DOCKER_OPT_BIP="--bip=172.17.78.1/24"
DOCKER_OPT_MTU="--mtu=8973"
```

eyakubovich:
@linfan FLANNEL_IPMASQ is defined in subnet.env and is sourced in.

linfan:
Hi Eugene,
Yes, I just got the point!!
In this case, the fix looks good 😀

eghobo:
@eyakubovich: but how I can pass `-m` option if i run flannel inside docker container at CoreOS. for `host-gw` case we need `--ip-masq=true` for flannel and docker.

eyakubovich:
@eghobo On CoreOS flannel is started with `--ip-masq=true` and thus docker with `--ip-masq=false`. You don't want to start both flannel and docker with `--ip-masq=true` or they'll conflict.

eghobo:
@eyakubovich: what kind of conflict? we run like thins at Ubuntu, i enabled `--ip-masq=true` to make kub-proxy works. And for CoreOS I need to enable it for `host-gw` mode, otherwise containers cannot ping each other #506

eyakubovich:
Both will install masquerade rules. Docker will install a rule to masquerade anything coming out of `docker0` and going elsewhere. That will cause container-to-container traffic (on different hosts) to get masqueraded. flannel installs a rule that only masquerades traffic if it's not destined for another flannel address.

Issue #267, opened 2015-07-29
VXLAN mode is broken for kernel 3.19.0


klizhentas:
I've installed flanneld and docker on Ubuntu 14.04 with upgraded kernel 3.19
The ARP entry of the remote flannel.1 was not added to the ARP table.

After downgrading kernel back to 3.13, the issue was resolved - and ARP entry was added. I'm happy to help fixing the issue, if you point me to the right direction.

Let me know if you have any questions.

klizhentas:
just tried 3.18.18-031818-generic with the same result. Seems something have changed since 3.13

eyakubovich:
Also reported in https://github.com/coreos/flannel/issues/260

We need to figure out which version fixed it again. Would actually be great to understand what the actual bug was.

klizhentas:
@eyakubovich do you know the version > 3.13 that works?

eyakubovich:
@klizhentas I don't. I'll try to try a few versions to see which ones work but it may not be till next week. Do you have time to try other ones?

klizhentas:
sure, is there any particular version you would like to check?

klizhentas:
also, can you point me to the chunk of code that you think is not triggered? I can add debugging info and see what's not working

eyakubovich:
I'd be interested in 3.18 and something from 4.x series.

I think the ARP misses might not be getting through: https://github.com/coreos/flannel/blob/master/backend/vxlan/device.go#L208

That's my only guess for now.

klizhentas:
I've tried latest stable `3.18.18-031818-generic` and it has the same problem, I will try 4.x and will update the ticket

changx:
I have the same issue on 3.19.3-031903-generic,  it works on 3.16.0-0.bpo ( aka 3.16.7-ckt11   wheezy-backports ) , but I can not reverse back due to 3.16.7 racing condition under kvm :/ .

eyakubovich:
I can reproduce this on my Ubuntu box (3.19.0-23). You can see that ARP (L3) misses are not delivered to flannel. Yet when I booted CoreOS 640.0.0 (3.19.3), it works fine. Could it be that the problem is with one of the Ubuntu patches?

eyakubovich:
Fixed in #272

klizhentas:
tested on the 3.19 and it now works fine, thanks!

PR #272, opened 2015-08-04
vxlan: check of nil IP in fdb entry

### LINKED ISSUE #260: Getting "fdb already populated with: <nil>" when starting flannel
I have set up flannel between two hosts using the client-server method. One of the hosts has the flannel server running and both hosts connect to it as clients. During the initial setup, everything went fine and containers inside both hosts were able to ping each other. However, after rebooting, I see that the pings fail and instead of having a host IP in when a new host joins, flannel gives the message 

    fdb already populated with: <nil> MACADDRESS

here MACADDRESS is some random mac ( I'm assuming it is the container's veth's mac)

Am I doing somthing wrong?

eyakubovich:
Bug fix where nil IP (byte slice) was dereferenced
and caused the goroutine to hang.

Fixes #260 and #267

### DIFF
--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -308,7 +308,7 @@ func (vb *VXLANBackend) handleInitialSubnetEvents(batch []subnet.Event) error {
 	}
 
 	for j, marker := range fdbEntryMarker {
-		if !marker {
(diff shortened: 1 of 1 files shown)


MohdAhmad:
LGTM

eyakubovich:
Note: not sure if this is a final solution or short term fix, will see.

PR #275, opened 2015-08-05
New etcd client

### LINKED ISSUE #174: Transition to etcd/client instead of go-etcd
The new etcd/client is maturing so it's worth moving over to it. This needs to be done after https://github.com/coreos/flannel/pull/169 is merged as it added golang/context library support.

eyakubovich:
go-etcd is dated and has logging that is either too silent or too verbose.

Fixes #174


yichengq:
generally LGTM, though i am not familiar with flannel code base.

PR #277, opened 2015-08-10
Add FLANNEL_NETWORK to subnet.env

eyakubovich:
CNI plugin needs to know the whole flannel network
in order to set up a route for it.


MohdAhmad:
LGTM

PR #282, opened 2015-08-11
remote: ensure testing HTTP requests are canceled

dcbw:
Due to https://github.com/golang/go/issues/11013 race conditions
in the remote testcases mean that HTTP requests aren't actually
canceled.  In client.go::httpDo() the goroutine that actually
runs the http.Do() may not execute until after that routine's
select{} loop, at which point the Context may have already been
canceled.  Unfortunately with Go < 1.5 there is a race in the
http.Transport package that drops cancellations on the floor.

This issue can be reproduced by using the remote_test.go hunks
that use a WaitGroup to wait for subnet.WatchLeases() to exit
after the Context is canceled, but not the client.go hunks.
Specifically, doTestWatch() calls cancel() before
watch.go::WatchLeases() calls sm.WatchLeases(), which in this
case is RemoteManager.WatchLeases(), which then calls
RemoteManager.httpGet() -> httpDo(), which then calls
http.Client.Do() which has the aformentioned cancellation race.

Work around the Go http.Transport cancellation issue by checking
a private cancellation variable, since context.Context has no
mechanism to check whether it has already been canceled.  This
can be removed once Flannel can depend on Go 1.5, at which point
the code should start to use http.Request.Cancel instead of
calling http.transport.CancelRequest().  This obviously doesn't
fix the race, but I think that has to wait for Go 1.5.

Second, use a WaitGroup in doTestWatch() to assert that the
outstanding watch is actually canceled.

### DIFF
--- a/remote/client.go
+++ b/remote/client.go
@@ -19,8 +19,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"path"
+	"time"
 
 	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/coreos/etcd/pkg/transport"
 	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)


dcbw:
So the commit fixes the hang on my local machine, but doesn't fix the hang in Travis.  Clearly a more sufficient fix is required.  Any thoughts on this problem?

The reason I ask is that I'm trying to implement Network watching, and the Lease listeners are hanging around and screwing up the Network watch testcases.

dcbw:
Ok, so we end up having to suck in most of http.Transport and then hack in some cancelation code until we can use Go 1.5.  The latest push fixes the issue on both my local machine and on the Travis builds.

eyakubovich:
LGTM. Not pretty, of course, but we should be very temporary. Thank you for tracking this down.

PR #284, opened 2015-08-12
subnet: add infrastructure and tests for network watches

dcbw:
This PR depends on https://github.com/coreos/flannel/pull/282 (which is the first commit here too) so that one should get sorted out first.

### DIFF
--- a/remote/client.go
+++ b/remote/client.go
@@ -164,13 +164,11 @@ func (m *RemoteManager) RenewLease(ctx context.Context, network string, lease *s
 	return nil
 }
 
-func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor interface{}) (subnet.WatchResult, error) {
-	url := m.mkurl(network, "leases")
-
+func (m *RemoteManager) watch(ctx context.Context, url string, cursor interface{}, wr interface{}) error {
 	if cursor != nil {
 		c, ok := cursor.(string)
 		if !ok {
-			return subnet.WatchResult{}, fmt.Errorf("internal error: RemoteManager.WatchLeases received non-string cursor")
+			return fmt.Errorf("internal error: RemoteManager.watch received non-string cursor")
 		}
 
 		url = fmt.Sprintf("%v?next=%v", url, c)
@@ -178,19 +176,44 @@ func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor 
 
 	resp, err := m.httpGet(ctx, url)
 	if err != nil {
-		return subnet.WatchResult{}, err
+		return err
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return subnet.WatchResult{}, httpError(resp)
+		return httpError(resp)
 	}
 
-	wr := subnet.WatchResult{}
-	if err := json.NewDecoder(resp.Body).Decode(&wr); err != nil {
-		return subnet.WatchResult{}, err
+	if err := json.NewDecoder(resp.Body).Decode(wr); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor interface{}) (subnet.LeaseWatchResult, error) {
+	url := m.mkurl(network, "leases")
+
+	wr := subnet.LeaseWatchResult{}
+	err := m.watch(ctx, url, cursor, &wr)
+	if err != nil {
+		return subnet.LeaseWatchResult{}, err
 	}
 	if _, ok := wr.Cursor.(string); !ok {
-		return subnet.WatchResult{}, fmt.Errorf("lease watch returned non-string cursor")
+		return subnet.LeaseWatchResult{}, fmt.Errorf("watch returned non-string cursor")
+	}
+
+	return wr, nil
+}
+
+func (m *RemoteManager) WatchNetworks(ctx context.Context, cursor interface{}) (subnet.NetworkWatchResult, error) {
+	wr := subnet.NetworkWatchResult{}
+	err := m.watch(ctx, m.base+"/", cursor, &wr)
+	if err != nil {
+		return subnet.NetworkWatchResult{}, err
+	}
+
+	if _, ok := wr.Cursor.(string); !ok {
+		return subnet.NetworkWatchResult{}, fmt.Errorf("watch returned non-string cursor")
 	}
 
 	return wr, nil
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -33,7 +33,8 @@ const expectedNetwork = "10.1.0.0/16"
 
 func TestRemote(t *testing.T) {
 	config := fmt.Sprintf(`{"Network": %q}`, expectedNetwork)
-	sm := subnet.NewMockManager(1, config)
+	serverRegistry := subnet.NewMockRegistry(1, "", config, nil)
+	sm := subnet.NewMockManager(serverRegistry)
 
 	addr := "127.0.0.1:9999"
 
@@ -45,7 +46,7 @@ func TestRemote(t *testing.T) {
 		wg.Done()
 	}()
 
-	doTestRemote(ctx, t, addr)
+	doTestRemote(ctx, t, addr, serverRegistry)
 
 	cancel()
 	wg.Wait()
@@ -76,7 +77,7 @@ func isConnRefused(err error) bool {
 	return false
 }
 
-func doTestRemote(ctx context.Context, t *testing.T, remoteAddr string) {
+func doTestRemote(ctx context.Context, t *testing.T, remoteAddr string, serverRegistry *subnet.MockSubnetRegistry) {
 	sm, err := NewRemoteManager(remoteAddr, "", "", "")
 	if err != nil {
 		t.Fatalf("Failed to create remote mananager: %v", err)
@@ -120,10 +121,12 @@ func doTestRemote(ctx context.Context, t *testing.T, remoteAddr string) {
 		t.Errorf("RenewLease failed: %v", err)
 	}
 
-	doTestWatch(t, sm)
+	doTestWatchLeases(t, sm)
+
+	doTestWatchNetworks(t, sm, serverRegistry)
 }
 
-func doTestWatch(t *testing.T, sm subnet.Manager) {
+func doTestWatchLeases(t *testing.T, sm subnet.Manager) {
 	ctx, cancel := context.WithCancel(context.Background())
 	wg := sync.WaitGroup{}
 	wg.Add(1)
@@ -160,11 +163,72 @@ func doTestWatch(t *testing.T, sm subnet.Manager) {
 	}
 
 	evt := evtBatch[0]
-	if evt.Type != subnet.SubnetAdded {
+	if evt.Type != subnet.EventAdded {
 		t.Fatalf("WatchSubnets produced wrong event type")
 	}
 
 	if evt.Lease.Key() != l.Key() {
 		t.Errorf("WatchSubnet produced wrong subnet: expected %s, got %s", l.Key(), evt.Lease.Key())
+	}
+}
+
+func doTestWatchNetworks(t *testing.T, sm subnet.Manager, serverRegistry *subnet.MockSubnetRegistry) {
+	ctx, cancel := context.WithCancel(context.Background())
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	defer func() {
+		cancel()
+		wg.Wait()
+	}()
+
+	events := make(chan []subnet.Event)
+	go func() {
+		subnet.WatchNetworks(ctx, sm, events)
+		wg.Done()
+	}()
+
+	// skip over the initial snapshot
+	<-events
+
+	expectedNetname := "foobar"
+	config := fmt.Sprintf(`{"Network": %q}`, expectedNetwork)
+	_, err := serverRegistry.CreateNetwork(ctx, expectedNetname, config)
+	if err != nil {
+		t.Errorf("create network failed: %v", err)
+	}
+
+	evtBatch := <-events
+
+	if len(evtBatch) != 1 {
+		t.Fatalf("WatchNetworks create produced wrong sized event batch")
+	}
+
+	evt := evtBatch[0]
+	if evt.Type != subnet.EventAdded {
+		t.Fatalf("WatchNetworks create produced wrong event type")
+	}
+
+	if evt.Network != expectedNetname {
+		t.Errorf("WatchNetwork create produced wrong network: expected %s, got %s", expectedNetname, evt.Network)
+	}
+
+	_, err = serverRegistry.DeleteNetwork(ctx, expectedNetname)
+	if err != nil {
+		t.Errorf("delete network failed: %v", err)
+	}
+
+	evtBatch = <-events
+
+	if len(evtBatch) != 1 {
+		t.Fatalf("WatchNetworks delete produced wrong sized event batch")
+	}
+
+	evt = evtBatch[0]
+	if evt.Type != subnet.EventRemoved {
+		t.Fatalf("WatchNetworks delete produced wrong event type")
+	}
+
+	if evt.Network != expectedNetname {
+		t.Errorf("WatchNetwork delete produced wrong network: expected %s, got %s", expectedNetname, evt.Network)
 	}
 }
--- a/remote/server.go
+++ b/remote/server.go
@@ -152,6 +152,32 @@ func handleWatchLeases(ctx context.Context, sm subnet.Manager, w http.ResponseWr
 	jsonResponse(w, http.StatusOK, wr)
 }
 
+// GET /?next=cursor watches
+// GET / retrieves all networks
+func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	cursor := getCursor(r.URL)
+	wr, err := sm.WatchNetworks(ctx, cursor)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, err)
+		return
+	}
+
+	switch wr.Cursor.(type) {
+	case string:
+	case fmt.Stringer:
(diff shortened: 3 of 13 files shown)


dcbw:
I'm not quite sure why the Travis build is failing, I cannot reproduce on two different machines with either Go 1.4.2 or Go 1.3.3... @rajatchopra any thoughts?  In the remote test on the Travis build, client.go::doTestWatchNetworks() times out waiting to receive the NetworkAdded event from the server.

dcbw:
[test]

eyakubovich:
I really like the functionality, thanks for adding this.

dcbw:
@eyakubovich ok I'll rework to separate the watch structs.  Still chasing down why the Travis tests are failing too.

dcbw:
Ignore any asdfasdfadsf commits too, that's just my sad attempt to figure out why the Travis builds are failing.

dcbw:
@eyakubovich are the EventType strings part of the API/ABI too?  Do I need to keep those backwards compatible?

eyakubovich:
@dcbw Yes, they're part of the API. Granted, I don't think too many people use this remote functionality yet so I don't think it's a big deal to break the API even if we don't bump the version. However in this case I don't think there's a reason to do so.

dcbw:
@eyakubovich I've reverted the EventType API string changes and fixed up the failing Travis test.  I think this branch is ready for more review.  Thanks!

remote/client.go
```diff
@@ -178,19 +178,50 @@ func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor
...
-		return subnet.WatchResult{}, fmt.Errorf("lease watch returned non-string cursor")
+		return subnet.LeaseWatchResult{}, fmt.Errorf("lease watch returned non-string cursor")
+	}
+
+	return wr, nil
+}
+
+func (m *RemoteManager) WatchNetworks(ctx context.Context, cursor interface{}) (subnet.NetworkWatchResult, error) {
+	url := m.base + "/"
+	if cursor != nil {
```
eyakubovich: Can you factor the common code out into something like:

```
func (m *RemoteManager) watch(ctx context.Context, cursor interface{}, url string, wr {}interface) error {
    if cursor != nil {
        c, ok := cursor.(string)
        if !ok {
            return subnet.NetworkWatchResult{}, fmt.Errorf("internal error: RemoteManager.watch received non-string cursor")
        }

        url = fmt.Sprintf("%v?next=%v", url, c)
    }

    resp, err := m.httpGet(ctx, url)
    if err != nil {
        return err
    }

    if resp.StatusCode != http.StatusOK {
        return httpError(resp)
    }

    if err := json.NewDecoder(resp.Body).Decode(wr); err != nil {
        return err
    }
    if _, ok := wr.Cursor.(string); !ok {
        return subnet.NetworkWatchResult{}, fmt.Errorf("watch returned non-string cursor")
    }
}
```

remote/remote_test.go
```diff
@@ -168,3 +171,84 @@ func doTestWatch(t *testing.T, sm subnet.Manager) {
...
+	go func() {
+		subnet.WatchNetworks(ctx, sm, events)
+		wg.Done()
+	}()
+
+	// skip over the initial snapshot
+	<-events
+
+	expectedNetname := "foobar"
+	go func() {
```
eyakubovich: Why a separate goroutine? The current goroutine is just going to block on line 214 anyway.

remote/remote_test.go
```diff
@@ -168,3 +171,84 @@ func doTestWatch(t *testing.T, sm subnet.Manager) {
...
+		// Add a new network a short time later to ensure that the
+		// client-side network watch triggers a server watch before the
+		// network is added directly in the server with CreateNetwork()
+		time.Sleep(time.Second)
```
eyakubovich: I don't think I fully understand the need for sleep. Is the desire to exercise the code where the server does an etcd watch prior to the network key being created? In that case I guess it's ok but the code should work regardless. That is, this also needs to work:

1. Client side issues a watch
2. A new network is created in etcd
3. The server receives the watch and processes it by immediately returning
4. The client gets notified of the new network

remote/remote_test.go
```diff
@@ -168,3 +171,84 @@ func doTestWatch(t *testing.T, sm subnet.Manager) {
...
+	evt := evtBatch[0]
+	if evt.Type != subnet.NetworkAdded {
+		t.Fatalf("WatchNetworks create produced wrong event type")
+	}
+
+	if evt.Network != expectedNetname {
+		t.Errorf("WatchNetwork create produced wrong network: expected %s, got %s", expectedNetname, evt.Network)
+	}
+
+	go func() {
```
eyakubovich: Same as comments above

remote/remote_test.go
dcbw: > I don't think I fully understand the need for sleep. Is the desire to exercise the code where the server does an etcd watch prior to the network key being created?

The server needs to be watching before the new network is created on the client side, otherwise the server-side will completely miss the event.  So subnet.WatchNetworks() must be running and sitting in RemoteManager::httpDo() (which triggers the server's handleWatchNetworks() function -> etcd.go::WatchNetworks()) before we execute the CreateNetwork() call.  But since subnet.WatchNetworks() is done in a goroutine and the actual HTTP request that triggers the server's handleWatchNetworks is way down the callstack, there's no way to guarantee that this will all happen because they are concurrent.

Obviously this is only a problem in the testcases, because normal clients will never be creating networks.

So possibly my approach of directly modifying the server's registry to add the network and generate the event isn't the right way to go about this, because of this problem...

eyakubovich: Your approach of directly modifying the server's register is just fine -- you're just simulating an external event.

> The server needs to be watching before the new network is created on the client side, otherwise the server-side will completely miss the event.

This is the scary statement. `subnet.WatchNetworks` internally maintains a cursor so it should not miss an event. If `CreateNetwork` executes before `etcd.go::WatchNetworks()`, that should be fine. When it does get to `etcd.go::WatchNetworks()`, the cursor will be pointing slightly behind the "end" and it will immediately return the added network. That is the whole point of maintaining and propagating the cursor.

subnet/subnet.go
```diff
@@ -44,26 +44,37 @@ type (
 	EventType int
 
 	Event struct {
-		Type  EventType `json:"type"`
-		Lease Lease     `json:"lease"`
+		Type    EventType `json:"type"`
+		Lease   Lease     `json:"lease"`
+		Network string    `json:"network"`
```
eyakubovich: Since we're using the same Event struct for both subnet and lease events (which is fine), add `omitempty` so it's clean on the wire.

subnet/subnet.go
```diff
@@ -44,26 +44,37 @@ type (
...
 	}
 )
 
 const (
 	SubnetAdded EventType = iota
 	SubnetRemoved
+	NetworkAdded
```
eyakubovich: Let's just have two events: `Added` and `Removed`. Or maybe call them `EventAdded` and `EventRemoved`. But they will be shared between both the lease and network events. There should be no confusion since the API endpoints and methods are different.

remote/remote_test.go
dcbw: > This is the scary statement.

Fair enough; I'll figure out why the cursor isn't doing what it should be doing here.

remote/client.go
dcbw: > Can you factor the common code out into something like:

Done.

subnet/subnet.go
dcbw: > add omitempty so it's clean on the wire.

Done

subnet/subnet.go
dcbw: > Let's just have two events: Added and Removed. Or maybe call them EventAdded and EventRemoved.

Done.

dcbw:
I've got the stuff commented as "Done" done locally; I have to fix up one or two other things as a result of these, then figure out the cursor testcase issue, then I'll re-push, hopefully tomorrow.  Keep the comments coming if you're still reviewing, I'll get to them tomorrow or next week.  Thanks!

(my next steps are factoring out the network code into a network/manager.go that watches for networks coming and going, and creates or removes networks as needed, but that's another PR...)

remote/remote_test.go
dcbw: > Why a separate goroutine? The current goroutine is just going to block on line 214 anyway.

This got removed in some later cleanups.

dcbw:
@eyakubovich repushed; I believe all your comments have been addressed.  Thanks!

remote/client.go
```diff
@@ -164,13 +164,11 @@ func (m *RemoteManager) RenewLease(ctx context.Context, network string, lease *s
...
 
-func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor interface{}) (subnet.WatchResult, error) {
-	url := m.mkurl(network, "leases")
-
+func (m *RemoteManager) watch(ctx context.Context, url string, cursor interface{}, wr interface{}) error {
 	if cursor != nil {
 		c, ok := cursor.(string)
 		if !ok {
-			return subnet.WatchResult{}, fmt.Errorf("internal error: RemoteManager.WatchLeases received non-string cursor")
+			return fmt.Errorf("internal error: RemoteManager.WatchLeases received non-string cursor")
```
eyakubovich: `RemoteManager.WatchLeases` -> `RemoteManager.watch`

subnet/etcd.go
```diff
@@ -347,36 +395,118 @@ func parseSubnetWatchResponse(resp *etcd.Response) (WatchResult, error) {
...
 	}
 
-	cursor := watchCursor{index}
+	wr.Cursor = watchCursor{index}
 	wr.Snapshot = leases
-	wr.Cursor = cursor
+	return wr, nil
+}
+
+// leaseWatchReset is called when incremental network watch failed and we need to grab a snapshot
```
eyakubovich: `leaseWatchReset` -> `networkWatchReset`

subnet/registry.go
```diff
@@ -131,13 +132,19 @@ func (esr *etcdSubnetRegistry) deleteSubnet(ctx context.Context, network, sn str
...
 		AfterIndex: since,
 		Recursive:  true,
 	}
-	return esr.client().Watcher(key, opts).Next(ctx)
+	e, err := esr.client().Watcher(key, opts).Next(ctx)
+	return e, err
+}
+
+func (esr *etcdSubnetRegistry) getNetworks(ctx context.Context) (*etcd.Response, error) {
+	key := path.Join(esr.etcdCfg.Prefix, "network")
```
eyakubovich: What is this "network"? The current struct is:
```
- etcd-prefix
     |- mynetwork1
           | - config
           | - subnets
                  |- mysubnet1
                  |- mysubnet2
     |- mynetwork2
           | - config
           | - subnets
                  |- mysubnet1
                  |- mysubnet2
```

subnet/subnet.go
```diff
@@ -96,5 +105,7 @@ type Manager interface {
 	GetNetworkConfig(ctx context.Context, network string) (*Config, error)
 	AcquireLease(ctx context.Context, network string, attrs *LeaseAttrs) (*Lease, error)
 	RenewLease(ctx context.Context, network string, lease *Lease) error
-	WatchLeases(ctx context.Context, network string, cursor interface{}) (WatchResult, error)
+	WatchLeases(ctx context.Context, network string, cursor interface{}) (LeaseWatchResult, error)
+	GetNetworks(ctx context.Context) ([]string, error)
```
eyakubovich: Let's combine `GetNetworks` and `WatchNetworks` into one. `GetNetworks` is just a watch with `nil` cursor. I'm actually working on another feature and doing exactly that for leases. We can rename `WatchNetworks` to `GetNetworks` if you think it's better.

subnet/registry.go
dcbw: Yeah, just found that out when refactoring the network/ code to watch networks.  It's bogus; removed.

subnet/subnet.go
dcbw: Done in the latest push.  I opted for "WatchNetworks" instead of GetNetworks, but could go either way if you really prefer GetNetworks.

subnet/etcd.go
dcbw: Done in latest push.

remote/client.go
dcbw: Done in latest push.

dcbw:
Repushed with fixes for all your previous comments.

remote/client.go
```diff
@@ -178,19 +176,49 @@ func (m *RemoteManager) WatchLeases(ctx context.Context, network string, cursor
...
+
+func (m *RemoteManager) WatchNetworks(ctx context.Context, cursor interface{}) (subnet.NetworkWatchResult, error) {
+	wr := subnet.NetworkWatchResult{}
+	err := m.watch(ctx, m.base+"/", cursor, &wr)
+	if err != nil {
+		return subnet.NetworkWatchResult{}, err
+	}
+
+	// Only return a cursor for the 'watch' operation
+	if cursor != nil {
```
eyakubovich: Sorry, I think you misunderstood me a bit. Yes, I proposed combining watch+get much like you've done here. However you should always return a cursor. Otherwise the watch won't work -- you kick the watch off by getting all the networks (no cursor passed in) and then watch for increments (using cursor) if you so desire. I think what you have for `WatchLeases` is perfect.

remote/server.go
```diff
@@ -152,6 +152,38 @@ func handleWatchLeases(ctx context.Context, sm subnet.Manager, w http.ResponseWr
...
+	cursor := getCursor(r.URL)
+	wr, err := sm.WatchNetworks(ctx, cursor)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, err)
+		return
+	}
+
+	// Only return a cursor for the 'watch' operation
+	if cursor != nil {
```
eyakubovich: Always return a cursor

subnet/etcd.go
```diff
@@ -347,36 +390,118 @@ func parseSubnetWatchResponse(resp *etcd.Response) (WatchResult, error) {
...
+func (m *EtcdManager) parseNetworkKey(s string) (string, error) {
+	if parts := m.networkRegex.FindStringSubmatch(s); len(parts) == 2 {
+		return parts[1], nil
+	}
+
+	return "", errors.New("Error parsing Network key")
+}
+
+func (m *EtcdManager) parseNetworkWatchResponse(resp *etcd.Response) (NetworkWatchResult, error) {
+	netname, err := m.parseNetworkKey(resp.Node.Key)
```
eyakubovich: The watch is recursive on something like `/coreos.com/network/`. Wouldn't then this fire on every key change, like subnets getting added? In that case `ParseConfig` will fail and that error propagated.

eyakubovich:
@dcbw I know it's a long review but I think this PR getting very close.

dcbw:
> I know it's a long review but I think this PR getting very close.

Yeah, I've been trying to track down the remote testcase hang before replying, not there yet.  Appears to happen about 1/5 the time on the Travis CI machines:

E0818 18:58:20.993249 02590 mock_registry.go:286] #### mockreg::CreateNetwork() sent add event; events (chan *client.Response)(0xc2080a2000)
E0818 18:58:20.993737 02590 remote_test.go:202] #### test::doTestWatchNetworks() added network
E0818 18:58:20.993757 02590 remote_test.go:207] #### test::doTestWatchNetworks() waiting for Add event
E0818 18:58:20.993922 02590 server.go:189] #### server::handleNetworks() start cursor "4"
E0818 18:58:20.993731 02590 mock_registry.go:188] #### mockreg::watch() got event &client.Response{Action:"add", Node:(*client.Node)(0xc208004ea0), PrevNode:(*client.Node)(nil), Index:0x0} modidx 5 since 4
E0818 18:58:20.993937 02590 etcd.go:335] #### etcd::WatchNetworks() start cursor "4"
E0818 18:58:20.993963 02590 mock_registry.go:193] #### mockreg::watch() returning event &client.Response{Action:"add", Node:(*client.Node)(0xc208004ea0), PrevNode:(*client.Node)(nil), Index:0x0}
E0818 18:58:20.993994 02590 etcd.go:346] #### etcd::WatchNetworks() about to watch networks nextIndex 0x4
E0818 18:58:20.994034 02590 mock_registry.go:180] #### mockreg::watch() starting watch net  since 4

Since the server-side misses the "add" event it just sits in etcd::WatchNetworks()/mockreg::watch() waiting for something to happen, but nothing does.  The root cause appears to be something still sitting in the MockRegistry's watch() method, which is grabbing the event from the queue before the server starts watching.

Say the cursor gets updated internally to "5" in response to some random event.  Then a client watches networks with an older cursor "4", and we get to server.go::handleNetworks().  That calls into etcd.go::WatchNetworks() and then to mock_registry.go::watch() with cursor "4".  Which just sits there waiting for some new event which (in the case of the remote tests) never comes.

Shouldn't something on the server-side return a snapshot when the client-provided cursor is less than the registry cursor?  Something like this perhaps?

func (msr *MockSubnetRegistry) watch(ctx context.Context, network string, since uint64) (*etcd.Response, error) {
	for {
+		if since < msr.index {
+			return nil, etcd.Error{
+				Code: etcd.ErrorCodeEventIndexCleared,
+				Cause: "out of date",
+				Message: "cursor is out of date",
+				Index: msr.index,
+			}
+		}

		select {
		case <-ctx.Done():

subnet/etcd.go
dcbw: > The watch is recursive on something like /coreos.com/network/. Wouldn't then this fire on every key change, like subnets getting added? In that case ParseConfig will fail and that error propagated.

Yes, it should be recursive, since we consider the 'config' key authoritative; if it's missing the network is dead to us.  I've fixed that up now to ignore sub-keys and not return an error.

dcbw:
> Shouldn't something on the server-side return a snapshot when the client-provided cursor is less than the registry cursor? Something like this perhaps?

So I added this and pushed it, and it seems to pass the testcases in Travis now much more reliably.  It seems like the correct thing to do, but your input would be appreciated here.  Thanks!

Repushed, hopefully addressing all your previous comments.

eyakubovich:
LGTM. Thank you for your work.

Issue #126, opened 2015-02-11
flannel falling behind etcd causes spurious message


eyakubovich:
```
eb 09 17:39:53 b7.dev.g8teway.com flanneld[7224]: W0209 17:39:53.562142 07224 subnet.go:410] Watch of subnet leases failed because etcd index outside history window
Feb 09 17:39:53 b7.dev.g8teway.com flanneld[7224]: I0209 17:39:53.562497 07224 udp.go:228] Subnet removed: 10.244.34.0/24
```

Subnet removed refers to its own subnet and this should be handled properly as to not print that line.

philipphug:
I have the same issue and it often results in losing connectivity between the hosts. I need to restart flanneld to restore connectivity again.

eyakubovich:
@philipphug I was not aware of this issue causing connectivity loss. This is with default "udp" backend, correct? Does this cause you to lose connectivity between all hosts or just some?

lemenkov:
I personally sometimes see this message ("Watch of subnet leases failed because etcd index outside history window") as well. But so far it never resulted in a connectivity issues. I'm using "UDP" backend if it matters.

philipphug:
The problem I experienced was that the subnet entries (/coreos.com/network/subnets/XXXX-24) were overwritten with an empty value for some reason which then caused the connectivity issues.
This happened mostly when there was too much load on the etcd machines which caused timeouts and leader elections.

eyakubovich:
Is this still an issue? Needs confirmation.

eyakubovich:
Should be fixed by https://github.com/coreos/flannel/pull/286

PR #286, opened 2015-08-13
Ignore own lease in subnet events

### LINKED ISSUE #241: Watch should not fire extraneous events
If there's a SubnetAdded event but flannel already knows about it, it should be suppressed. WatchLeases() already has a local cache of known subnets so filtering out should be simple.

eyakubovich:
Add an option to ignore a lease in WatchLeases.
Backends can pass their own lease so filter out events
related to them.

Fixes #241 and #126


MohdAhmad:
LGTM

PR #287, opened 2015-08-13
Remove code dup and use coreos/pkg/flagutil

### LINKED ISSUE #278: Replace main.flagsFromEnv with coreos/pkg#SetFlagsFromEnv
I noticed the TODO [0] in main.go. We already pulled this logic out [1], so it should be simple enough to replace it now.

[0] https://github.com/coreos/flannel/blob/master/main.go#L82
[1] https://github.com/coreos/pkg/blob/master/flagutil/env.go#L16

eyakubovich:
Fixes #278


MohdAhmad:
LGTM once rebased.

PR #290, opened 2015-08-15
Add the IPSEC support

MohdAhmad:
IPSEC mode: Tunnel
This means that the entire IP packet is encrypted. We encapsulate using
the host IP addresses therefore we can have a separate IPSEC backend
instead of integrating it with existing backends. Running in tunnel mode
has the added advantage of not leaking the container IPs.

Authentication Mechanism: Pre-shared key (PSK)
For authentication, we use a PSK, generated by flannel and distributed
through etcd which in turn uses client-side certificates for
authentication.

Internet key exchange (IKE): Charon
We use charon daemon from the strongSwan project to negotiate Security
Associations (SAs) between the endpoints. SAs define the security
attributes, used to enable secure communication between endpoints [1].
IKE has two phases, the first establishes a secure channel using the
Diffie-Hellman key exchange and the second phase uses this secure
channel to negotiate the security attributes like, for example the keys
used to encrypt traffic. We use, 4096 bit Diffie-Hellman, AES-256-CBC
for encryption and SHA-256 for integrity. Note IKEv1 is used and phase 1
is run in Main mode.

Configuration
The only configuration parameter required is whether UDP Encapsulation
should be enabled.

### DIFF
--- a/backend/ipsec/handle_charon.go
+++ b/backend/ipsec/handle_charon.go
@@ -1,0 +1,183 @@
+// Copyright 2015 flannel authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipsec
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+
+	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/bronze1man/goStrongswanVici"
+	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
+	"github.com/coreos/flannel/subnet"
+)
+
+type CharonIKEDaemon struct {
+	path string
+}
+
+func NewCharonIKEDaemon(charonPath string) (*CharonIKEDaemon, error) {
+	path, err := exec.LookPath(charonPath)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info("Launching IKE charon path: ", path)
+
+	return &CharonIKEDaemon{
+		path: path,
+	}, nil
+}
+
+func (charon *CharonIKEDaemon) Run() error {
+	cmd := exec.Cmd{
+		Path: charon.path,
+		SysProcAttr: &syscall.SysProcAttr{
+			Pdeathsig: syscall.SIGTERM,
+		},
+	}
+
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
+}
+
+func (charon *CharonIKEDaemon) LoadSharedKey(remotePublicIP, password string) error {
+	var err error
+	var client *goStrongswanVici.ClientConn
+
+	for {
+		client, err = goStrongswanVici.NewClientConnFromDefaultSocket()
+		if err == nil {
+			break
+		} else {
+			log.Error("ClientConnection failed: ", err)
+			log.Infof("Retrying in 1 second ...")
+			time.Sleep(1 * time.Second)
+		}
+	}
+
+	defer client.Close()
+
+	sharedKey := &goStrongswanVici.Key{
+		Typ:    "IKE",
+		Data:   password,
+		Owners: []string{remotePublicIP},
+	}
+
+	err = client.LoadShared(sharedKey)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Loaded shared key for: %v", remotePublicIP)
+	return nil
+}
+
+func (charon *CharonIKEDaemon) LoadConnection(localLease, remoteLease *subnet.Lease, reqID, encap string) error {
+	var err error
+	var client *goStrongswanVici.ClientConn
+
+	for {
+		client, err = goStrongswanVici.NewClientConnFromDefaultSocket()
+		if err == nil {
+			break
+		} else {
+			log.Info("ClientConnection failed: ", err)
+			log.Infof("Retying in 1 second ...")
+			time.Sleep(1 * time.Second)
+		}
+	}
+	defer client.Close()
+
+	childConfMap := make(map[string]goStrongswanVici.ChildSAConf)
+	childSAConf := goStrongswanVici.ChildSAConf{
+		Local_ts:     []string{localLease.Subnet.String()},
+		Remote_ts:    []string{remoteLease.Subnet.String()},
+		ESPProposals: []string{"aes256-sha256-modp4096"},
+		StartAction:  "start",
+		CloseAction:  "trap",
+		Mode:         "tunnel",
+		ReqID:        reqID,
+		//		RekeyTime:     rekeyTime,
+		InstallPolicy: "no",
+	}
+
+	childSAConfName := formatChildSAConfName(localLease, remoteLease)
+
+	childConfMap[childSAConfName] = childSAConf
+
+	localAuthConf := goStrongswanVici.AuthConf{
+		AuthMethod: "psk",
+	}
+	remoteAuthConf := goStrongswanVici.AuthConf{
+		AuthMethod: "psk",
+	}
+
+	ikeConf := goStrongswanVici.IKEConf{
+		LocalAddrs:  []string{localLease.Attrs.PublicIP.String()},
+		RemoteAddrs: []string{remoteLease.Attrs.PublicIP.String()},
+		Proposals:   []string{"aes256-sha256-modp4096"},
+		Version:     "1",
+		KeyingTries: "0", //continues to retry
+		LocalAuth:   localAuthConf,
+		RemoteAuth:  remoteAuthConf,
+		Children:    childConfMap,
+		Encap:       encap,
+	}
+	ikeConfMap := make(map[string]goStrongswanVici.IKEConf)
+
+	connectionName := formatConnectionName(localLease, remoteLease)
+	ikeConfMap[connectionName] = ikeConf
+
+	err = client.LoadConn(&ikeConfMap)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Loaded connection: %v", connectionName)
+	return nil
+}
+
+func (charon *CharonIKEDaemon) UnloadCharonConnection(localLease, remoteLease *subnet.Lease) error {
+	client, err := goStrongswanVici.NewClientConnFromDefaultSocket()
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	connectionName := formatConnectionName(localLease, remoteLease)
+	unloadConnRequest := &goStrongswanVici.UnloadConnRequest{
+		Name: connectionName,
+	}
+
+	err = client.UnloadConn(unloadConnRequest)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Unloaded connection: %v", connectionName)
+	return nil
+}
+
+func formatConnectionName(localLease, remoteLease *subnet.Lease) string {
+	return fmt.Sprintf("%s-%s-%s-%s", localLease.Attrs.PublicIP, localLease.Subnet, remoteLease.Subnet, remoteLease.Attrs.PublicIP)
+}
+
+func formatChildSAConfName(localLease, remoteLease *subnet.Lease) string {
+	return fmt.Sprintf("%s-%s", localLease.Subnet, remoteLease.Subnet)
+}
--- a/backend/ipsec/ipsec.go
+++ b/backend/ipsec/ipsec.go
@@ -1,0 +1,137 @@
+// Copyright 2015 flannel authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipsec
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
+	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
+
+	"github.com/coreos/flannel/backend"
+	"github.com/coreos/flannel/pkg/ip"
+	"github.com/coreos/flannel/subnet"
+)
+
+var CharonPath string
+
+const (
+	defaultCharonPath = "/opt/flannel/libexec/ipsec/charon"
+	passwordLength    = 40
+)
+
+func init() {
+	backend.Register("ipsec", New)
+}
+
+type IPSECBackend struct {
+	sm       subnet.Manager
+	extIface *backend.ExternalInterface
+}
+
+func New(sm subnet.Manager, extIface *backend.ExternalInterface) (backend.Backend, error) {
+	be := &IPSECBackend{
+		sm:       sm,
+		extIface: extIface,
+	}
+
+	return be, nil
+}
+
+func (be *IPSECBackend) RegisterNetwork(ctx context.Context, netname string, config *subnet.Config) (backend.Network, error) {
+	cfg := struct {
+		UDPEncap bool
+	}{}
+
+	if len(config.Backend) > 0 {
+		log.Info("i.config.backend length > 0")
+		if err := json.Unmarshal(config.Backend, &cfg); err != nil {
+			return nil, fmt.Errorf("error decoding IPSEC backend config: %v", err)
+		}
+	}
+
+	attrs := subnet.LeaseAttrs{
+		PublicIP:    ip.FromIP(be.extIface.ExtAddr),
+		BackendType: "ipsec",
+	}
+
+	l, err := be.sm.AcquireLease(ctx, netname, &attrs)
+
+	switch err {
+	case nil:
+
+	case context.Canceled, context.DeadlineExceeded:
+		return nil, err
+
+	default:
+		return nil, fmt.Errorf("failed to acquire lease: %v", err)
+	}
+
+	if CharonPath == "" {
+		CharonPath = defaultCharonPath
+	}
+
+	ikeDaemon, err := NewCharonIKEDaemon(CharonPath)
+	if err != nil {
+		return nil, fmt.Errorf("error creating CharonIKEDaemon struct: %v", err)
+	}
+
+	log.Info("UDPEncap: ", cfg.UDPEncap)
+
+	password, err := GenerateRandomString(passwordLength)
+	if err != nil {
+		return nil, fmt.Errorf("error generating random string: %v", err)
+	}
+
+	err = be.sm.CreateBackendData(ctx, netname, password)
+	if err != nil {
+		return nil, fmt.Errorf("error creating password: %v", err)
+	}
+
+	password, err = be.sm.GetBackendData(ctx, netname)
+	if err != nil {
+		return nil, fmt.Errorf("error getting password: %v", err)
+	}
+
+	return newNetwork(netname, be.sm, be.extIface, cfg.UDPEncap, password, ikeDaemon, l)
+}
+
+func (be *IPSECBackend) Run(ctx context.Context) {
+	<-ctx.Done()
+}
+
+func generateRandomBytes(n int) ([]byte, error) {
+	b := make([]byte, n)
+	_, err := rand.Read(b)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return b, nil
+}
+
+func GenerateRandomString(s int) (string, error) {
+	b, err := generateRandomBytes(s)
+
+	if err != nil {
+		return "", err
+	}
+
+	return base64.StdEncoding.EncodeToString(b), nil
+}
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -66,6 +66,10 @@
 			"Rev": "5c0313fff8cd85670ae4fc996254dd7d66cb4cf7"
 		},
 		{
+			"ImportPath": "github.com/bronze1man/goStrongswanVici",
+			"Rev": "47d13792a9d595f0d5d840fe25eae019cf204e6a"
+		},
+		{
 			"ImportPath": "github.com/coreos/etcd/client",
 			"Comment": "v2.1.1-135-ga1ef699",
 			"Rev": "a1ef699aebbe7ebb4da2f90a42e0f612eec08384"
--- a/Godeps/_workspace/src/github.com/bronze1man/goStrongswanVici/.travis.yml
+++ b/Godeps/_workspace/src/github.com/bronze1man/goStrongswanVici/.travis.yml
@@ -1,0 +1,6 @@
+language: go
+go:
+ - release
+
+script:
+ - go test -v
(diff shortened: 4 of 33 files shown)


MohdAhmad:
This is not meant to be merged. Putting this up to solicit feedback.

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,193 @@
...
+		path:           path,
+		configFilePath: configFilePath,
+	}, nil
+}
+
+func (charon *CharonIKEDaemon) RunCharonIKEDaemon() error {
+	cmd := exec.Command(charon.path)
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
```
eyakubovich: Why set it if not used?

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,193 @@
...
+
+	err := cmd.Run()
```
eyakubovich: `return cmd.Run()`

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,193 @@
...
+	childConfMap := make(map[string]goStrongswanVici.ChildSAConf)
+	childSAConf := goStrongswanVici.ChildSAConf{
+		Local_ts:      []string{localLease.Subnet.String()},
+		Remote_ts:     []string{remoteLease.Subnet.String()},
+		ESPProposals:  []string{"aes256-sha256-modp4096"},
+		StartAction:   "start",
+		CloseAction:   "trap",
+		Mode:          "tunnel",
+		ReqID:         reqID,
+		RekeyTime:     "10m",
```
eyakubovich: Make this a constant

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Unloaded connection: %v", connectionName)
+	return nil
+}
+
+func formatConnectionName(localLease, remoteLease *subnet.Lease) string {
+	return fmt.Sprintf(localLease.Attrs.PublicIP.String() + "-" + localLease.Subnet.String() + "-" + remoteLease.Subnet.String() + "-" + remoteLease.Attrs.PublicIP.String())
```
eyakubovich: You should use either Sprintf fmt string (prefered): `Sprintf("%s-%s-%s-%s", localLease.Attrs.PublicIP, localLease.Subnet, remoteLease.Subnet, remoteLease.Attrs.PublicIP.String())`

or just concatenate: `ocalLease.Attrs.PublicIP.String() + "-" + localLease.Subnet.String() + "-" + remoteLease.Subnet.String() + "-" + remoteLease.Attrs.PublicIP.String()`

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func formatChildSAConfName(localLease, remoteLease *subnet.Lease) string {
+	return fmt.Sprintf(localLease.Subnet.String() + "-" + remoteLease.Subnet.String())
```
eyakubovich: Same as above

backend/ipsec/ipsec.go
```diff
@@ -0,0 +1,391 @@
...
+		BackendType: "ipsec",
+	}
+
+	password := GenerateRandomString(passwordLength)
+	backendAttrs := subnet.BackendAttrs{
+		Password:    password,
+		BackendType: "ipsec",
+	}
+
+	log.Info("Generated password: ", password)
```
pquerna: while this is going into etcd, you probably don't want to log the actual password  -- logs get shipped all kinds of places with different ACLs, often inadequate, on them.

backend/ipsec/ipsec.go
```diff
@@ -0,0 +1,391 @@
...
+
+func (i *IPSECBackend) Name() string {
+	return "ipsec"
+}
+
+func GenerateRandomString(length int) string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+	b := make([]rune, length)
+	rand.Seed(int64(time.Now().Nanosecond()))
```
pquerna: i'd recommend using `crypto/rand.Read` here, seeding your own rand with the current time is a recipe for trouble.

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,185 @@
...
+	childConfMap := make(map[string]goStrongswanVici.ChildSAConf)
+	childSAConf := goStrongswanVici.ChildSAConf{
+		Local_ts:      []string{localLease.Subnet.String()},
+		Remote_ts:     []string{remoteLease.Subnet.String()},
+		ESPProposals:  []string{"aes256-sha256-modp4096"},
+		StartAction:   "start",
+		CloseAction:   "trap",
+		Mode:          "tunnel",
+		ReqID:         reqID,
+		RekeyTime:     rekeyTime,
```
pquerna: I'm not entirely sure with how `goStrongswanVici` the rekey time interacts with the copmonents of strongswan, for example on this page:

https://wiki.strongswan.org/projects/strongswan/wiki/ExpiryRekey
https://wiki.strongswan.org/projects/strongswan/wiki/ExpiryRekey#Formula

Does this mean that `margintime` is still defaulting to `9m` with the random fuzz addition... so basically as soon as the SA is setup, it's gonna be rekeyed?

backend/ipsec/ipsec.go
eyakubovich: @pquerna Great catch. Not sure how we missed it.

eyakubovich:
@gtank PTAL

backend/ipsec/handle_charon.go
MohdAhmad: Good point. Instead of `margintime`, Vici defines `rand_time` which is the time range from which to choose a random value to subtract from rekey times. It defaults to `10%` of `rekey_time`. So in this case the rekey process could begin anytime after `9m` and before `10m` of the SA being established. We plan to change the `rekey_time` to the default of `1h` before merging. 

(the strongswan website isn't currently working so sharing the github page)
https://github.com/strongswan/strongswan/blob/9322e5b398efcf0a6f3bf576ef4b4b12b5ae6528/src/swanctl/swanctl.opt

eyakubovich:
@eparis Can you have someone on your security team gives us feedback on the approach?

eyakubovich:
@mjg59 PTAL

eparis:
The first comment I got back:

- They may wish to use aes_gcm128 and not aes256. That will give you a lot more bang for our buck and you'll be able to run 100x more traffic on the same hardware, see: https://libreswan.org/wiki/Benchmarking_and_Performance_testing (it's really kernel benchmarking for IPsec, so it applies to any userland used)
- Quick skim of their patch seems to look that they are adding their own xfrm policies
into the kernel outside of an IKE daemon? At least based on my reading of the use of
netlink.XfrmPolicyAdd() and netlink.AddXFRMState(). Seeing things like generateRandomBytes()
in this code also does not look right.

Still quite a bit of discussion around the whole concept, but that was the most specific feedback I got.

npmccallum:
It would be helpful if PSK was not the only method. IKE supports multiple methods and we are actively working on Kerberos support. This would have multiple benefits for the current scenario. For instance, a compromise of a single host does not entail a compromise of all hosts. Additionally, transient connections can be supported via the KDC, including opportunistic encryption.

pquerna:
@npmccallum: well, all of the flannel nodes share access to an etcd.... so doing something besides PSK for that reason isn't highly valuable as it might be otherwise.

npmccallum:
@pquerna: That seems somewhat orthogonal to me.

eyakubovich:
@npmccallum The idea is that the authentication is delegated to etcd (etcd supports TLS client certs and username/password auth). Connecting to etcd is used to "join the cluster" and once you're inside the cluster, you don't need to do further authentication. Notice the even the PSK is generated by flannel and not the user. If Kerberos based auth is needed, it should be added to etcd. This way we're not creating a separate set of creds and auth methods for flannel.

The actual encryption keys are generated via Diffie-Hellman for each host pair so compromising a single host will only allow you to initiate a conversation with any other host, not eavesdrop on other flows.

gtank:
@eparis My thought on the choice of cipher was that the underlying CTR implementation is more likely to be solid than the GCM, but I don't have specific knowledge here- do your people? I agree that the somewhat marginal security gains of 256-bit aren't worth that performance hit.

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,185 @@
...
+
+	log.Info("Charon path: ", path)
+
+	return &CharonIKEDaemon{
+		path:           path,
+		configFilePath: configFilePath,
+	}, nil
+}
+
+func (charon *CharonIKEDaemon) RunCharonIKEDaemon() error {
```
eyakubovich: stylistic nit: I would rename func name to `RunDaemon` or maybe even better to just `Run`

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,185 @@
...
+	var err error
+	var client *goStrongswanVici.ClientConn
+
+	for {
+		client, err = goStrongswanVici.NewClientConnFromDefaultSocket()
+		if err == nil {
+			break
+		} else {
+			log.Info("ClientConnection failed: ", err)
+			log.Infof("Retying in 1 second ...")
```
eyakubovich: typo: `Retying` to `Retrying`

backend/ipsec/handle_charon.go
```diff
@@ -0,0 +1,185 @@
...
+func (charon *CharonIKEDaemon) LoadSharedKey(remotePublicIP, password string) error {
+	var err error
+	var client *goStrongswanVici.ClientConn
+
+	for {
+		client, err = goStrongswanVici.NewClientConnFromDefaultSocket()
+		if err == nil {
+			break
+		} else {
+			log.Info("ClientConnection failed: ", err)
```
eyakubovich: This should be `log.Error`

npmccallum:
@eyakubovich I understand. There are two modes here: delegated and parallel. If IPSec delegates to etcd, then if etcd is deployed without encryption IPSec gives the illusion of security while in reality there is none. There is also the general problem of delegating authentication from OSI level 3 to OSI level 7. This should be inverted. If you are going to delegate authentication, you should delegate the other direction (etcd should delegate to IPSecc).

A better alternative would be to do parallel authentication. That is, use the same X.509 cert for both etcd and IPSec. There is no cryptographic harm in doing this. It also means that the same certificate rotation can be used for both services. This also has other technical benefits (including the perpetual benefit of simplicity). This also permits other forms of host-based identity to be integrated at a later point (such as the aforementioned Kerberos).

Using PSK looks like a nice simple solution at first, but going down that road will be fraught with difficulty in the future.

Also, without a proper IKE, how do you plan to rotate the key-material in long-lived connections?

eyakubovich:
@npmccallum I agree that having etcd deployed without encryption is a real issue. We were planning to do a check and bail if etcd communication is at least not over TLS. I would argue however that even if etcd is not doing auth, not all of security goes out the window. Knowing PSK does not allow for eavesdropping, just an ability to initiate new flows at L3. But auth maybe done at an application level as well.

We do have an IKE daemon running that will re-key long lived connections.

dcbw:
@eyakubovich we discussed this internally yesterday with Nathan and others and so I'll dump some more comments in.

This solution is already requiring X.509, implying that they must exist on the system before flannel starts.  So why not use those certificates with the IPSec daemon and get host <-> host authentication and encryption via the CA certificate and the CRL?  The *Swan daemon (at least libreswan, openswan, and freeswan; I have no clue about StrongSwan) can already use certificates to handle all this, and no static PSK in etcd is required.  That way we're using one source for everything here instead of creating/deplying certificates *and* setting a PSK somewhere.

In any case, if you still think static PSK is a good idea, we (Red Hat/OpenShift) really want to ensure that the authentication type (eg, PSK or X.509) can be specified with flannel, so that we can do the X.509 thing in the future anyway.

eyakubovich:
@dcbw I am trying to understand your main hesitations about delegating authentication to etcd. The main advantage is separation of concerns and ease of use.

But let me respond to your specific concerns:
- gaining host<->host authentication: if you use the same certs as etcd, wouldn't you get the same set of nodes that can access any given host? What are you trying to gain in host<->host auth?
- CRL: this is valid but this should be done at etcd level and propagated to out. So if you want to revoke a cert (node), you revoke it from accessing etcd and regeneration of PSK so it won't be able to auth when re-keying. This still needs to be implemented.
- deploying certs and setting PSK: the PSK is generated by flannel and is an impl detail -- it does not have to be distributed so there's no administration overhead.
- Nathaniel's concern about delegating L3 auth to L7: IKE is also an L7 protocol :)

dcbw:
>     gaining host<->host authentication: if you use the same certs as etcd, wouldn't you get the same set of nodes that can access any given host? What are you trying to gain in host<->host auth?

Yes, you get the same set of nodes.  But the authentication would happen before any access to etcd, since the L3 communication channel between the node and host would be authenticated long before etcd is even in the picture.  So you don't just gain protection for etcd, you gain protection for all traffic that might happen before accessing etcd.

Again, we already have the certificates and can secure communication before talking to etcd, so why wouldn't we want to use them?

>     CRL: this is valid but this should be done at etcd level and propagated to out. So if you want to revoke a cert (node), you revoke it from accessing etcd and regeneration of PSK so it won't be able to auth when re-keying. This still needs to be implemented.

If you use certificates to secure communication before talking to etcd, then we don't need to do any of these steps, because the etcd node (and all other nodes) will simply refuse to talk to any node using a revoked certificate.  We don't even need to rekey anything or regenerate the PSK, because none of the nodes will accept traffic from the revoked node.  This also means there would be no network hiccups when revoking a cert, because the the PSK doesn't need to be regenerated and redistributed to all the nodes.

>     deploying certs and setting PSK: the PSK is generated by flannel and is an impl detail -- it does not have to be distributed so there's no administration overhead.

Isn't the PSK distributed to nodes by etcd?  I guess I just mean administration overhead when doing revocations.  With a PSK you need to (a) distribute the CRL to the etcd node (and possibly to all nodes) and then change the PSK to kick the old node off.  When using certs, you just have the CRL step (albeit to all nodes).  That can all probably be scripted, but it's an additional step.

>     Nathaniel's concern about delegating L3 auth to L7: IKE is also an L7 protocol :)

Yes; but I feel like I'm not adequately expressing why I (we) think the certificate-not-PSK path is better/simpler.  Forgive me if this sounds like a broken record...

The current proposal is:

1) use node certificates to talk to etcd via TLS to get the PSK
2) send that PSK to the IKE daemon and start talking to other hosts
3) start doing normal flannel stuff; reading networks, etc

A pure certificate proposal is:

1) send node certificate to the IKE daemon and start talking to other hosts
2) start doing normal stuff; reading networks, etc

Just using certificates means we don't have to bother with any of the etcd PSK handling.  Yes that's automated, but ISTM doing pure certificates does the following:

1) cuts out a couple round-trips in the architecture to go to etcd and get the PSK
2) cuts out required code in flannel to set/get the PSK in both the local and remote cases
3) removes one point of failure (anything to do with PSK retrieval)
4) secures host communication from the start and doesn't actually require etcd+TLS in the end

To me, PSK just seems like more work, and I don't get the benefit?  I could be wrong, but I don't think it's any simpler or less overhead than certs we're already installing and using?  Honestly interested in the advantages you see from the PSK scheme, probably I just don't realize what they are...

(also, I'm happy to pitch in here, not just spout hot air)

dcbw:
@npmccallum can you double-check my understanding of the pure cert method?  Hopefully I got it all right based on the conversation Tuesday.

eyakubovich:
@dcbw The biggest simplicity is conceptual -- you authenticate just via etcd (which you need to do anyway). Today that is done via certs but tomorrow it might be via OAuth token. And if that happens, you will first need to distribute the token for etcd authentication and then distribute the cert for IPSec/IKE.

Also, I want to clarify that using certs for IKE will not secure the communication any sooner. We're talking about a flannel backend. Before the backend becomes functional, it has to communicate with etcd (over plain IP, not IPsec).

pwnall:
@MohdAhmad If you need any encouragement, I could have really used this today! Best of luck finishing it up!

eyakubovich:
@npmccallum @dcbw I want to discuss and understand your case for using certs. Sorry, I am still not convinced. Is there a way we could do a Hangout to discuss in person? /cc @MohdAhmad

dcbw:
> I want to discuss and understand your case for using certs. Sorry, I am still not convinced. Is there a way we could do a Hangout to discuss in person? /cc @MohdAhmad

I'll poke @npmccallum to see if we can set something up.

npmccallum:
I'm game! @simo5 should probably come as well.

dcbw:
@npmccallum @eyakubovich @simo5 @MohdAhmad How about tomorrow (9/10) at 16:00 UTC (12:00 noon US EDT)?

dcbw:
@eyakubovich @MohdAhmad also I'd personally like to see the ipsec stuff as an internal package or library that can be used from *any* backend, not a backend on its own.  I'm sure it would be useful to secure routes for many of the backends (UDP, vxlan, ovs, whatever) via IPSec instead of having those backends be mutually exclusive with the ipsec backend.

simo5:
In meetings all morning till 12:30, maybe 1pm EST (17:00 UTC) ?

simo5:
@gtank AES GCM means authenticated encryption, and its faster than CTR + MAC, it is recommended in any case where you exchange ephemeral session keys and have and can use sequence numbers.

dcbw:
> In meetings all morning till 12:30, maybe 1pm EST (17:00 UTC) ?

@simo5 I'm fine with that.

npmccallum:
I can do 1pm EST.

dcbw:
1pm EST works for me too

dcbw:
Hangout link: https://talkgadget.google.com/hangouts/_/g4jytwjgja3yepse7jaqebaqjea

dcbw:
@eyakubovich @MohdAhmad any chance you could join us for a chat now?

dcbw:
Thinking through how we'd most likely utilize IPSec, there are a few pieces:

1) talking to etcd - flannel is passed the IP address of the etcd master on the command line, so it's trivial to set up an IPSec tunnel between the node and etcd using certificates
2) talking to a master flannel - same thing as (1) here, we know the IP address of the master so it's easy to set up an IPSec tunnel between them
3) talking to other nodes - here we *do* need to talk to etcd to figure out what other nodes are in the cluster; once we read the list of networks and leases (which include the IP address of the lessee) we'd create an IPSec tunnel for that node

However, this could all be done outside flannel, if the management plane (eg, openshift/kube) created the IPSec links.  Since openshift/kube are already watching for other nodes it could create the IPSec tunnels outside of flannel, and flannel could simply be used in unencrypted/non-TLS mode.  One problem here is that flannel and openshift/kube startup is often controlled by an init system like systemd, and ensuring that openshift/kube was started *and* had set up the IPSec tunnel to etcd before flannel starts would not be simple.

diranged:
For what its worth, I actually agree with the model of using the already-generated TLS/SSL certs ... _if Flanneld detects changes in the certs automatically without restarting_. I didn't see this point discussed above, so I'm curious what your thoughts are on it.

ibotty:
What's the advantage on having it inside flannel again?

It's reasonable easy to set up an ipsec tunnel with the flannel subnets and the only thing missing is a (separate) daemon that watches flannel's etcd, generates the right ipsec subnet config and reloads the ipsec stack. That way, before flannel starts, there already is host-to-host ipsec. It's also very easy to support different ipsec auth schemes, you just need to generate the config a little different.

dcbw:
> For what its worth, I actually agree with the model of using the already-generated TLS/SSL certs ... if Flanneld detects changes in the certs automatically without restarting. I didn't see this point discussed above, so I'm curious what your thoughts are on it.

@diranged I don't think flanneld currently doesn't detect cert changes for its etcd communication or for the flannel remote capability.  So unless those parts get updated to handle cert changes along with the ipsec code, then you'd be left with working IPSec but completely broken etcd/remote still.  Doesn't seem helpful to me.

But I agree that having flannel detect this (through inotify perhaps) and handle it is a great long-term goal...

dcbw:
> What's the advantage on having it inside flannel again?

@ibotty the advantage is that flannel is "the thing that controls node networking", and whether the node communications is encrypted is arguably something that flannel should also handle.  Granted, flannel isn't the only thing that benefits from having IPSec enabled to all other nodes in the cluster.  But having this handled in one place, where networking is done already, seems like the right place.  And if it was made into a library then all the backends could use it and it could become an option like 'IpMasq' currently is, perhaps.  Lastly, flannel is already watching for other nodes (via etcd) and that's the exact event that you'd use to set up/tear down an IPSec tunnel to that node.

The argument (as I see it) for having it outside of flannel is that technically, the IPSec operation really doesn't have anything to do with flannel and if it was configured before flannel, then flannel could be run un-encrypted over the encrypted link just like anything else on the node.  IPSec could be configured outside flannel, at node setup time, and then whatever is actually spawning containers on the node probably is already listening for other nodes, and it could do the required inter-node IPSec tunneling.

@eyakubovich I'm still leaning more towards having this in flannel itself.  That avoids startup race conditions, flannel is already watching for other nodes, already knows the certificates, already knows the etcd server address (or the flanneld master address), etc.  Plus, maybe the the orchestrator on the node doesn't actually do anything network-related (openshift is possibly special here)...

ibotty:
Generating and activating the ipsec subnet config in an internal flannel hook is reasonable. Even better if that happens before the subnet is routable.My point is that ipsec should not happen inside of flannel but with an external ipsec stack. That way: * it won't interfere with the host-to-host ipsec that might already exist; * It's also one less piece of software to audit and  * it allows for different ipsec stacks (future linux or bsd's)coreos/flannel wrote:
What's the advantage on having it inside flannel again?


@ibotty the advantage is that flannel is "the thing that controls node networking", and whether the node communications is encrypted is arguably something that flannel should also handle.  Granted, flannel isn't the only thing that benefits from having IPSec enabled to all other nodes in the cluster.  But having this handled in one place, where networking is done already, seems like the right place.  And if it was made into a library then all the backends could use it and it could become an option like 'IpMasq' currently is, perhaps.  Lastly, flannel is already watching for other nodes (via etcd) and that's the exact event that you'd use to set up/tear down an IPSec tunnel to that node.

The argument (as I see it) for having it outside of flannel is that technically, the IPSec operation really doesn't have anything to do with flannel and if it was configured before flannel, then flannel could be run un-encrypted over the encrypted link just like anything else on the node.  IPSec could be configured outside flannel, at node setup time, and then whatever is actually spawning containers on the node probably is already listening for other nodes, and it could do the required inter-node IPSec tunneling.

@eyakubovich I'm still leaning more towards having this in flannel itself.  That avoids startup race conditions, flannel is already watching for other nodes, already knows the certificates, already knows the etcd server address (or the flanneld master address), etc.  Plus, maybe the the orchestrator on the node doesn't actually do anything network-related (openshift is possibly special here)...

dcbw:
> Generating and activating the ipsec subnet config in an internal flannel hook is reasonable. Even better if that happens before the subnet is routable.My point is that ipsec should not happen inside of flannel but with an external ipsec stack. That way: * it won't interfere with the host-to-host ipsec that might already exist; * It's also one less piece of software to audit and * it allows for different ipsec stacks (future linux or bsd's)coreos/flannel wrote: What's the advantage on having it inside flannel again?

@ibotty Oh, I think I misunderstood your earlier comment.  In this PR and in our discussion, I don't think anyone is saying flannel should be doing a lot of IPSec by itself.  This PR only adds the XFRM policy in the kernel in a few LoC and then shells out to an external IKE to handle the key rotation.  I suppose the patch could instead do everything through the IKE daemon (xSwan or racoon or whatever) but those parts are obviously going to be implementation specific.

Most of my comments have been about the general architecture, where the original key/auth material come from, and where the IPSec *setup* happens (outside or inside flannel, regardless of whether it touches the kernel directly or uses a xSwan), not about the specific implementation.  Mainly because I think the current implementation could be generalized so we could use it in our backend too :)

philips:
@MohdAhmad do you have any progress updates on this?

israelshirk:
Hey there, any updates?  Thanks!

feliksik:
Did not look at this in detail, but note that strongswan has an IKEv2 api, take a look at Vici + charon if you care!

abruehl:
+1 I have some FDA/HIPAA requirements mandating 100% over the wire encryption.  Encryption between hosts would mean I don't need to  worry about implementing TLS in every service.  I would only have to worry about code-level (TLS) encryption when it leaves the flannel network.

Calpicow:
+1--really looking forward to a global, turn-key encryption solution that satisfies Kubernetes network requirements

eyakubovich:
Closing in favor of https://github.com/coreos/flannel/pull/516

PR #292, opened 2015-08-18
Add/remove networks when registry changes

dcbw:
Add a new -watch-networks option to watch for registry (either
remote or local) changes and add/remove networks for those events.
If this option is given along with -networks, only the networks
specified in -networks will be acted upon.  If -networks is
empty, then all network events will be acted upon.

### DIFF
--- a/network/manager.go
+++ b/network/manager.go
@@ -1,0 +1,306 @@
+// Copyright 2015 CoreOS, Inc.
+// Copyright 2015 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package network
+
+import (
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/coreos/go-systemd/daemon"
+	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
+	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
+
+	"github.com/coreos/flannel/backend"
+	"github.com/coreos/flannel/pkg/ip"
+	"github.com/coreos/flannel/subnet"
+)
+
+type CmdLineOpts struct {
+	publicIP      string
+	ipMasq        bool
+	subnetFile    string
+	subnetDir     string
+	iface         string
+	networks      string
+	watchNetworks bool
+}
+
+var opts CmdLineOpts
+
+func init() {
+	flag.StringVar(&opts.publicIP, "public-ip", "", "IP accessible by other nodes for inter-host communication")
+	flag.StringVar(&opts.subnetFile, "subnet-file", "/run/flannel/subnet.env", "filename where env variables (subnet, MTU, ... ) will be written to")
+	flag.StringVar(&opts.subnetDir, "subnet-dir", "/run/flannel/networks", "directory where files with env variables (subnet, MTU, ...) will be written to")
+	flag.StringVar(&opts.iface, "iface", "", "interface to use (IP or name) for inter-host communication")
+	flag.StringVar(&opts.networks, "networks", "", "run in multi-network mode and service the specified networks")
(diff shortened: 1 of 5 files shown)


network/manager.go
```diff
@@ -0,0 +1,296 @@
...
+				if !m.isNetAllowed(netname) {
+					continue
+				}
+
+				switch e.Type {
+				case subnet.EventAdded:
+					if _, ok := m.networks[netname]; ok {
+						continue
+					}
+				log.Warningf("Adding network %v", netname)
```
eyakubovich: gofmt

network/manager.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	// rename(2) the temporary file to the desired location so that it becomes
+	// atomically visible with the contents
+	return os.Rename(tempFile, path)
+}
+
+func (m *Manager) RunNetwork(net *Network, wg sync.WaitGroup) {
+	wg.Add(1)
+	defer wg.Done()
```
eyakubovich: Stylistic nit: so far I have been following the convention of putting the wait group's add/done calls outside:

```
go func() {
    wg.Add(1)
    defer wg.Done()
    nm.RunNetwork(net)
}
```

I'm not sure if it's "the right way" but would like to keep it for consistency.

network/manager.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+
+				log.Warningf("Running network %v", net.Name)
+		net.Run()
+		log.Infof("%v exited", net.Name)
+	}
+}
+
+func (m *Manager) watchNetworks(ctx context.Context, wg sync.WaitGroup) {
+	wg.Add(1)
```
eyakubovich: Same comment as above.

network/manager.go
eyakubovich: I just updated the comment above to reflect that `wg.Add(1)` should be before starting the goroutine to avoid a race.

eyakubovich:
@dcbw Apologies for slow review -- away at ContainerCon and such.

dcbw:
> @dcbw Apologies for slow review -- away at ContainerCon and such.

No problem; I was on vacation last week anyway.

network/manager.go
dcbw: > gofmt. Also, make these Infos instead of Warnings?

That was actually a left-over from debugging that I've since removed.

dcbw:
Rebased and repushed with fixes for comments...

network/network.go
```diff
@@ -26,30 +26,36 @@ import (
...
+
+	sm         subnet.Manager
+	ipMasq     bool
+	be         backend.Backend
+	lease      *subnet.Lease
 }
 
-func New(sm subnet.Manager, name string, ipMasq bool) *Network {
+func NewNetwork(sm subnet.Manager, name string, ipMasq bool) *Network {
+	ctx, cancel := context.WithCancel(context.Background())
```
eyakubovich: I recently realized that a slicker way is to wrap ctx with another `WIthCancel` (https://godoc.org/golang.org/x/net/context#WithCancel). This way you get another `cancel` function but the outer one still works. This way you don't have to:

```
<-ctx.Done()
for _, net := range m.networks {
    net.Cancel()
}
```

network/manager.go
```diff
@@ -0,0 +1,302 @@
...
+			daemon.SdNotify("READY=1")
+		}
+
+				log.Warningf("Running network %v", net.Name)
+		net.Run()
+		log.Infof("%v exited", net.Name)
+	}
+}
+
+func (m *Manager) watchNetworks(ctx context.Context, wg sync.WaitGroup) {
```
eyakubovich: passing `WaitGroup` by value actually copies it and I think will cause trouble. Just create a new `WaitGroup` here and wait for it before returning from this function.

network/network.go
dcbw: Oh nice.  Done.

dcbw:
Rebased on backend-split-init, updated for your comments, and a further fix to the previous getNetworks() functionality to match what etcd actually returns.

eyakubovich:
@dcbw Can you rebase again on top of master (just merged your singleton PR). This will make it easier to review.

dcbw:
@eyakubovich will do.  I'm working on a few fixups first though that were impacting further testing of this branch.  Hope to get those and another push of this branch done today.

network/manager.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+		} else {
+			if err := writeSubnetFile(opts.subnetFile, net.Config.Network, m.ipMasq, sn); err != nil {
+				log.Warningf("%v failed to write subnet file: %s", net.Name, err)
+				return
+			}
+			daemon.SdNotify("READY=1")
+		}
+
+				log.Warningf("Running network %v", net.Name)
```
eyakubovich: `log.Infof` and also fix indent / run gofmt

eyakubovich:
@dcbw LGTM. Will merge after that indent fix.

dcbw:
Repushed with fixups; also added network package to 'test' to get gofmt checking.  Should the 'gofmt -l' be changed to 'gofmt -d' to display diffs?  That's been useful to me in the past.

PR #298, opened 2015-08-28
Refactoring: single ctx and pull out LeaseRenewer

eyakubovich:
- Use a single copy of ctx and plumb it throughout.
This fixes the left over artifact that ctx went into
flannel in stages.

- Backends being responsible for lease renewal is
not really valuable and just leads to duplication.


MohdAhmad:
LGTM

PR #299, opened 2015-08-31
Bug fix: remote mode errors out with bad backend type

eyakubovich:
Due to code reorg, the config must now be parsed via
subnet.ParseConfig in remote client.


MohdAhmad:
LGTM

PR #304, opened 2015-09-02
Add support for singleton backends

dcbw:
We'd like to implement our multitenant OVS backend with a singleton object to set up and create the OVS bridge (we only need one bridge for all tenant networks) and maintain some host-wide flows.  To decrease the complexity of a singleton backend we can very slightly increase the complexity of non-singleton backends by splitting Init() into:

1) Init() which takes properties that won't ever change (ctx, interface properties, etc)
2) AddNetwork() which tells the backend to actually configure itself for the network

A singleton backend will now only be created once, and have its Init() only called once.  After that, each new Network object will trigger an AddNetwork() call to poke the singleton backend to do whatever it needs to do for the new network.

(once the add/remove networks branch lands, of course we'll also have a RemoveNetwork() call into the backend, which non-singleton backends would ignore...)

### DIFF
--- a/backend/alloc/alloc.go
+++ b/backend/alloc/alloc.go
@@ -11,30 +11,33 @@ import (
 )
 
 type AllocBackend struct {
-	sm      subnet.Manager
-	network string
-	lease   *subnet.Lease
+	sm       subnet.Manager
+	publicIP ip.IP4
+	mtu      int
+	lease    *subnet.Lease
 }
 
-func New(sm subnet.Manager, network string, config *subnet.Config) backend.Backend {
-	return &AllocBackend{
+func New(sm subnet.Manager, extIface *net.Interface, extIaddr net.IP, extEaddr net.IP) (backend.Backend, error) {
+	be := AllocBackend{
 		sm:      sm,
-		network: network,
+		publicIP: ip.FromIP(extEaddr),
+		mtu:      extIface.MTU,
 	}
+	return &be, nil
 }
 
-func (m *AllocBackend) Init(ctx context.Context, extIface *net.Interface, extIaddr net.IP, extEaddr net.IP) (*backend.SubnetDef, error) {
+func (m *AllocBackend) RegisterNetwork(ctx context.Context, network string, config *subnet.Config) (*backend.SubnetDef, error) {
 	attrs := subnet.LeaseAttrs{
-		PublicIP: ip.FromIP(extEaddr),
+		PublicIP: m.publicIP,
 	}
 
-	l, err := m.sm.AcquireLease(ctx, m.network, &attrs)
+	l, err := m.sm.AcquireLease(ctx, network, &attrs)
 	switch err {
 	case nil:
 		m.lease = l
 		return &backend.SubnetDef{
 			Lease: l,
-			MTU:   extIface.MTU,
+			MTU:   m.mtu,
 		}, nil
 
 	case context.Canceled, context.DeadlineExceeded:
--- a/backend/alloc/alloc.go
+++ b/backend/alloc/alloc.go
@@ -16,7 +16,7 @@ type AllocBackend struct {
 	lease   *subnet.Lease
 }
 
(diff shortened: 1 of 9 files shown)


eyakubovich:
I don't like the duality of a backend being a singleton or not, considering there's always `AddNetwork` method. I think either all backends should be singletons or none of them. If backends are not singletons, it should be pretty easy to just use global variables in the concrete backend package (e.g. `ovs`) to keep track of any global data.

The reason there's currently both `New` and `Init` is that the former does not take `ctx` and is not interruptable and the latter is. If you want the backend to be a singleton, I would just combine `New` and `Init` as both have nothing interruptable. I would then have `AddNetwork` and `Run` take `ctx` as first arg (convention established by context lib authors).

dcbw:
Ok, fair enough.  How about now?

eyakubovich:
I think the `Backend` interface is good now. However the implementations need to make sure that any data that is network specific is stored in some map (mapping from network name to network specific data). Stuff like lease, routes, UDP sockets, TUN devices, etc. Otherwise `AddNetwork` calls will stump over each other. There also needs to a mutex there to protect it as I understand the calls to AddNetwork are not serialized.

This is why it maybe easier to have each backend package just expose something like:
```
func NewNetwork(ctx context.Context, sm subnet.Manager, network string, config *subnet.Config, iface *Iface) Network`

// need a better name though
type Iface struct {
    extIface *net.Interface
    extIaddr net.IP
    extEaddr net.IP
}

type Network interface {
    Run(ctx context.Context)
}
```

Data common to all networks in a given backend would be stored in global vars.

I am fine with both designs but I would probably go with the `NewNetwork` one so that I don't have to keep all those maps in each backend. There would just be one map (with a mutex) in the `network` package.

dcbw:
> probably go with the NewNetwork one

Could you clarify which approach you mean here?

eyakubovich:
hehe -- I did leave it ambiguous :) I meant the latter one, having the backend expose a global function `func NewNetwork(ctx context.Context, sm subnet.Manager, network string, config *subnet.Config, iface *Iface) Network`

dcbw:
> hehe -- I did leave it ambiguous :) I meant the latter one, having the backend expose a global function func NewNetwork(ctx context.Context, sm subnet.Manager, network string, config *subnet.Config, iface *Iface) Network

Ah, so you mean having the backend take responsibility for creating the Network objects, rather than the Manager?

dcbw:
> I think the Backend interface is good now. However the implementations need to make sure that any data that is network specific is stored in some map (mapping from network name to network specific data). Stuff like lease, routes, UDP sockets, TUN devices, etc. Otherwise AddNetwork calls will stump over each other. There also needs to a mutex there to protect it as I understand the calls to AddNetwork are not serialized.

I was attempting modify the existing backends as little as possible here, and have singleton backends be opt-in.  So our in-progress OVS plugin uses Once in the New() call to create the singleton backend, and then protects AddNetwork() with a mutex.  So the backend itself handles storing the individual network info internally.

Existing backends still create a new backend instance for each Network object, so there would be no clashes for AddNetwork() in the existing backends with the branch as it currently stands.

eyakubovich:
I am sorry -- I didn't read the changes carefully enough. It's fine then with the exception that `AddNetwork` is now not an ideal name. Unless `New()` does the `Once` fanciness, there's a 1-1 mapping between a backend instance and a network. Therefore `AddNetwork` will be called only once but feels like it will be called many times.

However I think it's ok for now. Once OVS backend is merged, we can revisit the design and refactor it a bit more.

Can you add a comment in `backend/common.go` to explain it a bit -- I think the semantics may trip people up.

dcbw:
Would SetNetwork() or RegisterNetwork() or even NewNetwork() sound better to you?

eyakubovich:
Let's go with `RegisterNetwork`

dcbw:
Updated.

eyakubovich:
LGTM

PR #306, opened 2015-09-04
etcd node and error return handling fixups

dcbw:
The first patch fixes up my previous network watching stuff, where the testcases didn't quite match actual etcd behavior and thus the code was also slightly wrong.

The second patch I have questions on... When the etcd client got switched to etcd/client, it broke cancel handling. Because etcd/client returns ClusterError objects from its HTTP methods, those get passed down to registry.go and etcd.go, and then stuff like:

    case err == context.Canceled, err == context.DeadlineExceeded:

doesn't work because 'err' is a ClusterError, not a context error. How should this get fixed? I had a couple approaches:

1) strip ClusterErrors in registry.go and return only the first internal error; but this doesn't work because etcd.go checks for etcd.Error too.

2) strip ClusterErrors in etcd.go and return only the first internal error; but this doesn't quite work either because callers are looking for only one error, and ClusterError can hold multiple errors

3) create some kind of MatchError() function that knows about ClusterError and etcd.Error internals and does the right thing; this is kinda ugly because it leaks ClusterError semantics out to stuff that shouldn't know about etcd at all (like when the backends check for cancelation of AcquireLease()). I guess this function should go into the subnet.Manager interface and we mandate that all callers of subnet.Manager use it for error checking? This is what I ended up implementing.

Thoughts? The orignal problem I had was that at Ctrl^C/exit time, flannel never exited because watch.go::WatchLeases() never sees the cancel because it's checking for context.Canceled, not ClusterError.Errors[0] == context.Canceled.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -27,8 +27,13 @@
 		},
 		{
 			"ImportPath": "github.com/coreos/etcd/client",
-			"Comment": "v2.1.1-85-gff0b872",
-			"Rev": "ff0b8723c747e76d2651bcddaff26563128ab216"
+			"Comment": "v2.1.1-135-ga1ef699",
+			"Rev": "a1ef699aebbe7ebb4da2f90a42e0f612eec08384"
+		},
+		{
+			"ImportPath": "github.com/coreos/etcd/pkg/pathutil",
+			"Comment": "v2.1.1-135-ga1ef699",
+			"Rev": "a1ef699aebbe7ebb4da2f90a42e0f612eec08384"
 		},
 		{
 			"ImportPath": "github.com/coreos/etcd/pkg/transport",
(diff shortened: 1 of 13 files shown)


xiang90:
@dcbw This might be helpful https://github.com/coreos/etcd/tree/master/client#error-handling

dcbw:
@xiang90 Yeah, that looks helpful, and it turns out that flannel's etcd/client is not up-to-date and does not contain https://github.com/coreos/etcd/commit/a1ef699aebbe7ebb4da2f90a42e0f612eec08384 so I guess we should update it.  Thanks!

eyakubovich:
LGTM except please vendor and add github.com/coreos/etcd/pkg/pathutil

dcbw:
vendored in pkg/pathutil; repushed

PR #307, opened 2015-09-04
switch from glog to Sirupsen/logrus

### LINKED ISSUE #295: Move away from glog logger
glog is overly verbose and always prints a timestamp. The logger should detect if it runs under journald and not log the timestamp in that case.

kdomanski:
Fixes #295 

[`Sirupsen/logrus`](https://github.com/Sirupsen/logrus) is the logger used by [Docker](https://github.com/docker/docker). By default it produces no timestamps, uses colored text on interactive TTY and is generally more concise.


eyakubovich:
I need to think about which log package to go with. `logrus` is definitely an option but there's value in keeping all CoreOS projects using the same. And not there is a preference to move to https://github.com/coreos/pkg/tree/master/capnslog

PR #308, opened 2015-09-08
[Hack] Replace troublesome glog references with log

bprashanth:
Flannel's glog import strategy leads to a double initialization of glog flags when imported from another package that also uses glog. Most (all?) coreos packages use log. I've just replaced the instances I need to use flannel within kubernetes, open to non-hacky suggestions.

### DIFF
--- a/backend/vxlan/vxlan.go
+++ b/backend/vxlan/vxlan.go
@@ -36,11 +36,11 @@ const (
 )
 
 type VXLANBackend struct {
-	sm       subnet.Manager
-	network  string
-	cfg      struct {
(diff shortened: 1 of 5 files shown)


eyakubovich:
@bprashanth We want to move away from glog (it was added for leveled logging) but maybe to https://github.com/coreos/flannel/pull/307 or https://github.com/coreos/pkg/tree/master/capnslog. Would those be satisfactory to you?

bprashanth:
Just grepping the codebase, I think those should work. Will reopen if that isn't the case.

PR #312, opened 2015-09-11
Fix running multiple networks

dcbw:
Caused by my misunderstanding of scope and go funcs.

Fixes: 6b68c43610071f7cb568f601fd4c4ddbe8d710ef

### DIFF
--- a/network/manager.go
+++ b/network/manager.go
@@ -288,10 +288,10 @@ func (m *Manager) Run() {
 	// Run existing networks
 	for _, net := range m.networks {
 		wg.Add(1)
-		go func() {
-			m.RunNetwork(net)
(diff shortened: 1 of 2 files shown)


dcbw:
For the second patch, the issue is that when we added listening for network deletion, that is signified by an empty snapshot, but that was also returned by the etcd watch code when it wanted to ignore a key.  So we need a way for the network watch bits to do that without terminating the watch.

An alternative approach (if you don't like goto) would be to have an internal error that parseNetworkWatchResult could return that watch.go::WatchNetworks() would look for.  Happy to do that too.

eyakubovich:
LGTM

PR #317, opened 2015-09-15
Better handling of Ctrl+C

eyakubovich:
VXLAN should fail RegisterNetwork call as
outer code does the retrying. This causes
SIGINT to be properly processed.


eyakubovich:
@MohdAhmad PTAL

pbx0:
LGTM

PR #321, opened 2015-09-18
Add UnregisterNetwork backend method

dcbw:


eyakubovich:
`UnregisterNetwork` gets executed right after `Run`. Why not do the clean up in `Run`? Is this just for symmetry?

dcbw:
I could move it up to the be.Run() gofunc, if that's what you mean?  eg:

	go func() {
		n.be.Run(n.ctx)
		n.be.UnregisterNetwork(n.ctx, n.Name)
		wg.Done()
	}()

dcbw:
I didn't have it cleaned up in the backends themselves to make it easier for singleton backends, where Run() might be called only once.  For example in our OVS plugin we have:

    func (ovsb *OVSBackend) Run(ctx context.Context) {
	    onceRun.Do(func() { ovsb.start(ctx) })
    }

eyakubovich:
Right, although I just realized that I don't feel too good about our interface. Sometimes `Run()` gets called prior to `RegisterNetwork` and sometimes after. Maybe this would be OK if the implementations did not rely on the ordering but they do! Maybe `Run` should be called prior to any `RegisterNetwork` to force uniformity although then there's very little it can do until the first network gets registered. So perhaps the interface should not have `Run` at all. `RegisterNetwork` can just spin up a goroutine if it need one and shut it down when all UnregisterNetworks are done. Or launch one goroutine per network -- whatever it wants.

Having said that, I'll review the PR as is and wait for OVS backend to get merged before redoing the interface again.

eyakubovich:
Correction to above. Sorry, `Run` get executed for every network but it is not parameterized on network name so it's semantics are really unclear now.

eyakubovich:
LGTM. However the backend interface needs to be reworked in the near future.

PR #326, opened 2015-09-23
Notify systemd service when server is ready to listen

ingvagabund:
When running flannel with -listen option, systemd service is not notified and 'systemctl start flannel' does not return. Adding daemon.SdNotify("READY=1") to RunServer notifies systemd and the command returns.

The correct place for notifying the systemd is line 278 of remote/server.go [1]. It is the line where a listener is created and the execution is just before the server can start listening for incoming requests.

[1] https://github.com/coreos/flannel/blob/master/remote/server.go#L278


eyakubovich:
LGTM

PR #327, opened 2015-09-23
Masquerade host to flannel traffic.

eyakubovich:
Traffic coming from host might have src address not of
flannel interface but of some other one. This happened
in conjunction with REDIRECT rules. Masquerade will
force the src addr to be that of the flannel interface.

Fixes #318

PR #328, opened 2015-09-24
Robustness bug fixes and improvements

eyakubovich:
Fixes bugs related to flannel exiting and networks
being deleted.

### DIFF
--- a/network/ipmasq.go
+++ b/network/ipmasq.go
@@ -24,36 +24,47 @@ import (
 	"github.com/coreos/flannel/pkg/ip"
 )
 
+func rules(ipn ip.IP4Net) [][]string {
+	n := ipn.String()
+
+	return [][]string{
+		// This rule makes sure we don't NAT traffic within overlay network (e.g. coming out of docker0)
+		{"-s", n, "-d", n, "-j", "ACCEPT"},
+		// NAT if it's not multicast traffic
+		{"-s", n, "!", "-d", "224.0.0.0/4", "-j", "MASQUERADE"},
+		// Masquerade anything headed towards flannel from the host
+		{"!", "-s", n, "-d", n, "-j", "MASQUERADE"},
+	}
+}
+
 func setupIPMasq(ipn ip.IP4Net) error {
 	ipt, err := iptables.New()
 	if err != nil {
-		return fmt.Errorf("failed to setup IP Masquerade. iptables was not found")
+		return fmt.Errorf("failed to set up IP Masquerade. iptables was not found")
 	}
 
-	err = ipt.ClearChain("nat", "FLANNEL")
-	if err != nil {
-		return fmt.Errorf("Failed to create/clear FLANNEL chain in NAT table: %v", err)
-	}
-
-	rules := [][]string{
-		// This rule makes sure we don't NAT traffic within overlay network (e.g. coming out of docker0)
-		{"FLANNEL", "-d", ipn.String(), "-j", "ACCEPT"},
-		// NAT if it's not multicast traffic
-		{"FLANNEL", "!", "-d", "224.0.0.0/4", "-j", "MASQUERADE"},
-		// This rule will take everything coming from overlay and send it to FLANNEL chain
-		{"POSTROUTING", "-s", ipn.String(), "-j", "FLANNEL"},
-		// Masquerade anything headed towards flannel from the host
-		{"POSTROUTING", "!", "-s", ipn.String(), "-d", ipn.String(), "-j", "MASQUERADE"},
-	}
-
-	for _, rule := range rules {
+	for _, rule := range rules(ipn) {
 		log.Info("Adding iptables rule: ", strings.Join(rule, " "))
-		chain := rule[0]
-		args := rule[1:len(rule)]
-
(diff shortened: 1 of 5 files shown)


network/ipmasq.go
```diff
@@ -57,3 +57,17 @@ func setupIPMasq(ipn ip.IP4Net) error {
...
 	return nil
 }
+
+func teardownIPMasq(ipn ip.IP4Net) error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("failed to setup IP Masquerade. iptables was not found")
+	}
+
+	err = ipt.DeleteChain("nat", "FLANNEL")
```
MohdAhmad: Isn't a call to `ipt.ClearChain` required before `ipt.DeleteChain`. From the code comments [here] (https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go) I understand that the `DeleteChain` requires that the chain already be empty.

network/ipmasq.go
```diff
@@ -57,3 +57,17 @@ func setupIPMasq(ipn ip.IP4Net) error {
...
+	if err != nil {
+		return fmt.Errorf("Failed to create/clear FLANNEL chain in NAT table: %v", err)
+	}
+
+	return ipt.Delete("nat", "POSTROUTING", "-s", ipn.String(), "-j", "FLANNEL")
```
MohdAhmad: We also need to add delete the `POSTROUTING` rule added in #327 .

network/ipmasq.go
```diff
@@ -57,3 +57,17 @@ func setupIPMasq(ipn ip.IP4Net) error {
...
+
+func teardownIPMasq(ipn ip.IP4Net) error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("failed to setup IP Masquerade. iptables was not found")
+	}
+
+	err = ipt.DeleteChain("nat", "FLANNEL")
+	if err != nil {
+		return fmt.Errorf("Failed to create/clear FLANNEL chain in NAT table: %v", err)
```
MohdAhmad: Please change `Failed` to `failed`

network/ipmasq.go
eyakubovich: Yes, I'll merge #327 first and then I'll update this PR

network/ipmasq.go
eyakubovich: `teardownIPMasq` should only get called if `setupIPMasq` was called first. If `setupIPMasq` failed, `teardownIPMasq` won't get called. Now, if a user/another program ends up deleting `FLANNEL` chain, it will error (I think correctly) but it'll just get logged and go on.

eyakubovich: I see what you mean -- the chain must be empty before deletion. I think you're right, not sure how it seemed to work, I'll investigate.

eyakubovich:
@dcbw PTAL

dcbw:
Fixes look mostly OK to me and I've tested it with the OVS singleton plugin.

The only thing causing concern is the move of UnregisterNetwork() to a defer.  We're currently using this for our Run():

    func (ovsb *OVSBackend) Run(ctx context.Context) {
        onceRun.Do(func() { ovsb.start(ctx) })
    }

which means that no networks actually return from Run() until they all do.

I do like the defer handling for UnregisterNetwork() though, so I'm happy to adapt our Run() to just start our plugin's loop on the first network registration, and then select{} on every network's context.

But there's one missing piece; how does a singleton plugin know to exit?  Networks come and go, and even when the last network is removed, that doesn't mean we should cancel a singleton plugin's event loop.  So now the singleton plugin needs some kind of cancel too...  I'm not sure what to do there, except to pass the Manager's ctx in newBackend() so a singleton can grab it?

eyakubovich:
The way things are going, I think it will make sense to make all backends singletons and modify the interface to: 

```
type Backend interface {
    // Called when the backend should create or begin managing a new network
    RegisterNetwork(ctx context.Context, network string, config *subnet.Config) (*SubnetDef, error)
    // Called to clean up any network resources or operations
    UnregisterNetwork(ctx context.Context, network string)
}
```

eyakubovich:
The way things are going, I think it'll make sense to make every backend a singleton and change the interface to something closer to:

```
type Backend interface {
     // This is called first (before any networks). ctx is cancelled when flannel is exiting
    Run(ctx context.Context)

    // Called when the backend should create or begin managing a new network
    RegisterNetwork(network string, config *subnet.Config) (*SubnetDef, error)

    // Called to clean up any network resources or operations
    UnregisterNetwork(network string)
}
```

When flannel is exiting, it would first unregister all networks and then interrupt the ctx. Or maybe we still pass ctx to every function. But that would clean up that defer logic and I think make the interface less confusing (it is to me now). It would need to be a separate PR. And I have a bunch more of refactoring coming as I continue to work on the revoke lease patch.

eyakubovich:
@dcbw if you'd like, I can put up a PR to change the interface first and then rebase this patch on top.

dcbw:
@eyakubovich Sure, lets see what that looks like.  Maybe it's just easier in the long run.

eyakubovich:
Closing in favor of #338

PR #331, opened 2015-09-28
Change copyright from CoreOS to flannel authors

eyakubovich:
flannel is developed by more than just CoreOS, Inc
and the copyright is retained by all flannel
contributors (for their respective contributed code).

Also, fixed up some gofmt issues.


MohdAhmad:
LGTM

PR #338, opened 2015-10-06
Refactor the backend interfaces for multi-networks

eyakubovich:
Make backends be aware of multiple networks but creating
a separate backend.Network interface.

Also fixes bugs related to flannel exiting and networks
being deleted.


eyakubovich:
This is not yet ready for merge but seeking feedback on the new interfaces.

@dcbw PTAL

dcbw:
@eyakubovich Looks great from 10,000ft; I'll rebase our stuff on top of it today and see if I can shake anything out, especially with the parseNetworkConfig() changes.

dcbw:
One thing I noticed; hostgw has a 'networks' map, but a network never gets removed from the map.  We need to do the same thing in the ovs backend since UnregisterNetwork() is now gone.

I can work around this by passing a channel to each backend network subclass that the backend itself selects on, but there are two problems.  First, when flannel exits and cancels the context, the backend gets canceled before it can read all the events from the "network is canceled" channel and clean things up.  Second, we need to use nonblocking channel operations with select/default.

Seems that UnregisterNetowrk() would be simpler for all backends, but if you don't want it for some reason it's not hard to implement in other ways.

eyakubovich:
@dcbw You're right that there's basically no cleanup in the backends -- that should be improved. My general strategy for cleanup was going to be:

```
type network struct {
    name string
    be *backend
}

func (n *network) Run(ctx context.Context) {
    <-ctx.Done()
    // or use select to detect ctx.Done() getting closed
    n.cleanup()
}

func (n *network) cleanup() {
    be.removeNetwork(n.name)
}
```

`RegisterNetwork` would set a pointer to the backend in the `network` struct.

The reason I don't want `UnregisterNetwork` is because it doesn't play nice with `context.Context` (which, for better or worse, is now throughout flannel). Consider an interface like

```
type Backend interface {
    RegisterNetwork(ctx context.Context, ...) Network
    UnregisterNetwork(ctx context.Context, ...)
}
type Network interface {
    Run(ctx context.Context)
}
```

Let's think though what `UnregisterNetwork` would do. It would need to stop the network which would mean `Run` would exit. But that's the job of `ctx`! And how would you implement it? You would need to pass in some channel or Context into Network so Run could select on it. It's more context idiomatic to just pass it as first arg to functions that need to be interruptable.

BTW, `UnregisterNetwork()` would need to not accept Context as otherwise we have a function whose job is to interrupt but is itself interruptable.

dcbw:
@eyakubovich I'm fine with the be.removeNetwork() strategy; after prototyping that I don't see any problems on my end.

eyakubovich:
@MohdAhmad Can you also take a look please

backend/hostgw/network.go
```diff
@@ -0,0 +1,183 @@
...
+				continue
+			}
+
+			route := netlink.Route{
+				Dst:       evt.Lease.Subnet.ToIPNet(),
+				Gw:        evt.Lease.Attrs.PublicIP.ToIP(),
+				LinkIndex: n.linkIndex,
+			}
+			if err := netlink.RouteAdd(&route); err != nil {
+				log.Errorf("Error adding route to %v via %v: %v", evt.Lease.Subnet, evt.Lease.Attrs.PublicIP, err)
```
MohdAhmad: Change to `error` from `Error`

backend/hostgw/network.go
```diff
@@ -0,0 +1,183 @@
...
+				continue
+			}
+
+			route := netlink.Route{
+				Dst:       evt.Lease.Subnet.ToIPNet(),
+				Gw:        evt.Lease.Attrs.PublicIP.ToIP(),
+				LinkIndex: n.linkIndex,
+			}
+			if err := netlink.RouteDel(&route); err != nil {
+				log.Errorf("Error deleting route to %v: %v", evt.Lease.Subnet, err)
```
MohdAhmad: Change to `error` from `Error`

backend/hostgw/network.go
```diff
@@ -0,0 +1,183 @@
...
+				continue
+			}
+			n.removeFromRouteList(route)
+
+		default:
+			log.Error("Internal error: unknown event type: ", int(evt.Type))
```
MohdAhmad: Change to `internal` from `Internal`

backend/hostgw/network.go
```diff
@@ -0,0 +1,183 @@
...
+				}
+				if routeEqual(r, route) {
+					exist = true
+					break
+				}
+			}
+			if !exist {
+				if err := netlink.RouteAdd(&route); err != nil {
+					if nerr, ok := err.(net.Error); !ok {
+						log.Errorf("Error recovering route to %v: %v, %v", route.Dst, route.Gw, nerr)
```
MohdAhmad: Change to `error` from `Error`

backend/udp/network.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/coreos/flannel/pkg/ip"
+	"github.com/coreos/flannel/subnet"
+)
+
+const (
+	encapOverhead = 28 // 20 bytes IP hdr + 8 bytes UDP hdr
+)
+
+type network struct {
+	backend.SimpleNetwork
```
MohdAhmad: Shouldn't this be `SimpleNetwork backend.SimpleNetwork`?

backend/udp/network.go
```diff
@@ -0,0 +1,204 @@
...
+	return f1, f2, nil
+}
+
+func (n *network) initTun() error {
+	var tunName string
+	var err error
+
+	n.tun, tunName, err = ip.OpenTun("flannel%d")
+	if err != nil {
+		return fmt.Errorf("Failed to open TUN device: %v", err)
```
MohdAhmad: Change to `Failed` from `failed`

backend/udp/network.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// explicitly add a route since there might be a route for a subnet already
+	// installed by Docker and then it won't get auto added
+	err = netlink.RouteAdd(&netlink.Route{
+		LinkIndex: iface.Attrs().Index,
+		Scope:     netlink.SCOPE_UNIVERSE,
+		Dst:       ipn.Network().ToIPNet(),
+	})
+	if err != nil && err != syscall.EEXIST {
+		return fmt.Errorf("Failed to add route (%v -> %v): %v", ipn.Network().String(), ifname, err)
```
MohdAhmad: Change to `failed` from `Failed`

backend/udp/network.go
```diff
@@ -0,0 +1,204 @@
...
+
+			setRoute(n.ctl, evt.Lease.Subnet, evt.Lease.Attrs.PublicIP, n.port)
+
+		case subnet.EventRemoved:
+			log.Info("Subnet removed: ", evt.Lease.Subnet)
+
+			removeRoute(n.ctl, evt.Lease.Subnet)
+
+		default:
+			log.Error("Internal error: unknown event type: ", int(evt.Type))
```
MohdAhmad: Change to `internal` from `Internal`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
 		if iaddr = net.ParseIP(ifname); iaddr != nil {
 			iface, err = ip.GetInterfaceByIP(iaddr)
 			if err != nil {
-				return nil, nil, fmt.Errorf("Error looking up interface %s: %s", ifname, err)
+				return nil, fmt.Errorf("Error looking up interface %s: %s", ifname, err)
```
MohdAhmad: Change to `error` from `Error`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
...
 			}
 		} else {
 			iface, err = net.InterfaceByName(ifname)
 			if err != nil {
-				return nil, nil, fmt.Errorf("Error looking up interface %s: %s", ifname, err)
+				return nil, fmt.Errorf("Error looking up interface %s: %s", ifname, err)
```
MohdAhmad: Change to `error` from `Error`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
...
 			}
 		}
 	} else {
 		log.Info("Determining IP address of default interface")
 		if iface, err = ip.GetDefaultGatewayIface(); err != nil {
-			return nil, nil, fmt.Errorf("Failed to get default interface: %s", err)
+			return nil, fmt.Errorf("Failed to get default interface: %s", err)
```
MohdAhmad: Change to `failed` from `Failed`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
...
 		}
 	}
 
 	if iaddr == nil {
 		iaddr, err = ip.GetIfaceIP4Addr(iface)
 		if err != nil {
-			return nil, nil, fmt.Errorf("Failed to find IPv4 address for interface %s", iface.Name)
+			return nil, fmt.Errorf("Failed to find IPv4 address for interface %s", iface.Name)
```
MohdAhmad: Change to `failed` from `Failed`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
...
+		}
+	}
+
+	if iface.MTU == 0 {
+		return nil, fmt.Errorf("Failed to determine MTU for %s interface", iaddr)
```
MohdAhmad: Change to `failed` from `Failed`

network/manager.go
```diff
@@ -152,32 +136,56 @@ func lookupExtIface(ifname string) (*net.Interface, net.IP, error) {
...
+	}
+
+	var eaddr net.IP
+
+	if len(opts.publicIP) > 0 {
+		eaddr = net.ParseIP(opts.publicIP)
+		if eaddr == nil {
+			return nil, fmt.Errorf("Invalid public IP address", opts.publicIP)
```
MohdAhmad: Change to `invalid` from `Invalid`

network/network.go
```diff
@@ -25,108 +25,114 @@ import (
...
+		subnet.LeaseRenewer(n.ctx, n.sm, n.Name, n.bn.Lease())
 		wg.Done()
 	}()
 
-	<-n.ctx.Done()
-	n.be.UnregisterNetwork(n.ctx, n.Name)
+	defer func() {
+		if n.ipMasq {
+			if err := teardownIPMasq(n.Config.Network); err != nil {
+				log.Errorf("Failed to tear down IP Masquerade for network %v: %v", n.Name, err)
```
MohdAhmad: Change to `failed` from `Failed`

backend/hostgw/network.go
eyakubovich: Why? These are log statements, not error objects. Our convention is to have log statements start with a capital letter.

backend/udp/network.go
eyakubovich: This is struct embedding. Seems appropriate in this case so that we just pick up `backend.SimpleNetwork` fields.

backend/hostgw/network.go
MohdAhmad: Ah right. I mixed that up.

MohdAhmad:
LGTM

Issue #337, opened 2015-10-06
Flannel eats all memory and cpu.


ssergiienko:
flannel-0.5.1

On Master:

```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                                                  
13842 etcd      20   0   42232  24792   2916 S  93.4  0.9 607:33.77 etcd                                                                                                                     
   41 root      20   0       0      0      0 S   4.3  0.0  18:35.42 rcu_sched                                                                                                                
  418 root      20   0   51208  10708  10544 S   3.7  0.4  33:58.41 systemd-journal                                                                                                          
14116 root      20   0 2428468 1.992g   2096 S   3.3 73.4  51:59.73 flanneld 
```

```
[root@static ~]# systemctl status flanneld -l
flanneld.service - Flanneld overlay address etcd agent
   Loaded: loaded (/etc/systemd/system/flanneld.service; enabled)
   Active: active (running) since Fri 2015-09-25 13:30:58 EEST; 1 day 2h ago
 Main PID: 14116 (flanneld)
   CGroup: /system.slice/flanneld.service
           └─14116 /usr/bin/flanneld -etcd-endpoints=http://127.0.0.1:4001 -etcd-prefix=/some/network/ --iface=enp0s5

933703 14116 hostgw.go:136] Subnet added: 10.254.53.0/24 via IP.IP.IP.IP
933827 14116 hostgw.go:152] Error adding route to 10.254.53.0/24 via IP.IP.IP.IP: file exists
26 16:03:32 DEBUG: rawWatch /some/network/subnets []
26 16:03:32 DEBUG: get /some/network/subnets [http://127.0.0.1:4001]
26 16:03:32 DEBUG: Connecting to etcd: attempt 1 for keys/some/network/subnets?recursive=true&wait=true&waitIndex=654925
26 16:03:32 DEBUG: send.request.to http://127.0.0.1:4001/v2/keys/some/network/subnets?recursive=true&wait=true&waitIndex=654925 | method GET
26 16:03:32 DEBUG: recv.response.from
26 16:03:32 DEBUG: recv.response.from
26 16:03:32 DEBUG: recv.success
26 16:03:32 DEBUG: recv.success
26 16:03:32 DEBUG: rawWatch /some/network/subnets []
26 16:03:32 DEBUG: get /some/network/subnets [http://127.0.0.1:4001]
[...]

PR #339, opened 2015-10-07
remote: close response body during watch()

jonboulle:
From eyeballing #337 this might be the culprit.
/cc @eyakubovich @ssergiienko

### DIFF
--- a/remote/client.go
+++ b/remote/client.go
@@ -183,6 +183,7 @@ func (m *RemoteManager) watch(ctx context.Context, url string, cursor interface{
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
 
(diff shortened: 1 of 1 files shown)


eyakubovich:
Regardless, It's a good bug to catch.

eyakubovich:
LGTM

ravilr:
@eyakubovich is there a release coming soon that this is going to be part of? Thanks.

eyakubovich:
@jonboulle I was hoping that it'll go into the next release (0.6) but it looks like that one is still a bit out and, more importantly, might not be too stable due to lots of turmoil in the codebase lately.  Therefore, I will do 0.5.4 this week with cherry-picked commits from the master.

PR #344, opened 2015-10-14
Typed registry

eyakubovich:
Another refactor. This makes the registry interface more etcd agnostic. Not 100% though.


dcbw:
Overall looks like a great cleanup; I'm on PTO until early next week but I'll test it out when I get back.

eyakubovich:
This is ready for review now.

subnet/mock_registry.go
```diff
@@ -241,103 +252,91 @@ func configKeyToNetworkKey(configKey string) string {
...
-		return nil, fmt.Errorf("Network %s not found", network)
+		return fmt.Errorf("Network %s not found", network)
 	}
+	delete(msr.networks, network)
 
 	msr.index += 1
 
-	n.ModifiedIndex = msr.index
-
 	delete(msr.networks, network)
```
MohdAhmad: `delete` already called on line 327.

subnet/subnet.go
```diff
@@ -32,8 +32,10 @@ type LeaseAttrs struct {
 
 type Lease struct {
 	Subnet     ip.IP4Net
-	Attrs      *LeaseAttrs
+	Attrs      LeaseAttrs
 	Expiration time.Time
+
+	asof uint64
```
MohdAhmad: I do not follow how `asof` is being used. I initially thought it was passed to `updateSubnet` and then used in `etcd.SetOptions` to confirm the previous modified index of the key. But all calls to `updateSubnet` in `local_manager` pass `asof=0`.

subnet/local_manager.go
```diff
@@ -0,0 +1,329 @@
...
+
+	if len(bag) == 0 {
+		return ip.IP4Net{}, errors.New("out of subnets")
+	} else {
+		i := randInt(0, len(bag))
+		return ip.IP4Net{IP: bag[i], PrefixLen: config.SubnetLen}, nil
+	}
+}
+
+func nodeToLease(node *etcd.Node) (*Lease, error) {
```
MohdAhmad: I can't see where `nodeToLease` gets used.

subnet/local_manager.go
```diff
@@ -0,0 +1,329 @@
...
+			Events: []Event{evt},
+			Cursor: watchCursor{index},
+		}, nil
+
+	case err == ErrTryAgain:
+		nextIndex = index
+		goto DoWatch
+
+	case isIndexTooSmall(err):
+		log.Warning("Watch of subnet leases failed because etcd index outside history window")
```
dcbw: "Watch of networks failed..."

subnet/mock_registry.go
```diff
@@ -68,117 +73,121 @@ func (msr *MockSubnetRegistry) setConfig(network, config string) error {
...
-	if msr.ttl > 0 {
-		ttl = msr.ttl
+	// check for existing
+	for _, sub := range n.subnets {
+		if sub.Subnet.Equal(sn) {
+			return time.Time{}, etcd.Error{
+				Code:  etcd.ErrorCodeNodeExist,
+				Index: msr.index,
+			}
+		}
```
dcbw: There's at least 4 places here with the "for sub in n.subnets; if Equal(sn)" thing happens; maybe just have a getter that does the search and returns a *Lease?  Then:

if sub := findSubnet(n, sn); sub != nil {
    return time.Time{}, ...
}

subnet/local_manager.go
```diff
@@ -0,0 +1,329 @@
...
+
+func NewLocalManager(config *EtcdConfig) (Manager, error) {
+	r, err := newEtcdSubnetRegistry(config)
+	if err != nil {
+		return nil, err
+	}
+	return newLocalManager(r, config.Prefix), nil
+}
+
+func newLocalManager(r Registry, prefix string) Manager {
```
dcbw: 'prefix' doesn't seem to be used anywhere

subnet/mock_registry.go
```diff
@@ -17,50 +17,55 @@ package subnet
...
+}
 
-	node := &etcd.Node{Key: normalizeNetKey(network), Value: config, ModifiedIndex: 0, Nodes: make([]*etcd.Node, 0, 20)}
-	for _, n := range initialSubnets {
-		if n.ModifiedIndex > index {
-			index = n.ModifiedIndex
-		}
-		node.Nodes = append(node.Nodes, n)
-	}
+const networkKeyPrefix = "/coreos.com/network"
```
dcbw: Don't need this anymore since we're not emulating etcd.  It's only used by normalizeNetKey() at the bottom of the file, which can now be deleted since it's not called anywhere either.

subnet/mock_registry.go
```diff
@@ -241,103 +252,91 @@ func configKeyToNetworkKey(configKey string) string {
...
 	}
 
-	return &etcd.Response{
-		Node:  n,
-		Index: msr.index,
-	}, nil
+	return nil
 }
 
 func normalizeNetKey(key string) string {
```
dcbw: This function is now unused.

subnet/subnet.go
eyakubovich: You're right. I started with a branch for doing reservations and then broke it apart into what is now 4 separate PRs (this is 2nd). The `asof` is only needed in the last one, the one doing reservations. I guess I could yank it here and add it into the last one where it's needed but it's not doing harm so can we let it hang out here for just a bit?

subnet/local_manager.go
eyakubovich: Same story as with `asof`. It's used in the next PR but I'll move it there.

subnet/mock_registry.go
eyakubovich: Very good point.

subnet/subnet.go
MohdAhmad: Yes, that sounds good.

eyakubovich:
@dcbw I think I fixed everything you pointed out. PTAL

pkg/vtime/time.go
```diff
@@ -0,0 +1,38 @@
+package vtime
```
jonboulle: can you use jonboulle/clockwork instead? that's what we use in other CoreOS projects.

eyakubovich: Nice! It has `After` which was on my todo list to implement.

subnet/mock_registry.go
```diff
@@ -68,117 +70,118 @@ func (msr *MockSubnetRegistry) setConfig(network, config string) error {
...
-				Action: "delete",
-				Node:   sub,
-			}
-
-			return &etcd.Response{
-				Node:  sub,
-				Index: msr.index,
-			}, nil
+	for i, sub := range n.subnets {
+		if sub.Subnet.Equal(sn) {
```
dcbw: Could use findSubnet() here.

dcbw:
Other than the last findSubnet() thing, LGTM

subnet/mock_registry.go
eyakubovich: You're right.

PR #349, opened 2015-10-20
build: use `git describe` output in version

jonboulle:
This is much more helpful for debugging when flanneld is built from
master.

### DIFF
--- a/dist/bump-release.sh
+++ b/dist/bump-release.sh
@@ -13,8 +13,8 @@ VERSIONTAG="v${VERSION}"
 TAGBR="v${VERSION}-tag"
 
 replace_version() {
-	sed -i -e "s/const Version.*/const Version = \"$1\"/" ../version.go
-	git commit -m "version: bump to v$1" ../version.go
+	sed -i -e "s/const Version.*/const Version = \"$1\"/" version/version.go
+	git commit -m "version: bump to v$1" version/version.go
 }
 
 # make sure we're up to date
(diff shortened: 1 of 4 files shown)


eyakubovich:
Great idea but can you also fix up https://github.com/coreos/flannel/blob/master/dist/bump-release.sh to take this into account.

jonboulle:
@eyakubovich tweaked, thoughts?

dist/bump-release.sh
```diff
@@ -13,8 +13,8 @@ VERSIONTAG="v${VERSION}"
 TAGBR="v${VERSION}-tag"
 
 replace_version() {
```
eyakubovich: I don't think we need this function at all now. There's no point in creating these bump commits any more. We can just tag the release and that's it.

version/version.go
```diff
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package version
 
-const Version = "0.5.3+git"
+var Version = "0.5.3+was-not-built-properly"
```
eyakubovich: So this can go back to `0.0.0+was-not-built-properly`

jonboulle: The motivation for this was to make the go get case a bit cleaner.

eyakubovich: Yeah but then we still need the bump revision dance. I like that `git describe` magic as it cleans it up.

jonboulle: Kind of unpacking a can of worms I didn't want to, but thoughts on etcd's
approach?
```
etcd Version: 2.2.1+git
Git SHA: Not provided (use ./build instead of go build)
Go Version: go1.5.1
Go OS/Arch: linux/amd64
```

On Wed, Oct 21, 2015 at 6:42 PM, Eugene Yakubovich <notifications@github.com

eyakubovich: Fancy. I think this should be standardized across the company. I don't care what really gets included apart from the version but I do now like the idea of not using `sed` and generating those bump commits.

eyakubovich:
Closing in favor of #354

PR #354, opened 2015-10-30
build: use `git describe` output in version

eyakubovich:


dist/bump-release.sh
```diff
@@ -24,8 +24,8 @@ git pull --ff-only ${ORIGIN} master
 replace_version ${VERSION}
 git tag -a -m "${VERSIONTAG}" "${VERSIONTAG}"
 
-# bump ver to +git and push to origin
-replace_version "${VERSION}+git"
+# bump ver to placeholder and push to origin
+replace_version "${VERSION}+was-not-built-properly"
```
jonboulle: I think you intended to revert this bit right..?

eyakubovich: It's fine. I wanted to get rid of the bumping logic until https://github.com/coreos/docs/issues/637#issuecomment-152661595

jonboulle: er, right, but we want it to be +git don't we? or did you disagree with that premise of having it as descrptive as possible?

eyakubovich: I actually don't care that much. I want people to build with the `build` script so this will inform that to do so and thus I like this +was-not-built-properly. But if you have strong feelings for +git, I'll be happy to revert.

eyakubovich:
@jonboulle is this version cleaner?

dist/bump-release.sh
jonboulle: Please revert this for now (at the very least for consistency with version/version.go), and we can revisit once we figure out the more general approach.

jonboulle:
LGTM after last comment

PR #355, opened 2015-11-02
Actually track backends in the active map

dcbw:


dcbw:
Missed this before because I was only testing with one network.

dcbw:
If we want testcase coverage for the localmanager subnet watch fix, we're probably going to have to bring back the mock etcd in some form.

eyakubovich:
LGTM

Issue #280, opened 2015-08-11
Reservation Mode


bcwaldon:
I'd like to be able to run flannel in such a mode where each node blocks until a subnet has been allocated to it via the API. This means that the node is no longer responsible for allocating and maintaining a lease itself.

I have two functional requirements for this:
1. flannel nodes running in reservation mode block until a reservation is available - I don't want a race between the reservation being available in etcd and the node starting up
2. the flannel API must provide an endpoint that allows me to create and destroy reservations

I'm not sure what the behavior should be when a reservation is deleted

bprashanth:
I solved a similar problem on Kubernetes with a proxy flannel server https://github.com/kubernetes/kubernetes/pull/13877. The flannel daemons block till the master assigns the node a subnet, @eyakubovich let me know if you're interested on collaborating to a similar end goal. With my patch, the flannel daemon can only read subnets, not create or destroy them.

eyakubovich:
@bprashanth I actually have the PR for this ticket mostly done. It's being held up by a few things. First, the 2 refactoring PRs (#338 and #344), and then a patch to add lease revocations. The reservation in etcd is denoted by having the subnet node have a TTL=0 (no expiration). There's also a remote API calls for `AddReservation` and `RemoveReservation`. This does not make etcd read-only for flannel as it still has to set things like BackendData but with "flannel server" it is possible to gate the flannel daemons from the having general etcd access. Is that sufficient for your needs?

bprashanth:
sounds like the proxy flannel server will still have to manage any persistence to etcd and redirect to apiserver. That's ok for node specific metadata (eg VTEP), not ok if it's going to delete a subnet from the global pool. IMO global resources should fall under the purview of the cluster manager, so as long as the daemon doesn't go mad if the flannel server no-ops a Delete/Add we should be ok. 

I actually don't need reservation mode currently because the daemon just hangs on /subnet/leases (whichever one of those endpoints it hits first) till the master tells it what subnet to use.

eyakubovich:
There is no authorization per API call so a node could end up deleting a subnet. The server does need to have methods like "RevokeLease" or "SetNetworkConfig" but the question is who should be allowed to call that. I think we could add access control in the future to solve this problem.

andrewkrug:
How do you use this feature?   It looks like it was merged without updating any docs.

PR #356, opened 2015-11-02
Ability to revoke lease

eyakubovich:
This adds an ability to revoke a lease by deleting it
from etcd. A corresponding API for server case is also added.
Once the lease is revoked, flannel will just acquire a new
one. This has limited use today but is part of #280 to
add reservations. With reservations, there will be a mode
to not acquire a lease if a reservation was not found.


eyakubovich:
@MohdAhmad @jonboulle PTAL

subnet/registry.go
```diff
@@ -209,6 +211,21 @@ func (esr *etcdSubnetRegistry) watchSubnets(ctx context.Context, network string,
 	return evt, e.Node.ModifiedIndex, err
 }
 
+func (esr *etcdSubnetRegistry) watchSubnet(ctx context.Context, network string, since uint64, sn ip.IP4Net) (Event, uint64, error) {
+	key := path.Join(esr.etcdCfg.Prefix, "subnets", network, sn.StringSep(".", "-"))
```
dcbw: Need to swap 'network' and 'subnets' here.

subnet/registry.go
```diff
@@ -127,27 +127,29 @@ func (esr *etcdSubnetRegistry) getSubnets(ctx context.Context, network string) (
...
 		}
+
+		leases = append(leases, *l)
 	}
 
 	return leases, resp.Index, nil
 }
 
+func (esr *etcdSubnetRegistry) getSubnet(ctx context.Context, network string, sn ip.IP4Net) (*Lease, uint64, error) {
+	key := path.Join(esr.etcdCfg.Prefix, network, "subnets", sn.StringSep(".", "-"))
```
dcbw: Maybe use "MakeSubnetKey(sn)" here (and everywhere else) instead of StringSep?

dcbw:
Overall LGTM

PR #358, opened 2015-11-05
Periodically retry getting initial networks

dcbw:
In a cluster the master flannel or etcd might not be up yet when
flanneld starts, so periodically retry getting initial networks
instead of failing hard.

### DIFF
--- a/network/manager.go
+++ b/network/manager.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/coreos/go-systemd/daemon"
 	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
@@ -106,22 +107,6 @@ func NewNetworkManager(ctx context.Context, sm subnet.Manager) (*Manager, error)
 		if name != "" {
 			manager.allowedNetworks[name] = true
 		}
-	}
-
-	if manager.isMultiNetwork() {
-		// Get list of existing networks
-		result, err := manager.sm.WatchNetworks(ctx, nil)
-		if err != nil {
-			return nil, err
-		}
-
-		for _, n := range result.Snapshot {
-			if manager.isNetAllowed(n) {
-				manager.networks[n] = NewNetwork(ctx, sm, bm, n, manager.ipMasq)
-			}
-		}
-	} else {
-		manager.networks[""] = NewNetwork(ctx, sm, bm, "", manager.ipMasq)
 	}
 
 	return manager, nil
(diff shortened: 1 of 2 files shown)


network/manager.go
```diff
@@ -324,9 +309,54 @@ func (m *Manager) watchNetworks() {
...
+		}
+
+		// Otherwise retry in a few seconds
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(time.Second):
+		}
+
+		if i%5 == 0 {
```
jonboulle: why are you only logging this every 5?

network/manager.go
```diff
@@ -324,9 +309,54 @@ func (m *Manager) watchNetworks() {
...
+			log.V(2).Info("Retrying initial network retrieval...")
```
jonboulle: IMHO it'd be better to log the error when this happens rather than just that we're retrying; I think the signal-to-noise trade-off is acceptable

network/manager.go
```diff
@@ -324,9 +309,54 @@ func (m *Manager) watchNetworks() {
...
+			return false
+		}
+
+		for _, n := range result.Snapshot {
+			if m.isNetAllowed(n) {
+				m.networks[n] = NewNetwork(ctx, m.sm, m.bm, n, m.ipMasq)
+			}
+		}
+	} else {
+		if _, err := m.sm.GetNetworkConfig(ctx, ""); err != nil {
```
jonboulle: hrm, why wasn't this necessary previously?

eyakubovich:
I'm confused why it's failing hard. There's a big for loop inside `Network.Run`: https://github.com/coreos/flannel/blob/master/network/network.go#L98

network/manager.go
dcbw: To reduce logspam, otherwise it would get printed every second.  I'd imagine this would often be a short-term log message though, until the etcd master or flannel master gets started up.

network/manager.go
dcbw: Ok.

dcbw:
> I'm confused why it's failing hard. There's a big for loop inside Network.Run: https://github.com/coreos/flannel/blob/master/network/network.go#L98

It was failing long before the Network.Run(), because with multinetwork mode, to even get a Network object, you have to ask etcd/flannel-remote for the list of networks.

network/manager.go
dcbw: I added the GetNetworkConfig() for consistency with the multinetwork case.  Previously, this would have simply passed because NewNetwork() doesn't require any kind of etcd/server connectivity.  So you would then get to Network.Run() and hit that retry loop, I suppose.  This way, both codepaths ask for the etcd/server data at the same time.

I'm happy to remove this new bit if you like, and leave it up to Network.Run()?  Just thought some consistency of behavior might be useful.

eyakubovich:
>It was failing long before the Network.Run(), because with multinetwork mode, to even get a Network object, you have to ask etcd/flannel-remote for the list of networks.

Makes sense though the retry loop is best placed around the operation of get the list of networks (https://github.com/dcbw/flannel/blob/6b38ffa5f879e3711f0b129f5edc9c198acc11b2/network/manager.go#L315) for more clarity.

dcbw:
@eyakubovich @jonboulle PTAL

network/manager.go
```diff
@@ -324,9 +309,33 @@ func (m *Manager) watchNetworks() {
...
 }
 
-func (m *Manager) Run() {
+func (m *Manager) Run(ctx context.Context) {
 	wg := sync.WaitGroup{}
 
+	if m.isMultiNetwork() {
+		for {
+			// Try adding initial networks
+			if result, err := m.sm.WatchNetworks(ctx, nil); err == nil {
```
eyakubovich: Looks much better. But like Jon said, it's better to log the error to help the user diagnose the problem (even though in the case of etcd not being up, the error is confusing). Logging that a retry is coming is less useful.

dcbw:
@eyakubovich does that look like what you meant now?

eyakubovich:
LGTM

dcbw:
Thanks!

PR #362, opened 2015-11-09
Allow network leases to be nil and networks to contribute to subnet file

dcbw:
For the OpenShift OVS stuff we have an IPAM model where each node grabs a subnet lease, rather than each tenant network having its own subnet.  Mutli-tenancy is enforced with the OpenFlow rules.

The approach I took for the plugin is one well-known "cluster network" which every node running flannel grabs a lease in.  Tenant networks never have leases (since pod IPs are allocated out of the common node subnet), but they do have other data that we'd like to write to subnet files (like the VNID, or Virtual Network ID).  See https://github.com/dcbw/flannel/tree/ovs for details.

### DIFF
--- a/backend/hostgw/network.go
+++ b/backend/hostgw/network.go
@@ -45,6 +45,10 @@ func (n *network) MTU() int {
 	return n.extIface.Iface.MTU
 }
 
+func (n *network) SubnetFileVars() map[string]string {
+	return nil
+}
+
 func (n *network) Run(ctx context.Context) {
 	wg := sync.WaitGroup{}
 
--- a/network/network.go
+++ b/network/network.go
@@ -130,14 +130,6 @@ func (n *Network) runOnce(extIface *backend.ExternalInterface, inited func(bn ba
 		wg.Done()
 	}()
 
-	evts := make(chan subnet.Event)
-
-	wg.Add(1)
-	go func() {
-		subnet.WatchLease(ctx, n.sm, n.Name, n.bn.Lease().Subnet, evts)
-		wg.Done()
-	}()
-
 	defer func() {
 		if n.ipMasq {
 			if err := teardownIPMasq(n.Config.Network); err != nil {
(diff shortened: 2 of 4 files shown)


backend/hostgw/network.go
```diff
@@ -45,6 +45,10 @@ func (n *network) MTU() int {
 	return n.extIface.Iface.MTU
 }
 
+func (n *network) SubnetFileVars() string {
```
eyakubovich: It's cleaner to return `map[string]string` and render it to text in the function that writes out the file.

dcbw:
@eyakubovich Right you are, that's much better. Updated, PTAL

network/network.go
```diff
@@ -148,33 +140,46 @@ func (n *Network) runOnce(extIface *backend.ExternalInterface, inited func(bn ba
...
-
-			log.Info("Lease renewed, new expiration: ", n.bn.Lease().Expiration)
-			dur = n.bn.Lease().Expiration.Sub(time.Now()) - renewMargin
-
-		case e := <-evts:
-			if e.Type == subnet.EventRemoved {
-				log.Warning("Lease has been revoked")
-				interruptFunc()
-				return errInterrupted
+	if n.bn.Lease() != nil {
```
eyakubovich: This one is hard to swallow as it makes it unclear what a network does if not for getting itself a lease for this node. I looked at your ovs branch and I see how you're using it but I feel uneasy merging this change. In essence the only thing that the network does in ovs branch is read in the config with VNI from etcd and output that VNI to the subnet.env file. I feel that if you'd like to use one big address space for all the networks, it's easier to have flannel just set up one network and do flow management (with VNI tagging) completely out of band from flannel. You can still store tenant->VNI mapping somewhere in etcd.

dcbw:
Fair enough, I'll take a bit of time to think about how to rework. Thanks for the review.

PR #369, opened 2015-11-14
Add reservations to admin control subnet allocs

### LINKED ISSUE #280: Reservation Mode

eyakubovich:
This adds ability to add and remove reservations
for subnet allocations. A subnet is reserved with
the host IP (PublicIP). flannel will then use this
subnet for a host requesting a subnet.

Reservations are denoted in etcd as subnets with no
expiration (TTL is zero). When a reserved subnet is
allocated, it's TTL continues to be not set. If a
reservation is removed, the TTL is set to the usual
24 hours.

Also adds corresponding client/server APIs for
reservation management.

Fixes #280

### DIFF
--- a/remote/server.go
+++ b/remote/server.go
@@ -46,8 +46,6 @@ func jsonResponse(w http.ResponseWriter, code int, v interface{}) {
 
 // GET /{network}/config
 func handleGetNetworkConfig(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -65,8 +63,6 @@ func handleGetNetworkConfig(ctx context.Context, sm subnet.Manager, w http.Respo
 
 // POST /{network}/leases
 func handleAcquireLease(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -91,8 +87,6 @@ func handleAcquireLease(ctx context.Context, sm subnet.Manager, w http.ResponseW
 
 // PUT /{network}/{lease.network}
 func handleRenewLease(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -115,8 +109,6 @@ func handleRenewLease(ctx context.Context, sm subnet.Manager, w http.ResponseWri
 }
 
 func handleRevokeLease(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -134,8 +126,6 @@ func handleRevokeLease(ctx context.Context, sm subnet.Manager, w http.ResponseWr
 		fmt.Fprint(w, err)
 		return
 	}
-
-	w.WriteHeader(http.StatusOK)
 }
 
 func getCursor(u *url.URL) interface{} {
@@ -148,8 +138,6 @@ func getCursor(u *url.URL) interface{} {
 
 // GET /{network}/leases/subnet?next=cursor
 func handleWatchLease(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -186,8 +174,6 @@ func handleWatchLease(ctx context.Context, sm subnet.Manager, w http.ResponseWri
 
 // GET /{network}/leases?next=cursor
 func handleWatchLeases(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	network := mux.Vars(r)["network"]
 	if network == "_" {
 		network = ""
@@ -218,8 +204,6 @@ func handleWatchLeases(ctx context.Context, sm subnet.Manager, w http.ResponseWr
 // GET /?next=cursor watches
 // GET / retrieves all networks
 func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-
 	cursor := getCursor(r.URL)
 	wr, err := sm.WatchNetworks(ctx, cursor)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


remote/server.go
```diff
@@ -241,6 +241,73 @@ func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWrite
 	jsonResponse(w, http.StatusOK, wr)
 }
 
+// POST /{network}/reservations
+func handleAddReservation(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
```
xiang90: The handler does not need to close the request body. The http server will do that for you.

Refer: https://golang.org/pkg/net/http/#Request

remote/server.go
```diff
@@ -241,6 +241,73 @@ func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWrite
...
+
+	network := mux.Vars(r)["network"]
+	if network == "_" {
+		network = ""
+	}
+
+	rsv := &subnet.Reservation{}
+	if err := json.NewDecoder(r.Body).Decode(rsv); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprint(w, "JSON decoding error: ", err)
```
xiang90: The error string does not need to be json format?

remote/server.go
```diff
@@ -241,6 +241,73 @@ func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWrite
...
+		return
+	}
+
+	if err := sm.AddReservation(ctx, network, rsv); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, fmt.Errorf("internal error: %v", err))
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
xiang90: You do not have to write http.StatusOK. But it looks OK.

remote/server.go
```diff
@@ -241,6 +241,73 @@ func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWrite
...
+
+	network := mux.Vars(r)["network"]
+	if network == "_" {
+		network = ""
+	}
+
+	leases, err := sm.ListReservations(ctx, network)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, err)
```
xiang90: we should return right after 305, right?

remote/server.go
```diff
@@ -241,6 +241,73 @@ func handleNetworks(ctx context.Context, sm subnet.Manager, w http.ResponseWrite
...
+	if err := sm.RemoveReservation(ctx, network, *sn); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func handleListReservations(ctx context.Context, sm subnet.Manager, w http.ResponseWriter, r *http.Request) {
```
xiang90: also add a comment for this handler like the above for consistency?

subnet/local_manager.go
```diff
@@ -344,3 +364,126 @@ func isSubnetConfigCompat(config *Config, sn ip.IP4Net) bool {
...
+	}
+
+	// This subnet or its reservation already exists.
+	// Get what's there and
+	// - if PublicIP matches, remove the TTL make it a reservation
+	// - otherwise, error out
+	sub, asof, err := m.registry.getSubnet(ctx, network, r.Subnet)
+	switch {
+	case err == nil:
+	case isErrEtcdNodeExist(err):
```
xiang90: Should this be isErrNodeNotExist instead of Exist?

subnet/subnet.go
```diff
@@ -26,6 +26,11 @@ import (
 	"github.com/coreos/flannel/pkg/ip"
 )
 
+var (
+	ErrLeaseTaken  = errors.New("lease already taken")
```
xiang90: The public error usually have a prefix. 

```go
ErrLeaseTake = errors.New("subnet: lease already take")
```

subnet/local_manager.go
```diff
@@ -344,3 +364,126 @@ func isSubnetConfigCompat(config *Config, sn ip.IP4Net) bool {
...
+			continue
+		default:
+			return err
+		}
+	}
+
+	return ErrNoMoreTries
+}
+
+func (m *LocalManager) tryRemoveReservation(ctx context.Context, network string, subnet ip.IP4Net) error {
```
xiang90: I feel this function worth a comment. RemoveReservation remove the subnet by setting TTL back to subnetTTL (24hours)

subnet/subnet.go
```diff
@@ -129,4 +139,8 @@ type Manager interface {
 	WatchLease(ctx context.Context, network string, sn ip.IP4Net, cursor interface{}) (LeaseWatchResult, error)
 	WatchLeases(ctx context.Context, network string, cursor interface{}) (LeaseWatchResult, error)
 	WatchNetworks(ctx context.Context, cursor interface{}) (NetworkWatchResult, error)
+
+	AddReservation(ctx context.Context, network string, r *Reservation) error
```
xiang90: I would like to have comments for all public interface. This might be a good start.

subnet/subnet_test.go
```diff
@@ -41,7 +41,7 @@ func newDummyRegistry() *MockSubnetRegistry {
 		{ip.IP4Net{ip.MustParseIP4("10.3.5.0"), 24}, attrs, exp, 13},
 	}
 
-	config := `{ "Network": "10.3.0.0/16", "SubnetMin": "10.3.1.0", "SubnetMax": "10.3.5.0" }`
+	config := `{ "Network": "10.3.0.0/16", "SubnetMin": "10.3.1.0", "SubnetMax": "10.3.25.0" }`
```
xiang90: I failed to see why do we need this change in this commit.

subnet/subnet_test.go
```diff
@@ -402,3 +403,129 @@ func TestWatchNetworkRemoved(t *testing.T) {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	r := Reservation{
+		Subnet:   newIP4Net("10.4.3.0", 24),
+		PublicIP: ip.MustParseIP4("52.195.12.13"),
+	}
+	if err := sm.AddReservation(ctx, "_", &r); err == nil {
+		t.Fatalf("added reservation outside of configured network")
```
xiang90: I feel this fatal msg is not clear enough. How about `unexpected adding reservation outside of configured network`

subnet/subnet_test.go
```diff
@@ -402,3 +403,129 @@ func TestWatchNetworkRemoved(t *testing.T) {
...
+		t.Fatalf("failed to acquire subnet: %v", err)
+	}
+	if !l.Subnet.Equal(r.Subnet) {
+		t.Fatalf("acquired subnet is not the reserved one: expected %v, got %v", r.Subnet, l.Subnet)
+	}
+	if !l.Expiration.IsZero() {
+		t.Fatalf("acquired lease (prev reserved) has expiration set")
+	}
+
+	// Add the same reservation -- should succeed
```
xiang90: nit: I would suggest to move the last two subtests to line 426. So that we have reservation related tests all together. Then we try to acquireLease on a reservation.

subnet/subnet_test.go
```diff
@@ -402,3 +403,129 @@ func TestWatchNetworkRemoved(t *testing.T) {
...
+		t.Fatalf("failed to remove reservation: %v", err)
+	}
+
+	// The node should not have a TTL
+	sub, _, err := msr.getSubnet(ctx, "_", r.Subnet)
+	if err != nil {
+		t.Fatalf("getSubnet failed: %v", err)
+	}
+
+	if sub.Expiration.IsZero() {
```
xiang90: shouldnt we get a 24 hour ttl?

xiang90:
Feel good about the overall approach.

remote/server.go
eyakubovich: Nice, never knew that.

subnet/local_manager.go
eyakubovich: Great catch. Obviously I'm missing a unit test for this.

subnet/subnet_test.go
eyakubovich: Yes but the real invariant is that removed reservation has non-zero expiration. 24 hr lease is an implementation detail as far as reservation logic is concerned.

subnet/subnet_test.go
eyakubovich: It used to check that AcquireLease worked by making sure it go `10.3.3.0/24` subnet because it configured a very narrow range and there was only one valid subnet left to allocate. This doesn't work with reservations b/c it may allocate out of SubnetMin/SubnetMax range if there's a reservation there. I now check via `inAllocatableRange` and so gave "room to breathe" -- this way it can do more than one allocation, if needed.

remote/server.go
eyakubovich: I think that's a good point. I'll file an issue to evaluate changing a API on how the errors are reported.

subnet/subnet_test.go
xiang90: Hmm... OK. Can you explain why it is 24hours not 1hour if this is just an implementation detail?

subnet/subnet_test.go
```diff
@@ -402,3 +403,129 @@ func TestWatchNetworkRemoved(t *testing.T) {
...
+	}
+	if err := sm.AddReservation(ctx, "_", &r); err != nil {
+		t.Fatalf("failed to add reservation: %v", err)
+	}
+
+	if err := sm.RemoveReservation(ctx, "_", r.Subnet); err != nil {
+		t.Fatalf("failed to remove reservation: %v", err)
+	}
+
+	// The node should not have a TTL
```
xiang90: The node should have a TTL?

subnet/subnet_test.go
xiang90: Got you. Thanks!

subnet/subnet_test.go
xiang90: Also I actually feel the removed one with ttl and the reservation without ttl is a little bit intuitive. But I have no better suggestion.

eyakubovich: It is not intuitive and not even ideal. But we can't just delete it as it might be in use.

subnet/local_manager.go
eyakubovich: Oh yeah, I remember why I don't have a test for it -- it's very difficult to write as I would need to perfectly time a race.

subnet/local_manager.go
eyakubovich: I added a comment on `RemoveReservation`.

xiang90:
Do we have an e2e test or integration test yet? 

The code LGTM.

eyakubovich:
We don't have e2e tests. That's on a todo list and it's non-trivial task as it requires setting up etcd and multiple containers or VMs for flannel daemons.

PR #372, opened 2015-11-17
Revendor netlink library

eyakubovich:
This is primarily to pick up kernel filtering in NeighList function.
The fix was previously vendored off a fork:
https://github.com/coreos/flannel/commit/c5ba18b863788d701d9f8fd9faefa423f1e63a96
but now the fix is upstreamed.


MohdAhmad:
LGTM

xiang90:
LGTM

PR #380, opened 2015-11-30
Add mock etcd and etcd-backed registry testcases

dcbw:
Main motivation here was making sure that the "subnet" and "network" don't get switched around in registry.go any more.  I chose to re-implement a mock etcd since I wasn't sure if flannel testing could necessarily spin up its own etcd on all configurations.

### DIFF
--- a/subnet/mock_etcd.go
+++ b/subnet/mock_etcd.go
@@ -1,0 +1,426 @@
+// Copyright 2015 flannel authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package subnet
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	etcd "github.com/coreos/flannel/Godeps/_workspace/src/github.com/coreos/etcd/client"
+	"github.com/coreos/flannel/Godeps/_workspace/src/golang.org/x/net/context"
+)
+
+const DEFAULT_TTL time.Duration = 8760 * time.Hour // one year
+
+type mockEtcd struct {
+	nodes    map[string]*etcd.Node
+	watchers map[*watcher]*watcher
+	// A given number of past events must be available for watchers, because
+	// flannel always uses a new watcher instead of re-using old ones, and
+	// the new watcher's index may be slightly in the past
+	events []*etcd.Response
+	index  uint64
+}
+
+func newMockEtcd() *mockEtcd {
+	me := &mockEtcd{
+		index:    1000,
+		nodes:    make(map[string]*etcd.Node),
+		watchers: make(map[*watcher]*watcher),
+		events:   make([]*etcd.Response, 0, 50),
+	}
+	me.nodes["/"] = me.newNode("/", "", true)
+
+	return me
+}
+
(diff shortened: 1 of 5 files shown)


subnet/registry_test.go
```diff
@@ -0,0 +1,425 @@
...
+		}
+
+		if numEvents == len(expected) {
+			result <- nil
+			break
+		}
+	}
+}
+
+func TestMockEtcd(t *testing.T) {
```
eyakubovich: nit: would be good to mock etcd testing into a separate file.

eyakubovich:
Thank you for doing the work. Now that registry does more, this was badly needed.

LGTM except for one nit.

jonboulle:
@eyakubovich actually, @xiang90 had a couple of concerns about this
approach - paging him before this lands..

xiang90:
@eyakubovich @dcbw 

Can we simply embed etcd for testing? I do not think we need to mock etcd. 

Another simpler solution I can see is to mock the registry instead of etcd.

subnet/mock_etcd.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+	me.events = append(me.events, resp)
+
+	// and notify watchers
+	for w := range me.watchers {
+		w.notifyEvent(resp)
+	}
+}
+
+// path[len(path)-1] is the node to make; returns the final node created and its creation response
```
dcbw: Also this comment is out-of-date, I'll respin.

eyakubovich:
@xiang90 We actually do have mock registry. The registry used to be a thin wrapper around etcd and was there to basically mock etcd for testing. It has grown to do more stuff so it needs to be tested. Therefore I think MockRegistry actually has little value now and MockEtcd makes sense.

So I do think mock etcd makes sense but the bigger question is whether it belongs in flannel or etcd or someplace else. I know this has come up before, people want to have a drop-in mock etcd.

What do you mean by "embed etcd for testing"?

xiang90:
> What do you mean by "embed etcd for testing"?

You do not have to mock it. You can simply import etcd and run it within your application for testing. 

You can do https://github.com/coreos/etcd/blob/master/integration/cluster_test.go#L93, then you get a running etcd cluster.

eyakubovich:
It is an option, I guess. However we would need to be careful that it uses a temporary (unique) data dir. Also, for testing things like TTLs, it's desirable for it to integrate with https://github.com/jonboulle/clockwork so time can be manually advanced.

dcbw:
I was trying to entirely avoid the need for things like on-disk storage during tests and figuring out what IP address and port to use for etcd itself on test machines.  Hence all the mockage :)  But if we did import etcd we could get rid of the testcase for testing the mock etcd at least.

Either way, just let me know what to do.  Obviously the mock etcd is much less of a real-world test than running etcd itself is, but it has a much smaller bug surface and much more control over how we test things internally and what we can tweak to exercise behavior of flannel itself.

xiang90:
I would still argue we can mock a registry for exercising flannel without the assumption that etcd is the only backend. Mocking etcd is hard since you have to make assumptions about etcd's behavior.

I just do not feel good about introducing etcd logic into flannel codebase.

eyakubovich:
You're not really introducing etcd logic into flannel. You're mocking out an etcd interface. flannel already is coupled to it b/c it has to interface with etcd.

I do feel that having a mock etcd inside github.com/coreos/etcd/client is actually a good idea. It would be very similar to how libkv does it: https://github.com/docker/libkv/tree/master/store/mock

eyakubovich:
My view is that etcd is a database. Mocking out databases is as old as unit testing itself. I have no qualms about merging this. @dcbw if you have a sec to slit out `TestMockEtcd` into a separate file, it would be awesome. But I can merge as is.

dcbw:
@eyakubovich I'll split it out and fix the comment I posted about.

dcbw:
@eyakubovich PTAL

eyakubovich:
LGTM

PR #384, opened 2015-12-03
Add Open vSwitch-based multitenant backend for use with OpenShift / Kubernetes

dcbw:
Reworked for @eyakubovich  comments; the fake networks have been removed and the code simplified.  Without the fake networks the plugin only needs to handle a single network, the cluster network.  I chose to keep that network using a well-known network name for a couple reasons:

1) OpenShift or Kubernetes only allow one container orchestration instance running at a given time on a given node anyway; that means a single 'network' controlled by this backend.  No great reason to allow that name to be configurable.

2) this plugin only configures the OVS bridge and associated interfaces; it does not have anything to do with the actual containers.  The container orchestrator must know a few things in common with this plugin; at least the flannel network name (so that it can write the cluster network and services network to etcd that flannel reads) and the OVS bridge device name (so that it can attach containers to the OVS interface).  Since there can be only one network, and one OVS bridge at this time, they are hardcoded.  (yes, we could make the OVS bridge and associated device names dynamic and write them to the subnet file I guess?  Seems like a lot of work for no reason...)

Also, perhaps we need a better name than 'ovs' since there could be other plugins that use OVS for something else, that aren't this one.  But I'd rather not have 'openshift' since we're hoping to spin this out to pure Kubernetes+flannel.  kube-ovs-multitenant?  Something else?

Thoughts?

### DIFF
--- a/backend/ovs/device.go
+++ b/backend/ovs/device.go
@@ -1,0 +1,661 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ovs
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"net"
+	"os/exec"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	log "github.com/coreos/flannel/Godeps/_workspace/src/github.com/golang/glog"
+	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/vishvananda/netlink"
+	"github.com/coreos/flannel/Godeps/_workspace/src/github.com/vishvananda/netlink/nl"
+
+	"github.com/coreos/flannel/pkg/ip"
+)
+
+type ovsDevice struct {
+	bridgeName      string
+	oflowProto      string
+	opHandler       OpHandler
+	clusterNetwork  ip.IP4Net
+	nodeNetwork     ip.IP4Net
+	servicesNetwork ip.IP4Net
+	gwAddr          net.IP
+}
+
+func newOVSDevice() (*ovsDevice, error) {
+	return newOVSDeviceWithHandler(&liveOpHandler{})
+}
+
+func newOVSDeviceWithHandler(handler OpHandler) (*ovsDevice, error) {
+	bridge := &ovsDevice{
+		bridgeName: "br0",
+		oflowProto: "OpenFlow13",
+		opHandler:  handler,
+	}
+	err := bridge.genericSetup()
+	return bridge, err
+}
+
+//*****************************************************************************
+
+type OpHandler interface {
+	OvsExec(cmd string, args ...string) ([]byte, error)
+	Modprobe(module string) ([]byte, error)
+	LinkAdd(link netlink.Link) error
+	LinkDel(link netlink.Link) error
+	LinkByIndex(index int) (netlink.Link, error)
+	LinkByName(name string) (netlink.Link, error)
+	LinkSetUp(link netlink.Link) error
+	LinkSetMaster(link netlink.Link, master *netlink.Bridge) error
+	AddrAdd(link netlink.Link, addr *netlink.Addr) error
+	AddrList(link netlink.Link, family int) ([]netlink.Addr, error)
+	RouteList(link netlink.Link, family int) ([]netlink.Route, error)
+	RouteAdd(route *netlink.Route) error
+	RouteDel(route *netlink.Route) error
+	RouteDelWithProtocol(route *netlink.Route, protocol uint8) error
+	Sysctl(item, value string) error
+}
+
+type liveOpHandler struct{}
+
+func (h *liveOpHandler) OvsExec(cmd string, args ...string) ([]byte, error) {
+	return exec.Command(cmd, args...).CombinedOutput()
+}
+
+func (h *liveOpHandler) Modprobe(module string) ([]byte, error) {
+	return exec.Command("modprobe", module).CombinedOutput()
+}
+
+func (h *liveOpHandler) LinkAdd(link netlink.Link) error {
+	return netlink.LinkAdd(link)
+}
+
+func (h *liveOpHandler) LinkDel(link netlink.Link) error {
+	return netlink.LinkDel(link)
+}
+
+func (h *liveOpHandler) LinkByName(name string) (netlink.Link, error) {
+	return netlink.LinkByName(name)
+}
+
+func (h *liveOpHandler) LinkByIndex(index int) (netlink.Link, error) {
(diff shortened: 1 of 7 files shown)


backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+	gwAddr          net.IP
+}
+
+func newOVSDevice() (*ovsDevice, error) {
+	return newOVSDeviceWithHandler(&liveOpHandler{})
+}
+
+func newOVSDeviceWithHandler(handler OpHandler) (*ovsDevice, error) {
+	bridge := &ovsDevice{
+		bridgeName: "br0",
```
eyakubovich: If hard coding, please make a constant

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+
+func (h *liveOpHandler) RouteAdd(route *netlink.Route) error {
+	return netlink.RouteAdd(route)
+}
+
+func (h *liveOpHandler) RouteDel(route *netlink.Route) error {
+	return netlink.RouteDel(route)
+}
+
+func (h *liveOpHandler) RouteDelWithProtocol(route *netlink.Route, protocol uint8) error {
```
eyakubovich: Can we get this upstreamed to vishvananda/netlink? Ok, short term to be here but should see if we can move it to netlink lib somehow.

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+	native.PutUint32(b, uint32(route.LinkIndex))
+
+	req.AddData(nl.NewRtAttr(syscall.RTA_OIF, b))
+
+	_, err := req.Execute(syscall.NETLINK_ROUTE, 0)
+	return err
+}
+
+func (h *liveOpHandler) Sysctl(item, value string) error {
+	o, err := exec.Command("sysctl", "-w", fmt.Sprintf("%s=%s", item, value)).CombinedOutput()
```
eyakubovich: This is very to implement without calling out to `sysctl`. Just open `/proc/sys/...` and write a value. `vxlan` backend does tweak sysctls as well and has a function for that. Maybe move it into `pkg`.

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+}
+
+//*****************************************************************************
+
+func (dev *ovsDevice) vsCtl(cmd string, args ...string) error {
+	allArgs := make([]string, 0, 1+len(args))
+	allArgs = append(allArgs, cmd)
+	for _, a := range args {
+		allArgs = append(allArgs, a)
+	}
```
eyakubovich: `allArgs := append([]string{cmd}, args...)`

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+}
+
+func (dev *ovsDevice) addFlow(flow string, args ...interface{}) error {
+	if len(args) > 0 {
+		flow = fmt.Sprintf(flow, args...)
+	}
+	log.Infof("OVS ADD FLOW: %s", flow)
+	out, err := dev.ofCtl("add-flow", flow)
+	if err != nil {
+		log.Errorf("Error adding OVS flow %s: %s (%v)", flow, out, err)
```
eyakubovich: It's better to just `return fmt.Errorf` this stuff and log higher up.

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+
+//*****************************************************************************
+
+func (dev *ovsDevice) addGatewayAddressToLink(linkName string) (netlink.Link, error) {
+	link, err := dev.opHandler.LinkByName(linkName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s: %v", linkName, err)
+	}
+	addrStr := fmt.Sprintf("%s/%d", dev.gwAddr.String(), dev.nodeNetwork.PrefixLen)
+	addr, err := netlink.ParseAddr(addrStr)
```
eyakubovich:
```
addr := netlink.Addr{
    net.IPNet: ip.IP4Net{
        dev.gwAddr,
        dev.nodeNetwork.PrefixLen
    }.ToIPNet()
}
```

and you don't need to go via parsing

backend/ovs/ovs.go
```diff
@@ -0,0 +1,187 @@
...
+	ovsb.wg.Add(1)
+	go func() {
+		subnet.WatchLeases(ctx, ovsb.sm, CLUSTER_NETWORK_NAME, nil, evts)
+		log.Infof("WatchNodeLeases exited")
+		ovsb.wg.Done()
+	}()
+
+	log.Errorf("rcrcrc - watching %s\n", CLUSTER_NETWORK_NAME)
+	initialEvtsBatch := <-evts
+	ovsb.handleNodeEvents(initialEvtsBatch)
```
eyakubovich: No need to special case initial event batch

eyakubovich:
@dcbw I'm still reviewing it. And I will also provide my feedback on the general approach.

backend/ovs/device.go
```diff
@@ -0,0 +1,661 @@
...
+		return nil
+	}
+
+	dev.vsCtl("del-br", dev.bridgeName)
+	dev.vsCtl("add-br", dev.bridgeName, "--", "set", "Bridge", dev.bridgeName, "fail-mode=secure")
+	dev.vsCtl("set", "bridge", dev.bridgeName, "protocols=OpenFlow13")
+	dev.vsCtl("del-port", dev.bridgeName, "vxlan0")
+	dev.vsCtl("add-port", dev.bridgeName, "vxlan0", "--", "set", "Interface", "vxlan0", "type=vxlan", "options:remote_ip=\"flow\"", "options:key=\"flow\"", "ofport_request=1")
+
+	link, err := dev.opHandler.LinkByName("br0")
```
eyakubovich: s/br0/dev.bridgeName

eyakubovich:
@dcbw Overall I think this backend is doing too much and is too coupled to Docker and other foreign concepts like services.

Can we reduce the scope to just creating `br0` and setting up `tun0` and `vxlan0`? Something downstream (e.g. CNI or a script) can hook up `lbr0` to it (if is necessary, is it really impossible to get Docker to work directly with OVS?).

This will give this backend semantics similar to others:. There's this endpoint that knows how to route the flannel network addresses within the fabric.

dcbw:
@eyakubovich in reviews on the corresponding OpenShift/Kube parts of this effort, a good point was brought up.  Which is that the orchestration manager parts of this are fairly tied to the flannel/backend parts, so they should probably all live in the same place.  eg, the flannel side (or appc/cni?) would provide the scripts that the orchestration side would need to call to connect the endpoints/containers to the infrastructure.  Part of the reason for that is that the OVS flow rules must be coordinated between the two parts, and the interfaces are obviously specifically named.

The lbr0/vovsbr/vlinuxbr parts are mainly for IPAM and secondarily for OpenShift image builder support.  We'd like to transition that to CNI IPAM in the near future instead of relying on docker, which should simplify this.  I'll look at moving our stuff to a CNI plugin and see if that makes things somewhat simpler.

eveliovila:
hi! any ideas when is this going to be released?

dcbw:
@eveliovila I'm reworking it substantially so it's going to be a bit before it can go through another review.

philips:
Hey @dcbw how is the rework going?

dcbw:
@philips still going in this general direction, but some of this is on hold while we sort out upstream Kubernetes plugin and multi-network issues, since this would eventually be used in conjunction with OpenShift and Kubernetes.  I have not forgotten about it...

dcbw:
Way out of date, just going to close this one for now.

PR #385, opened 2015-12-04
Fixes a number of races

eyakubovich:
Couple of race related bugs found by straneous
unit testing. Most of the problems are in
mocks/tests themselves.

### DIFF
--- a/subnet/mock_registry.go
+++ b/subnet/mock_registry.go
@@ -65,6 +65,7 @@ type event struct {
 }
 
 type MockSubnetRegistry struct {
+	mux           sync.Mutex
 	networks      map[string]*netwk
 	networkEvents chan event
 	index         uint64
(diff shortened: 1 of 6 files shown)


subnet/mock_registry.go
```diff
@@ -225,7 +250,16 @@ func (msr *MockSubnetRegistry) deleteSubnet(ctx context.Context, network string,
 }
 
 func (msr *MockSubnetRegistry) watchSubnets(ctx context.Context, network string, since uint64) (Event, uint64, error) {
+	msr.mux.Lock()
+	locked := true
+	defer func() {
```
xiang90: I feel just unlock where you want is more simpler.

xiang90:
The test fails due to a deadlock. Let's fix it.

eyakubovich:
Yeah, that's actually yet another race somewhere that I was going to do in another PR but looks like it has gotten worse so I need to track it down in this one.

subnet/mock_registry.go
eyakubovich: The `defer` is there because I had things like `return msr.index` and I needed that expression to be evaluated under lock. Otherwise I would need to introduce a temporary variable.

But looking at it again I see that I'm returning `msr.index` not under lock in one place. So I'll need to reorg the code anyway.

eyakubovich:
@xiang90 PTAL

xiang90:
LGTM

PR #398, opened 2016-01-24
[security] Fix UDP binding interface

smessmer:
Fixes https://github.com/coreos/flannel/issues/394 , a bug that makes the UDP backend bind to 0.0.0.0 instead of to the specified interface.

If flannel is running over a VPN but the nodes also have an external public non-VPN IP (on which flannel should not be reachable), this is a severe security issue.

### DIFF
--- a/backend/udp/network.go
+++ b/backend/udp/network.go
@@ -64,7 +64,7 @@ func newNetwork(name string, sm subnet.Manager, extIface *backend.ExternalInterf
 	}
 
 	var err error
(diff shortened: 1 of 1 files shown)


philips:
Thank you for the fix! I submitted it directly to master after cleaning up the commit message a bit: https://github.com/coreos/flannel/commit/6c95ca9086cf7f5fc6e1d415fb23c50aeec34c04

PR #417, opened 2016-03-12
mk-docker-opts: Remove bash dependency

glevand:
Rewrite the option handling to avoid bash specific features.  Can be processed by busybox shell support.
Removal of bash and its dependencies reduces the flannel docker image size by 5 MiB.


glevand:
I noticed my 1st version doesn't do quite the same thing as the original.  I'll push out an update when ready.

tomdee:
@glevand Any update on this? It would be great to be able to remove the dependency on bash.

PR #421, opened 2016-03-23
Convert vxlan port to big endian

hustcat:
When I use flannel create network with port 8472 , vxlan module will listen on UDP port 6177. We should convert vxlan port to big endian when create vxlan device. This PR fix this bug.

Signed-off-by: Ye Yin <eyniy@qq.com>

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -59,7 +59,7 @@ func newVXLANDevice(devAttrs *vxlanDeviceAttrs) (*vxlanDevice, error) {
 		VxlanId:      int(devAttrs.vni),
 		VtepDevIndex: devAttrs.vtepIndex,
 		SrcAddr:      devAttrs.vtepAddr,
-		Port:         devAttrs.vtepPort,
+		Port:         int(nl.Swap16(uint16(devAttrs.vtepPort))), //network endian order
 		Learning:     false,
 	}
 


philips:
@hustcat I am missing why only this field needs this. Can you explain it?

hustcat:
Here is some test case. When I  create vxlan device with vishvananda/netlink, the port from kernel is not  specified port:

```go
package main

import (
        "fmt"
        "github.com/vishvananda/netlink"
)

func main() {
        parent, err := netlink.LinkByName("eth0")
        if err != nil {
                fmt.Printf("Get host interface error: %v", err)
                return
        }

        vxlan := &netlink.Vxlan{
                LinkAttrs:    netlink.LinkAttrs{Name: "vxlan1"},
                VxlanId:      258,
                VtepDevIndex: parent.Attrs().Index,
                Port:         8472,
        }
        err = netlink.LinkAdd(vxlan)
        if err != nil {
                fmt.Printf("add vxlan interface error: %v", err)
        }
}
```

```sh
# go build vxlan.go
# ./vxlan

# ip -d link show vxlan1
51: vxlan1: <BROADCAST,MULTICAST> mtu 1450 qdisc noop state DOWN mode DEFAULT 
    link/ether f6:56:51:cb:da:85 brd ff:ff:ff:ff:ff:ff promiscuity 0 
    vxlan id 258 dev eth0 srcport 0 0 dstport 6177 nolearning ageing 300

# ip link set vxlan1 up
# netstat -ulnp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name 
udp        0      0 0.0.0.0:6177            0.0.0.0:*                           -             
```

As shown above, I want use 8472, but result is 6177.

hustcat:
From [here](https://github.com/torvalds/linux/blob/master/drivers/net/vxlan.c#L2932), kernel use big endian for port:
```c
	if (data[IFLA_VXLAN_PORT])
		conf.dst_port = nla_get_be16(data[IFLA_VXLAN_PORT]);
```

tomdee:
This has now been fixed upstream. Let's fix this by taking a new upstream version

https://github.com/vishvananda/netlink/pull/111

PR #428, opened 2016-04-20
Add Windows support

robdaemon:
This adds Windows support to flannel, primarily for the host-gw backend.
Utilizes the iphlpapi DLL using syscall (see pkg/windows) - this is
exactly how the Go sys packages are interacting with iphlpapi.

The UDP and VxLAN backends will raise an error if they attempt to be used
on Windows.

Adds a minor abstraction over netlink / iphlpapi so that very little
needed to be changed in the backend/hostgw package.

Adds a build.ps1 for compiling on Windows

Adds additional notes in the README with respect to Windows.


robdaemon:
If this PR is accepted, I'll make another PR that can register flannel as a Windows service.

glevand:
This is a lot of change for one patch.  I recommend you break it up with  each smaller patch a self contained update.

robdaemon:
The largest part of this is new code (to talk to the Windows API) - I'm not sure I understand how you'd like it broken up?

robdaemon:
@glevand Can I get some additional feedback on the changes you'd like to see in this PR? Thanks!

glevand:
@robdaemon As I said, it is a big patch, it would be better to review and to maintain if it was broken up.

robdaemon:
@glevand I can move the interaction with the iphlpapi to a separate project, but when I add it as a Godep, the effective size of this PR will remain the same. I'm struggling to understand how I can make this a smaller PR without it being an incomplete implementation of host-gw networking on Windows.

philips:
@robdaemon two commits would be helpful for review. One for the library code and a second one for the changes to flannel.

Also, would you maintain this overtime if the code gets merged?

Do you have more info on how this is being used or the motivation? Are you also able to test/maintain the etcd port to Windows? I assume you are also running etcd on Windows.

backend/hostgw/network.go
```diff
@@ -87,15 +87,24 @@ func (n *network) handleSubnetEvents(batch []subnet.Event) {
...
+				Destination: evt.Lease.Subnet.ToIPNet(),
+				Gateway:     evt.Lease.Attrs.PublicIP.ToIP(),
+				LinkIndex:   n.extIface.Iface.Index,
 			}
-			if err := netlink.RouteAdd(&route); err != nil {
+
+			if err := routes.AddRoute(&route); err != nil {
+				errno, ok := err.(syscall.Errno)
+				// The Windows errno for "The object already exists" is 0x1392
+				if ok && errno == syscall.EEXIST || errno == 0x1392 {
```
philips: Eh, this should be in a seperate package with build windows tags

backend/hostgw/network.go
```diff
@@ -106,12 +115,18 @@ func (n *network) handleSubnetEvents(batch []subnet.Event) {
...
-				LinkIndex: n.linkIndex,
+			route := routes.Route{
+				Destination: evt.Lease.Subnet.ToIPNet(),
+				Gateway:     evt.Lease.Attrs.PublicIP.ToIP(),
+				LinkIndex:   n.extIface.Iface.Index,
 			}
-			if err := netlink.RouteDel(&route); err != nil {
+			if err := routes.DeleteRoute(&route); err != nil {
+				errno, ok := err.(syscall.Errno)
+				if ok && errno == syscall.EEXIST {
```
philips: delete will return EEXIST?!

backend/hostgw/network.go
```diff
@@ -162,21 +177,21 @@ func (n *network) checkSubnetExistInRoutes() {
...
 				}
 			}
 		}
 	}
 }
 
-func routeEqual(x, y netlink.Route) bool {
-	if x.Dst.IP.Equal(y.Dst.IP) && x.Gw.Equal(y.Gw) && bytes.Equal(x.Dst.Mask, y.Dst.Mask) {
+func routeEqual(x, y routes.Route) bool {
+	if x.Destination.IP.Equal(y.Destination.IP) && x.Gateway.Equal(y.Gateway) && bytes.Equal(x.Destination.Mask, y.Destination.Mask) {
```
philips: why rename ?

philips: Please hold the rename for a separate commit or PR.

robdaemon: This isn't a rename as much as a creating of a Route object to abstract between netlink and iphlpapi. I didn't want to create a copy/paste of the entire network.go file, so I created a small wrapper around the routes.

backend/hostgw/network.go
robdaemon: In Windows. :(

backend/hostgw/network.go
robdaemon: Yeah, I was weighing that. I was trying to avoid a copy/paste of this entire file, but the small differences between Windows and Linux had to leak in if I don't do that.

robdaemon:
@philips - Thanks so much for the review! I'll redo it into two commits and move the iphlpapi wrapper to a separate project.

As for maintenance, it would be myself and other members of my team at HPE that would definitely maintain it. Our use-case is a Kubernetes cluster where we have pods on the Kubernetes minions, and workloads that are technically part of the same namespace but run on a Windows VM. We don't have a need to port etcd over to Windows, since we have the Kubernetes cluster running alongside it at this time.

robdaemon:
Closing this, this work was taken on by my colleagues at HPE.

PR #438, opened 2016-05-17
BUILDS: Use vendor directory instead of Godeps

tomdee:
Closes #407

PR #439, opened 2016-05-18
BUILD: Move to simple Makefile and make more use of Docker

tomdee:
Create a static flanneld binary in a Docker container
Also, run tests in a container.

Note: this relies on PR #438 being merged


Dockerfile
```diff
@@ -1 +1,7 @@
-FROM golang:1.4.1-onbuild
+FROM golang:1.6-alpine
+
+MAINTAINER Tom Denham <tom@tigera.io>
+
+RUN apk -U add bash git musl-dev gcc linux-headers
+WORKDIR /go/src/github.com/coreos/flannel
+
```
glevand: I'd like to see a Dockerfile that works for cross compiling arm64 builds.
Does alpine have aarch64 cross gcc?  If not, maybe golang:1.6-wheezy would be better.
Does golang:1.6-alpine have arm64 cgo support?

Makefile
```diff
@@ -0,0 +1,71 @@
...
+
+artifacts/flanneld: flannel_build.created $(shell find . -type f  -name '*.go')
+	mkdir -p artifacts
+	docker run --rm \
+	-v $(ROOT_DIR)/artifacts:/mnt/artifacts \
+	-v $(ROOT_DIR)://go/src/github.com/coreos/flannel:ro \
+	flannel_build bash -c '\
+		go build -o /mnt/artifacts/flanneld -ldflags "-extldflags -static \
+		-X github.com/coreos/flannel/version.Version=$(shell git describe --dirty)" ; \
+		chown -R $(shell id -u):$(shell id -u) /mnt/artifacts'
```
glevand: It seems we'll need to propagate the makefile go environment to the flannel_build container so it picks up GOARCH, etc.

glevand:
You are removing flannelbox.config, but I didn't see how you create the final flannel docker image.  What are your ideas?

Dockerfile
tomdee: That sounds cool, but I don't think it's a requirement for getting this PR merged. I'm happy to change golang:1.6-alpine to something else if you think that would make future cross compilation easier. However, it looks like alpine is probably fine - e.g. see https://github.com/tianon/rawdns/blob/master/Dockerfile.cross

Makefile
```diff
@@ -0,0 +1,71 @@
...
+
+flannel_build.created: Dockerfile
+	docker build -t flannel_build .
+	touch flannel_build.created
+
+artifacts/flanneld: flannel_build.created $(shell find . -type f  -name '*.go')
+	mkdir -p artifacts
+	docker run --rm \
+	-v $(ROOT_DIR)/artifacts:/mnt/artifacts \
+	-v $(ROOT_DIR)://go/src/github.com/coreos/flannel:ro \
```
luxas: `//go` maybe should be `/go`

luxas:
Here I've built `flannel` for Kubernetes for four different arches. 
I'll help with getting this into core.
See the contents in this dir: https://github.com/kubernetes/kubernetes/tree/master/cluster/images/flannel
Full proposal: kubernetes/kubernetes#17981

Dockerfile
glevand: @tomdee In that link they are disabling cgo support for the build, but flannel needs cgo.

I checked wheezy, and it does not have aarch64 compiler support.

I think an answer is to use a recent debian or ubuntu image and install the cross compiler packages (gcc-aarch64-linux-gnu and libc6-dev-arm64-cross), then use a gimme script that can install a go compiler with arm64 cgo.  See https://github.com/coreos/ignition/pull/235/commits/787b5104676d2477bd460eac343c1d307d7dc5a5

glevand:
@luxas It seems you are using this debian-iptables-xxx image as your base for the final flannel image.  The current flannel build uses buildroot to create the final docker image.  I would think your debian based image is considerably larger, and takes longer to start up?

luxas:
Yes, the ideal thing would be to use a buildrooted env, but it's much harder.
Kubernetes already had the `debian-iptables` image for `amd64`, so I'm cross-building that one instead, and using it as base. It is larger yes (~100MB), which is a bad thing of course.

If you've got buildroot configs for `armv6`, `arm64` and `ppc64le`, ping me and I'll take a look

glevand:
@luxas I have this https://github.com/glevand/buildroot--buildroot.  I used the master branch to create the arm64 flannel image I use for CoreOS. It should be straight forward to add defconfigs for armv6 and ppc64le.

tomdee:
So, just to make I'm following the discussion here...
1) I would like a have a standard/consistent way of building the flannel binary. This would make build (and releases) reproducible. The built binary would need to be static.
2) There's a desire to be able to build the binary for other platforms (e.g. armv6, arm64, ppc64le) but I don't think we could call them "supported" platforms
3) Flannel should be deployable in a container. But if it's a static binary, then the container isn't too important right?

For the building, can we not use the golang:onbuild container because of the need for CGO support?

ping @glevand @luxas

luxas:
We may use `gcr.io/google_containers/kube-cross:v1.6.2-2` as the building image like I do in: https://github.com/kubernetes/kubernetes/tree/master/cluster/images/flannel

The only thing we need in the container is `iptables`, and it's dynamically linked, so the container have to include `glibc` => we may use `busybox:glibc`, `armel/busybox:glibc`, `aarch64/busybox:glibc` and `ppc64le/busybox:glibc`

I think we should compile iptables from source, and put it in a busybo:glibc image along with a statically linked flannel binary

glevand:
@tomdee What do we need to make arm64  a "supported" platform?

I think the build system should at least have a way to output a container, or file system for a container.

@luxas Would just building busybox with iptables enabled be simpler?

tomdee:
A build based on kube-cross looks like the best option to me.

Packaging looks tricky though. I'm leaning towards the suggestion from @luxas 

> I think we should compile iptables from source, and put it in a busybo:glibc image along with a statically linked flannel binary

@glevand How would we go about building busybox with iptables enabled?

tomdee:
This should all be covered by #462 so closing.

PR #442, opened 2016-05-31
Remove race-condition when setting up masquerade rules

tomdee:
Using ACCEPT means that no other rules in the POSTROUTING chain can
fire. Using RETURN allows other rules to be run (but relies on default
being ACCEPT which is the norm)


tomdee:
See https://github.com/kubernetes/kubernetes/issues/20391

steveeJ:
I'd suggest to add a note in the docs that flannel relies on either an ACCEPT rule or default policy in the nat POSTROUTING chain.

Otherwise, LGTM

PR #444, opened 2016-06-02
DOCS: Add note to AWS docs about why it might be used

tomdee:
The biggest advantage of using Flannel AWS-VPC backend is that the AWS knows about that IP. That makes it possible to setup ELB to route directly to that container.

### DIFF
--- a/Documentation/aws-vpc-backend.md
+++ b/Documentation/aws-vpc-backend.md
@@ -2,6 +2,8 @@
 
 When running within an Amazon VPC, we recommend using the aws-vpc backend which, instead of using encapsulation, manipulates IP routes to achieve maximum performance. Because of this, a separate flannel interface is not created.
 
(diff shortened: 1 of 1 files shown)


emmanuel:
This sounds (potentially) very powerful, and I think I want it :). However, the APIs for ELB (the ones I've found, at least) are all based on registering *EC2 instances* with an ELB, not an arbitrary IP. I'm confused how to achieve the described config.

Is there any chance that you can point me to something that would help explain how to setup an ELB to route directly to a Pod IP?

rc-olx:
@emmanuel you are correct. You can't register an IP to an ELB, currently the supported targets have to be running EC2 instances. Even considering ECS you register an instance id with a port --https://docs.aws.amazon.com/cli/latest/reference/elbv2/register-targets.html

PR #459, opened 2016-07-13
BUILDS: Replace some shell scripts with Makefile

tomdee:


Makefile
```diff
@@ -0,0 +1,52 @@
+.PHONY: all test cover gofmt gofmt-fix license-check
+
+# Grab the absolute directory that contains this file.
+ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+# These variables can be overridden by setting an environment variable.
+PACKAGES?=pkg/ip subnet network remote
```
steveeJ: as long as there's no tests in `network` we should prevent it from being tested but only run `gofmt` and license checks against it

steveeJ:
LGTM with a minor nit

Issue #450, opened 2016-06-19
Docker build errors


mharj:
When checkout from git Docker seems to use golang:1.4.1-onbuild and fails
```
../golang.org/x/net/context/ctxhttp/ctxhttp.go:35: req.Cancel undefined (type *http.Request has no field or method Cancel)
../github.com/coreos/etcd/client/cancelreq.go:13: req.Cancel undefined (type *http.Request has no field or method Cancel)
```
If I change golang:1.5-onbuild (which have Cancel method)
```
../github.com/coreos/flannel/subnet/registry.go:79: not enough arguments in call to transport.NewTransport
```
If change to golang:1.6-onbuild ... seems to compile ok, so should this Dockerfile updated to use 1.6 instead?

tomdee:
Dockerfile is now updated

sunflower9973:
Still an error:
core@coreos01vt ~ $ docker run -i -t google/golang /bin/bash -c "go get github.com/coreos/flannel"

# github.com/coreos/flannel/subnet
../github.com/coreos/flannel/subnet/registry.go:79: not enough arguments in call to transport.NewTransport

nwjeeper33:
I am using the same docker command above, and got same exact error. Can you help please?
Is it related to etcd2 configuration, not installing CA on the box? Any feedback would be appreciated!!

migeorge:
@sunflower9973 @nwjeeper33 I had success building with the newer golang container:
`docker run -i -t golang:1.6 /bin/bash -c "go get github.com/coreos/flannel"`

The binary will be located at `/go/bin/flannel`

kdx99:
Many thanks, with the version 1.6 it worked !
(...
Status: Downloaded newer image for golang:1.6)

PR #460, opened 2016-07-14
BUILDS: Update release container

tomdee:
- Use Go 1.6
- Update to use make instead of build script


tomdee:
Fixes #450 and #454

tomdee:
Also superseded by #462 so closing

PR #461, opened 2016-07-15
mk-docker-opts.sh: replace with busybox shell compatible version

tomdee:
Tested with lots of different options and it works just fine.

It avoids us needing to ship bash in the container.


tomdee:
Closes #417

dist/mk-docker-opts.sh
```diff
@@ -11,7 +11,7 @@ OPTIONS:
 	-c	Output combined Docker options into DOCKER_OPTS var
 	-k	Set the combined options key to this value (default DOCKER_OPTS=)
 	-m	Do not output --ip-masq (useful for older Docker version)
-" >/dev/stderr
```
steveeJ: What's the reason for this change? AFAICT this isn't related from the sh->bash switch

steveeJ:
@tomdee 
Would be awesome to have a test that uses a predefined env file and tests against expected output options. Can I motivate you to add such a test?

dist/mk-docker-opts.sh
tomdee: I think /dev/stderr is a bashism and >&2 is more POSIX compliant. See https://lists.debian.org/debian-user/2009/07/msg00014.htmlfor some discussion on the subject.

tomdee:
I really don't think it's worth having tests for this. It's bad enough having this ugly piece of functionality, but adding a test and a way to run the test is only going to make it harder to remove. A better long term approach would be to wrap this functionality into flanneld itself, but it's not a good use of by time to do that at the moment.

steveeJ:
@tomdee I haven't touched this script (using or writing) at all. But I think that as long as it's hosted in this repository and users rely on this to use flannel with Docker it's worth to ensure it's functionality. E.g. it's used in CoreOS directly https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel/files/flanneld-rkt.service#L48, so we definitely want this to work as long as it's there.

tomdee:
@steveeJ I've added a test script. PTAL.

steveeJ:
@tomdee 

> I've added a test script. PTAL.

Fantastic, thanks! Shall we run it on travis?

dist/mk-docker-opts_tests.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+set -e
+
+# Check help text
```
steveeJ: I'm impressed by the effort, but I don't think we need to check the help output. Focusing on the generated output should be fine. Keeping the help output/test in sync doesn't seem worth the effort.

tomdee:
OK, updates pushed @steveeJ PTAL.

dist/mk-docker-opts.sh
```diff
@@ -72,23 +72,30 @@ if [ -n "$FLANNEL_IPMASQ" ] && [ $ipmasq = true ] ; then
...
 
 echo -n "" >$docker_env
-for opt in $(compgen -v DOCKER_OPT_); do
-	eval val=\$$opt
 
-	if [ "$indiv_opts" = true ]; then
-		echo "$opt=\"$val\"" >>$docker_env
+for opt in $(set | grep "DOCKER_OPT_"); do
+
+    OPT_NAME=$(echo $opt | awk -F "=" '{print $1;}');
```
eyakubovich: Indent doesn't match the rest

dist/mk-docker-opts.sh
steveeJ: Interesting, thanks for the reference!

dist/mk-docker-opts.sh
steveeJ: Nice catch @eyakubovich.
Looks like a typical case of space vs. tab indentation.

tomdee:
OK, indentation fixed. @steveeJ PTAL

steveeJ:
@tomdee I think this one is almost good to go. The last steps before merging IMO are:
* enable the mk-docker-opts-test in travis
* split the mk-docker-opts rewrite / test into a separate commit

tomdee:
It is already being tested on travis. I'm not sure what you mean by splitting it into a separate commit?

steveeJ:
@tomdee somehow I was tricked by my own thoughts, I actually meant the exact opposite: squashing the commits.

tomdee:
OK, squashed. I'm going to merge this when it goes green.

steveeJ:
LGTM

Issue #389, opened 2015-12-23
Build flannel statically


luxas:
Hello,
Would it be supported to build `flanneld` statically with the argument `--ldflags '-extldflags "-static"'` to `go build`?

I'm wondering if it would be possible to package `flannel` in an `alpine` container with `iptables`
That would decrease image size a lot
I'm testing it right now

philips:
@luxas Yea, that should work. Do you want to give us a PR that makes this change? I just tried it and it works.

```
diff --git a/build b/build
index f56cabf..7973006 100755
--- a/build
+++ b/build
@@ -18,7 +18,7 @@ function linker_dashX {
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/flannel"
 VERSION=$(git describe --dirty)
-GLDFLAGS="-X $(linker_dashX github.com/coreos/flannel/version.Version ${VERSION})"
+GLDFLAGS="-X $(linker_dashX github.com/coreos/flannel/version.Version ${VERSION}) -extldflags '-static'"

 if [ ! -h gopath/src/${REPO_PATH} ]; then
        mkdir -p gopath/src/${ORG_PATH}
```

luxas:
Yeah, I can make the PR soon.

davidsiefert:
Be so nice to get this merged... +1 :smiley:

tomdee:
See also #439

tomdee:
This is now merged in master.

PR #462, opened 2016-07-15
BUILDS: Overhaul build process

tomdee:
- Move shell scripts to Makefile
- Build for multiple architectures
- Simplify release process


tomdee:
Needs #461 to be merged first (since busybox doesn't have bash)

tomdee:
Close, fixes or supersedes the following PRs #460 #439 #417 and maybe #201 
Close, fixes or supersedes the following Issues #454 #450 #389

tomdee:
Ping @luxas @glevand

Makefile
```diff
@@ -0,0 +1,139 @@
...
+	IMAGE_SUFFIX=-$(ARCH)
+endif
+GOARM=6
+KUBE_CROSS_TAG=v1.6.2-2
+
+dist/flanneld: $(shell find . -type f  -name '*.go')
+	go build -o dist/flanneld \
+	  -ldflags "-extldflags -static -X github.com/coreos/flannel/version.Version=$(TAG)"
+
+test: license-check gofmt
```
steveeJ: I must be missing something, so for the sake of educating me: is the current test approach only license and fmt checking?

Makefile
```diff
@@ -0,0 +1,139 @@
...
+
+cover:
+	# A single package must be given - e.g. 'PACKAGES=pkg/ip make cover'
+	go test -coverprofile cover.out $(PACKAGES_EXPANDED)
+	go tool cover -html=cover.out
+
+# Throw an error if gofmt finds problems.
+# "read" will return a failure return code if there is no output. This is inverted wth the "!"
+gofmt:
+	bash -c '! gofmt -d $(PACKAGES) 2>&1 | read'
```
steveeJ: This looks quite ugly, Will `gofmt` not return non-zero if it finds a problem?

Makefile
```diff
@@ -0,0 +1,139 @@
...
+# These variables can be overridden by setting an environment variable.
+TEST_PACKAGES?=pkg/ip subnet remote
+TEST_PACKAGES_EXPANDED=$(TEST_PACKAGES:%=github.com/coreos/flannel/%)
+PACKAGES?=$(TEST_PACKAGES) network
+PACKAGES_EXPANDED=$(PACKAGES:%=github.com/coreos/flannel/%)
+
+# Set the (cross) compiler to use for different architectures
+ifeq ($(ARCH),amd64)
+	CC=gcc
+	IMAGE_SUFFIX=
```
luxas: I'd prefer if you set `IMAGE_SUFFIX` to `-$(ARCH)` for all cases
And then do an `ifeq ($(ARCH),amd64)` and push without suffix as well for amd64

Makefile
luxas: First it checks for the license, then gofmt and finally runs the `go test` command below.
The license-check and gofmt targets are dependencies of the test target

luxas: @tomdee Why isn't cover a dep of test?

Makefile
```diff
@@ -0,0 +1,139 @@
...
+      --mounts=run-flannel,path=/run/flannel,readOnly=false:etc-ssl-etcd,path=/etc/ssl/etcd,readOnly=true:dev-net,path=/dev/net,readOnly=false \
+      dist/flanneld-$(ARCH)-$(TAG).aci
+
+docker-push: dist/flanneld-$(ARCH)-$(TAG).docker
+	docker push $(REGISTRY)/flannel$(IMAGE_SUFFIX):$(TAG)
+
+## Build an architecture specific static flanneld binary
+dist/flanneld-$(ARCH):
+    # Build for other platforms with ARCH=$ARCH make build
+    # valid values for $ARCH are amd64 amd arm64 and ppc64le
```
luxas: ... are `[amd64 arm arm64 ppc64le]`

Makefile
```diff
@@ -0,0 +1,139 @@
...
+	docker run -e CC=$(CC) -e GOARM=$(GOARM) -e GOARCH=$(ARCH) -it \
+	    -v ${PWD}:/go/src/github.com/coreos/flannel:ro \
+        -v ${PWD}/dist:/go/src/github.com/coreos/flannel/dist \
+	    gcr.io/google_containers/kube-cross:$(KUBE_CROSS_TAG) /bin/bash -c '\
+		cd /go/src/github.com/coreos/flannel && \
+		CGO_ENABLED=1 make -e dist/flanneld && \
+		mv dist/flanneld dist/flanneld-$(ARCH) && \
+		file dist/flanneld-$(ARCH) && \
+		chown -R $(shell id -u):$(shell id -u) dist'
```
luxas: I think is much safer to do outside the container.

Makefile
```diff
@@ -0,0 +1,139 @@
...
+
+## Build an architecture specific iptables binary
+dist/iptables-$(ARCH):
+	docker run -e CC=$(CC) -e GOARM=$(GOARM) -e GOARCH=$(ARCH) -it \
+            -v ${PWD}:/go/src/github.com/coreos/flannel:ro \
+            -v ${PWD}/dist:/go/src/github.com/coreos/flannel/dist \
+            gcr.io/google_containers/kube-cross:$(KUBE_CROSS_TAG) /bin/bash -c '\
+            curl http://www.netfilter.org/projects/iptables/files/iptables-1.4.21.tar.bz2 | tar -jxv && \
```
luxas: `curl -sSL` is more reliable I think. Also, it might be worth it to make the iptables version a variable

Makefile
```diff
@@ -0,0 +1,139 @@
...
+                --disable-shared \
+                --disable-devel \
+                --disable-nftables \
+                --enable-static \
+                --host=amd64 && \
+            make && \
+            cp iptables/xtables-multi /go/src/github.com/coreos/flannel/dist/iptables-$(ARCH) && \
+            cd /go/src/github.com/coreos/flannel && \
+            file dist/iptables-$(ARCH)  && \
+            chown -R $(shell id -u):$(shell id -u) dist'
```
luxas: Same here, preferably outside the container

luxas:
Looks overall good to me.
Minor nit: to avoid duplication (although it's little code), you could refactor the Dockerfiles to a Dockerfile.arch or .tmpl and sed it with the right values to a Dockerfile before building.

Makefile
tomdee: Not that I can see - look here for more info https://groups.google.com/forum/#!topic/golang-nuts/pdrN4zleUio

Makefile
tomdee: The tests always run with coverage. The `cover` target is just generating the html report, but it can only do that for a single package. Hence they are kept separate.

Makefile
tomdee: Yes, that makes sense, I'll do that.

Makefile
tomdee: Why do you think that?

Makefile
tomdee: Done

tomdee:
Thanks @luxas for the review, I've made markups and pushed another commit.

I prefer the duplication of multiple Dockerfiles to needing to create temp files and use sed, so I've left that as it is.

Makefile
steveeJ: I'd be in favor to pass UID and GID environment variables to the container for this purpose to prevent possible issues with user namespace setups.

Makefile
steveeJ: I could swear the next line wasn't there when I wrote that comment, apologies!

.gitignore
```diff
@@ -1,3 +1,8 @@
-flannel
-coverage
-bin
+dist/*.tar.gz
+dist/flanneld*
+dist/iptables-*
+dist/*.aci
+dist/*.docker
```
luxas: nit: Why not ignore the whole `dist`?

.gitignore
```diff
@@ -1,3 +1,8 @@
...
+cover.out
+.editorconfig
+.idea/
```
luxas: missing newline

Makefile
```diff
@@ -0,0 +1,146 @@
...
+	bash -c '! gofmt -d $(PACKAGES) 2>&1 | read'
+
+gofmt-fix:
+	gofmt -w $(PACKAGES)
+
+license-check:
+	./license-check.sh
+
+clean:
+	rm -f dist/flanneld*
```
luxas: Again, why not just `rm -rf dist` to make it look cleaner?

Makefile
```diff
@@ -0,0 +1,146 @@
...
+                --disable-devel \
+                --disable-nftables \
+                --enable-static \
+                --host=amd64 && \
+            make && \
+            cp iptables/xtables-multi /go/src/github.com/coreos/flannel/dist/iptables-$(ARCH) && \
+            cd /go/src/github.com/coreos/flannel && \
+            file dist/iptables-$(ARCH)'
+
+## Build a .tar.gz for the amd64 flanneld binary
```
luxas: Why not build tar.gz for all arches while we're at it?

steveeJ: Why is the architecture hardcoded in this case?

.gitignore
tomdee: Because it contains the mk-docker-opts.sh script

Makefile
tomdee: As above, because of the mk-docker-opts.sh script

Makefile
tomdee: The reasoning behind just having the amd64 one is that I don't think the .tgz is that useful. I was planning that we would attach the static flanneld for all platforms to the github releases (which is much more useful) and also the ACI. I was keeping the amd64 tar.gz for legacy reasons. If we add all the platforms then that's something that we'll need to support forever...

luxas: If you're uploading the flannel binary and ACI for all arches, then it's okay for me.
Probably worth leaving a comment about the decision though.

steveeJ: IMHO container images that include the binary + dependencies are the way to go. No other binaries or archives needed. Feel free to ignore my radical opinion on this one

luxas:
Great!

PR #465, opened 2016-07-18
Support quorum read option

### LINKED ISSUE #383: flannel should use quorum reads
To avoid odd situations of flannel not reusing last lease upon restart (see https://github.com/coreos/etcd/issues/3924), flannel should use quorum=true when reading from etcd to get linearizability.

mtanino:
To avoid odd situation of flannel not reusing last lease
upon restart, flannel should use quorum=true when reading
from etcd to get linearizability.

Fix #383

### DIFF
--- a/subnet/mock_etcd_test.go
+++ b/subnet/mock_etcd_test.go
@@ -127,7 +127,7 @@ func TestMockEtcd(t *testing.T) {
 	// Sanity tests for our mock etcd
 
 	// Ensure no entries yet exist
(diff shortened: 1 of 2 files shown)


eyakubovich:
LGTM

mtanino:
@tomdee @steveeJ 

Hi Tom, Stefan,
Could I get a review for the patch? Thank you for your attention.

tomdee:
LGTM - merging

mtanino:
@tomdee 

Thank you for your help.

PR #466, opened 2016-07-19
build with rkt and ignore the artifacts folder

rbucker:


build-rkt
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+export SRC_DIR=`pwd`
+
+sudo rkt run \
+    --volume src-dir,kind=host,source=${SRC_DIR} \
+    --volume build-dir,kind=host,source=${SRC_DIR} \
+    --interactive \
```
steveeJ: not needed in this case

steveeJ:
@lucab mind taking a look?

build-rkt
rbucker: please ellaborate

.gitignore
```diff
@@ -1,3 +1,4 @@
+artifacts
```
tomdee: The artifacts directory is being removed

build-rkt
steveeJ: The `--interactive` argument is not needed here, as the build process will not require human interaction. I hope I'm not wrong with the latter ;-)

rbucker: On Tue, Jul 19, 2016 at 5:50 PM, Stefan Junker <notifications@github.com>
wrote:

> he --interactive argument is not needed here, as the build process will
> not require human interaction. I hope I'm not wrong with the latter ;-)
>

probably correct. I'm indifferent. This is what was presented to me as how
rkt was built withing rkt. I did not question it; just passed it along

*--Richard Bucker *

*http://about.me/richard.bucker <http://about.me/richard.bucker>*

lucab:
Hey @rbucker, it looks you are trying to retrofit rkt-builder as a generic build environment. I'm not sure if there is a bit of misunderstanding here, at that aci is meant to "build rkt" and not "build via rkt"; as such, it is custom tailored to rkt needs (which may be the same as flannel at the moment, just by chance).

I would not recommend going this way, as it requires some ugly hacks and we can easily find ourself entangled in diverging requirements.

On the other hand, I have no idea about how flannel is currently built (there are multiple reference to a missing ./build script) and what needs to be changed, but I guess there is some margin for improvements.

rbucker:
Since I'm not a regular user or DEV on the project I do not have a strong
opinion one way or the other. I can offer that since I use ChromeOS devices
for my UI and SSH into my CoreOS boxes for development having a container
based build system gets me one step closer to idempotent CI/CD. While I
prefer rkt to docker I think having both types of build agents would be
ideal. As for using rkt-builder... that is what was recommended when I
provided the same changes to acbuild and docker2aci. It's possible that
there is a better image implementation for flannel, however, I'm not the
one to build it. Maybe someone with CoreOS knowledge could float a common
build image upstream.

*--Richard Bucker *

*http://about.me/richard.bucker <http://about.me/richard.bucker>*

PR #483, opened 2016-08-04
add a subnet manager backed by the kubernetes apiserver

mikedanese:
Still figuring out glide... It builds in my gopath :) but havent' tested it end to end yet.

Requires an etcd client bump and a whole lot of new dependencies of the kubernetes client.

cc @aaronlevy  @luxas


Documentation/kube-flannel-cfg.yml
```diff
@@ -0,0 +1,22 @@
...
+  namespace: kube-system
+  labels:
+    tier: node
+    app: flannel
+data:
+  cni-conf.json: |
+    {
+      "name": "cbr0",
+      "type": "flannel"
+    }
```
aaronlevy: I believe we will need:

```
{
  "name": "cbr0",
  "type": "flannel",
  "delegate": {
    "isDefaultGateway": true
  }
}
```

ref: https://github.com/containernetworking/cni/pull/50

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+	"time"
+
+	"github.com/coreos/flannel/pkg/ip"
+	"github.com/coreos/flannel/subnet"
+
+	"github.com/golang/glog"
+	"golang.org/x/net/context"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/client/cache"
+	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/internalclientset"
```
aaronlevy: Random question: When do you want to use versioned clientset package vs the internal clientset?

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+	cfg, err := restclient.InClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize inclusterconfig: %v", err)
+	}
+	c, err := clientset.NewForConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize client: %v", err)
+	}
+
+	if os.Getenv("NODE_NAME") == "" {
```
aaronlevy: Given that this still requires an un-merged upstream change, what do you think about also trying to determine nodename using downward api namespace/pod name (then retrieving nodeName from pod object)?

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, fmt.Errorf("node %q pod cidr not assigned", ksm.nodeName)
+	}
+	bd, err := attrs.BackendData.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+	if n.Annotations[backendDataAnnotation] != string(bd) ||
+		n.Annotations[backendTypeAnnotation] != attrs.BackendType ||
+		n.Annotations[backendPublicIPAnnotation] != attrs.PublicIP.String() {
+		n, err = ksm.client.Core().Nodes().Get(ksm.nodeName)
```
aaronlevy: Why re-retrieve node here? Minimize the chance that nodestore is stale?

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+	time.Sleep(time.Second)
+	leases := make([]subnet.Lease, 0)
+	nl, err := ksm.nodeStore.List()
+	if err != nil {
+		return subnet.LeaseWatchResult{}, err
+	}
+	for _, n := range nl.Items {
+		l, err := nodeToLease(n)
+		if err != nil {
+			glog.Infof("error turning node to lease: %v", err)
```
aaronlevy: nit: adding node.name to err might help in debug scenarios

subnet/kube/kube.go
luxas: Voting for ns/podname + lookup initially

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, fmt.Errorf("NODE_NAME not set")
+	}
+
+	var cniConf, netConf []byte
+	cniConf, err = ioutil.ReadFile(cniConfPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read cni conf: %v", err)
+	}
+	netConf, err = ioutil.ReadFile(netConfPath)
+	if err != nil {
```
luxas: A default `NetConf` file should exist, it shouldn't be dependent on the ConfigMap

subnet/kube/kube.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, fmt.Errorf("failed to read net conf: %v", err)
+	}
+
+	err = ioutil.WriteFile(cniConfTargetPath, []byte(cniConf), 0644)
```
luxas: Why are you reading/writing the cni netconf?
It's not used in the flannel image, but with kubelet

So I don't see the point of it here, but if it's required, you probably just should copy from cniConf to cniConfTargetPath

subnet/kube/kube.go
luxas: The even better solution would be that after flannel knows in which pod it runs, it could just fetch the values of it's annotations (with default values) and create a NetConf from those values.

luxas:
@tomdee @steveeJ is it possible to get this into the v0.6.0 release?
It will probably be required for v1.4 turnup UX...

subnet/kube/kube.go
mikedanese: I don't know if this works yet. We at least need "Network" (subnet for pods). We can conceivably do this once we configure kube-controller-manager with componentconfig and that info is stored in the API.

mikedanese:
Ok, I think I got glide working

subnet/kube/kube.go
mikedanese: internalclientset is a way of weining people off the unversioned not generated client (old one). We can move to versioned clientsets once the controller.Framework supports them.

subnet/kube/kube.go
mikedanese: That was my thinking although I did not test to see if this would be an issue. Node objects are frequently updated because of the heartbeat.

subnet/kube/kube.go
mikedanese: Please see the daemonset config. We share /etc/cni/net.d with the host so writing this file tells kubelet to use the cni plugin to provision pod nets.

subnet/kube/kube.go
mikedanese: Done.

subnet/kube/kube.go
mikedanese: ok, looks like progress on nodeName has slowed unfortunately

subnet/kube/kube.go
mikedanese: Probably don't need this...

mikedanese:
@luxas @aaronlevy PTAL

Documentation/kube-flannel-ds.yml
```diff
@@ -0,0 +1,48 @@
...
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: flannel
+    spec:
+      hostNetwork: true
+      containers:
+      - name: kube-flannel
+        image: gcr.io/mikedanese-k8s/flannel:v9
```
aaronlevy: Might need to replace this with place-holder. Or a follow up PR once there is a tagged (or built from hash) release image. Just making note, but prob up to maintainers on preference.

subnet/kube/kube.go
```diff
@@ -0,0 +1,286 @@
...
+const (
+	resyncPeriod = 5 * time.Minute
+
+	subnetKubeManagedAnnotation = "flannel.alpha.coreos.com/kube-subnet-manager"
+	backendDataAnnotation       = "flannel.alpha.coreos.com/backend-data"
+	backendTypeAnnotation       = "flannel.alpha.coreos.com/backend-type"
+	backendPublicIPAnnotation   = "flannel.alpha.coreos.com/public-ip"
+
+	cniConfPath       = "/etc/kube-flannel/cni-conf.json"
+	cniConfTargetPath = "/etc/cni/net.d/10_flannel.conf"
```
aaronlevy: convention nit: s/10_flannel/10-flannel/

subnet/kube/kube.go
```diff
@@ -0,0 +1,286 @@
...
+	podName := os.Getenv("POD_NAME")
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podName == "" || podNamespace == "" {
+		return nil, fmt.Errorf("env variables POD_NAME and POD_NAMESPACE must be set")
+	}
+
+	pod, err := c.Pods(podNamespace).Get(podName)
+	nodeName := pod.Spec.NodeName
+	if nodeName == "" {
+		return nil, fmt.Errorf("node name not est in pod spec '%s/%s'", podNamespace, podName)
```
luxas: node name can't be calculated from pod spec %s/%s

subnet/kube/kube.go
luxas: Alright, missed that `hostPath` mount.

subnet/kube/kube.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	var cniConf, netConf []byte
+	cniConf, err = ioutil.ReadFile(cniConfPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read cni conf: %v", err)
```
luxas: If cniconf doesn't exist, it should just pick the default:
```go
defaultCniConf := []byte(`
{
    "name": "cbr0",
    "type": "flannel",
    "delegate": {
      "isDefaultGateway": true
    }
}`)

I think the ConfigMap is great for customization, but the flannel daemonset should do just fine without it.

File: subnet/kube/kube.go (hunk 6 above)
Comment (luxas) [inline subnet/kube/kube.go]: I meant that it initially could read from the pod's (aka from the daemonset metadata) own annotations, e.g.
flannel.alpha.coreos.com/backend-type would default to vxlan
flannel.alpha.coreos.com/network-cidr would default to 10.1.0.0/16 or 10.244.0.0/16 or whatever.
Later, it should try to read the cidr option from controller-manager like you said.

It would then be easy to customize the values by (`kubectl apply`) instead of using the configmap
```yaml
spec:
  template:
    metadata:
      labels:
        tier: node
        app: flannel
      annotations:
        flannel.alpha.coreos.com/network-cidr: 10.155.0.0/16
        flannel.alpha.coreos.com/backend-type: udp
```

subnet/kube/kube.go
mikedanese: We at least need the cluster cidr to run so for now we need a config map.

subnet/kube/kube.go
aaronlevy: The ultimate source of truth for the network cidr is going to be the `--cluster-cidr` in the controller-manager. Eventually we should be using that configuration (as @mikedanese mentions, we can do this once we are sourcing that info from api based config).

I think in the interim, I could go either way. Sourcing directly from a configMap makes it pretty straight-forward / easy to understand. Or, if we go with the annotations, I think we should just leave the whole config as raw JSON in a single annotation, because there are other options a user might want to configure (https://github.com/coreos/flannel#example-configuration-json).

subnet/kube/kube.go
```diff
@@ -0,0 +1,286 @@
...
+		n.Annotations[backendTypeAnnotation] = attrs.BackendType
+		n.Annotations[backendDataAnnotation] = string(bd)
+		n.Annotations[backendPublicIPAnnotation] = attrs.PublicIP.String()
+		n.Annotations[subnetKubeManagedAnnotation] = "true"
+		n, err = ksm.client.Core().Nodes().Update(n)
+		if err != nil {
+			return nil, err
+		}
+	}
+	_, cidr, err := net.ParseCIDR(n.Spec.PodCIDR)
```
aaronlevy: Unlikely error, but should we verify this prior to updating the node object annotations? If my understanding is correct, once the annotations are updated it is effectively a successful lease from the perspective of other nodes.

aaronlevy: It looks like we do the same check in `nodeToLease` so if the podCidr didn't actually parse, other nodes would ignore it.

aaronlevy:
Only a few minor comments. I'll start testing this out using the documented objects.

Ping @steveeJ @tomdee @eyakubovich -- as maintainers mind taking a pass on this?

subnet/kube/kube.go
mikedanese: I can add a default cni conf.

subnet/kube/kube.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func (ksm *kubeSubnetManager) RenewLease(ctx context.Context, network string, lease *subnet.Lease) error {
+	// noop
+	glog.V(4).Infof("renew lease")
+	return nil
+}
+
+func (ksm *kubeSubnetManager) WatchLease(ctx context.Context, network string, sn ip.IP4Net, cursor interface{}) (subnet.LeaseWatchResult, error) {
+	glog.V(4).Infof("watch lease")
```
aaronlevy: Started testing this, and both "watch lease" and "watch leases" are spamming logs, and not particularly helpful (granted it's a v4 log -- so maybe just a matter of changing the default documented daemonset to be less verbose

mikedanese: I will remove these logs

mikedanese:
Addressed comments

tomdee:
I can review this but it won't be until next week.

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil, fmt.Errorf("node name not present in pod spec '%s/%s'", podNamespace, podName)
+	}
+
+	var cniConf, netConf []byte
+	if _, err := os.Stat(cniConfPath); os.IsNotExist(err) {
+		cniConf = defaultCniConf
+	} else {
+		cniConf, err = ioutil.ReadFile(cniConfPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read cni conf: %v", err)
```
luxas:
```go
glog.Warning("failed to read cni conf: %v", err)
cniConf = defaultCniConf
```

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	podName := os.Getenv("POD_NAME")
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podName == "" || podNamespace == "" {
+		return nil, fmt.Errorf("env variables POD_NAME and POD_NAMESPACE must be set")
+	}
+
+	pod, err := c.Pods(podNamespace).Get(podName)
+	nodeName := pod.Spec.NodeName
```
luxas: @mikedanese @aaronlevy 
Since we have the node name here, could we as a intermediate step while waiting for componentconfig parse the two first bytes in ``.spec.PodCIDR` and make a `/16` network of it?

While I think the configmap is good, I also think that the one-line `kubectl apply -f https://k8s.io/flannel-network.yaml` is more important, and should
a) default to something (like weave does [here](https://github.com/weaveworks/weave-kube/blob/master/weave-daemonset.yaml#L22)) along with the `vxlan` backend.
b) guess it and hope for the best
c) entirely skip the configmap and use annotations on the Pod instead, since it is only read once (updating the configmap doesn't update the flannel conf anyway)

WDYT?

subnet/kube/kube.go
mikedanese: If we specified a file in a configmap, we should hard fail if it can't be parsed.

luxas: Ah, yea, thinking about it that way makes sense

subnet/kube/kube.go
aaronlevy: > parse the two first bytes in `.spec.PodCIDR and make a /16 network of it? 

I don't think we can safely assume the `cluster-cidr` size just based on the podCidr.

> While I think the configmap is good, I also think that the one-line kubectl apply -f https://k8s.io/flannel-network.yaml is more important

I agree ease of use should be a priority, but I'd be more or less fine with either configMap or annotation. 

I think I'd lean toward annotation (on daemonset) right now just for the sake of easy-of-deployment. But longer term we are going to need to source info from configMaps regardless (e.g. kcm `--cluster-cidr` should be source of truth for flannel network cidr). Also, "configMap + component" is going to be a really common deployment pattern -- so not opposed to following that right now.

Really we're just talking about the [backend options](https://github.com/coreos/flannel#backends). In general these options should not be changed after deployment (not even sure why you would want to) -- but neither configMap nor annotations would protect against this.

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil, fmt.Errorf("unable to initialize client: %v", err)
+	}
+
+	podName := os.Getenv("POD_NAME")
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podName == "" || podNamespace == "" {
+		return nil, fmt.Errorf("env variables POD_NAME and POD_NAMESPACE must be set")
+	}
+
+	pod, err := c.Pods(podNamespace).Get(podName)
```
aaronlevy: Can you log error here? Otherwise api connectivity errors give generic "node name not in pod spec" error.

subnet/kube/kube.go
luxas: @mikedanese I let you decide about annotation vs configmap in this initial PoC, but it would be nice to have something easy in the beginning, just to kick the tires.

At least it would be great to finish this PR (fixing Aarons last comment), and maybe build a unofficial release to `gcr.io` with v0.6.0 flannel + this PR.

I can do that if we want to

luxas:
@mikedanese Should it validate the minimum k8s version maybe?
It seems worthwhile to do, and since https://github.com/kubernetes/kubernetes/pull/27880#issuecomment-241068443 is on it's way in, we maybe can use it and require `v1.4.0-alpha.3+`.
WDYT?

main.go
```diff
@@ -73,6 +75,7 @@ func init() {
 	flag.StringVar(&opts.remoteKeyfile, "remote-keyfile", "", "SSL key file used to secure client/server communication")
 	flag.StringVar(&opts.remoteCertfile, "remote-certfile", "", "SSL certification file used to secure client/server communication")
 	flag.StringVar(&opts.remoteCAFile, "remote-cafile", "", "SSL Certificate Authority file used to secure client/server communication")
+	flag.BoolVar(&opts.kubeSubnetMgr, "kube-subnet-mgr", false, "Contact the Kubernetes API for subnet assignement instead of etcd or flannel-server.")
```
luxas: @tomdee @steveeJ ^

mikedanese:
@luxas I think that's a good idea if this doesn't get reviewed soon.

mikedanese:
@luxas we are not using that feature currently but I think that's a good idea.

luxas:
Yes, do you have bandwidth to update this PR with the (hopefully) last comments today?
 - k8s compability check (at least v1.4.0-alpha.2 or 3)
 - log nodename == "" error
 - annotation vs configmap

I can make the PR with v0.6.0 + this PR cherrypicked to `gcr.io`

mikedanese:
> log nodename == "" error

What are you referring to?

> annotation vs configmap

I think we should only support configmap initially.  This is exactly the usecase that configmap was designed for. I don't see a benefit to using annotations.

luxas:
> log nodename == "" error

Sorry if it was unclear: https://github.com/coreos/flannel/pull/483#discussion_r74844847

> I think we should only support configmap initially. This is exactly the usecase that configmap was designed for. I don't see a benefit to using annotations.

OK, but can we package them in the same file then for easy flannel turnup when running `kubectl apply -f https://k8s.io/turnup/network/flannel.yaml` for example?
(But I know that's an implementation question, not direct coding question for this PR)

mikedanese:
> Sorry if it was unclear: #483 (comment)

@luxas That is fixed?

> OK, but can we package them in the same file then for easy flannel turnup

Yes you can have a list of resources in a single file.

mikedanese:
@luxas I combined the config to a single file.

aaronlevy:
FWIW we very likely will backport the required patches so that this will work -- so a version check could block us from using this even though it would work. Wouldn't be opposed to a log message warning though.

ping @tomdee @eyakubovich @steveeJ -- any chance we can get a review?

aaronlevy:
@luxas per release notes (https://github.com/coreos/flannel/releases/tag/v0.6.0) the release images were broken in 0.6.0. If you are going to build this out of band, maybe use 0.6.1 /cc @tomdee

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+		},
+		&api.Node{},
+		resyncPeriod,
+		framework.ResourceEventHandlerFuncs{},
+	)
+	return &ksm, nil
+}
+
+func (ksm *kubeSubnetManager) GetNetworkConfig(ctx context.Context, network string) (*subnet.Config, error) {
+	time.Sleep(time.Second)
```
eyakubovich: Why sleep here?

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+		Expiration: time.Now().Add(24 * time.Hour),
+	}, nil
+}
+
+func (ksm *kubeSubnetManager) RenewLease(ctx context.Context, network string, lease *subnet.Lease) error {
+	return nil
+}
+
+func (ksm *kubeSubnetManager) WatchLease(ctx context.Context, network string, sn ip.IP4Net, cursor interface{}) (subnet.LeaseWatchResult, error) {
+	time.Sleep(time.Second)
```
eyakubovich: And why sleep here?

subnet/kube/kube.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return subnet.LeaseWatchResult{}, err
+	}
+	return subnet.LeaseWatchResult{
+		Snapshot: []subnet.Lease{l},
+	}, nil
+}
+
+func (ksm *kubeSubnetManager) WatchLeases(ctx context.Context, network string, cursor interface{}) (subnet.LeaseWatchResult, error) {
+	time.Sleep(time.Second)
```
eyakubovich: And why sleep here?

subnet/kube/kube.go
aaronlevy: I think we should be able to drop this. Looks to be just init step - and we would only support the default network in this mode.

mikedanese:
No sleeps should be necessary. Leftover cruft from debugging and now removed.

aaronlevy:
Without the sleeps will the watch functions just be called repeatedly? If so, maybe an option is to add handler functions to the informer which just throw events on a channel, which we can pull/block from in the watch functions?

mikedanese:
All watches already sleep a second between retries.

mikedanese:
Oops I take that back.

subnet/kube/kube.go
mikedanese: Since the subnet manager doesn't support incremental updates, we need to ratelimit full relists to avoid unnecessary busy looping.

subnet/kube/kube.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil, fmt.Errorf("env variables POD_NAME and POD_NAMESPACE must be set")
+	}
+
+	pod, err := c.Pods(podNamespace).Get(podName)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving pod spec for '%s/%s': %v", podNamespace, podName, err)
+	}
+	nodeName := pod.Spec.NodeName
+	if nodeName == "" {
+		return nil, fmt.Errorf("node name not present in pod spec '%s/%s': %v", podNamespace, podName, err)
```
aaronlevy: nit: sorry I may have missed this in my previous comment (which this would have addressed). But err will always be empty here.

subnet/kube/kube.go
luxas: Does calico and weave also do this?
I guess we should be consistent about how a network daemonset behaves...

main.go
steveeJ: +1 for not making this a special thing but rather make this an option among others.

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,74 @@
...
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: flannel
+    spec:
+      hostNetwork: true
+      containers:
+      - name: kube-flannel
+        image: gcr.io/mikedanese-k8s/flannel:v9
```
steveeJ: This needs to be changed to something more official. @eyakubovich @tomdee please assist in getting the automatic quay builds online again

subnet/kube/kube.go
steveeJ: Just for my education, where is the work about generating clients happening and how does a consumer - which flannel will be - know that the client needs to be updated? Shouldn't that be somewhat tied to the API?

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,74 @@
...
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kube-flannel-cfg
+  namespace: kube-system
+  labels:
+    tier: node
+    app: flannel
+data:
+  cni-conf.json: |
```
steveeJ: Where is it declared and by what consumed that this files gets written to _/etc/kube-flannel_?

subnet/kube/kube.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil, fmt.Errorf("error retrieving pod spec for '%s/%s': %v", podNamespace, podName, err)
+	}
+	nodeName := pod.Spec.NodeName
+	if nodeName == "" {
+		return nil, fmt.Errorf("node name not present in pod spec '%s/%s'", podNamespace, podName)
+	}
+
+	var cniConf, netConf []byte
+	if _, err := os.Stat(cniConfPath); os.IsNotExist(err) {
+		cniConf = defaultCniConf
```
steveeJ: I'd require a configuration file to be existent at this point. We can distribute one with flannel but I'm against writing one at runtime.

subnet/kube/kube.go
steveeJ: What's this method called by? Would it make sense to to only sleep when an error indicates that the _system_ is not ready yet?

subnet/kube/kube.go
```diff
@@ -0,0 +1,293 @@
...
+	}, nil
+}
+
+func (ksm *kubeSubnetManager) RenewLease(ctx context.Context, network string, lease *subnet.Lease) error {
+	return nil
+}
+
+func (ksm *kubeSubnetManager) WatchLease(ctx context.Context, network string, sn ip.IP4Net, cursor interface{}) (subnet.LeaseWatchResult, error) {
+	time.Sleep(time.Second)
+	nobj, found, err := ksm.nodeStore.Store.GetByKey(ksm.nodeName)
```
steveeJ: Can't we use the functionality from https://github.com/kubernetes/kubernetes/blob/master/pkg/client/cache/reflector.go to get real watching?

philips:
Sorry for the delay, I was traveling.  

I am OK with this approach overall. I would want two things:

1. Can you write a couple of paragraphs on the theory of operation for users to read and developers to understand the flow?
2. Can you work out an alternative to read/writing the CNI configuration through the manager? Even reading the discussion I don't understand why that needs to be done.

subnet/kube/kube.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize inclusterconfig: %v", err)
+	}
+	c, err := clientset.NewForConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize client: %v", err)
+	}
+
+	podName := os.Getenv("POD_NAME")
+	podNamespace := os.Getenv("POD_NAMESPACE")
```
aaronlevy: https://github.com/kubernetes/kubernetes/pull/27880 landed, maybe switch this back to using nodeName?

luxas: Was thinking about that, but that would force v1.4.0-alpha.3+ (which soon is gonna be released)

subnet/kube/kube.go
tomdee: I'd be interested in an update on this too. One of the less nice bits of this PR is that fact that it adds >7,000 files to the repo and increases the size of the flanneld binary from ~20MB to ~60MB.

Documentation/kube-flannel.yml
tomdee: I can try to find some time to help with that. But it's not clear to me how that helps. It's always a bit tricky having config in a repo that builds to versioned builds of that repo...

glide.yaml
```diff
@@ -72,3 +72,14 @@ import:
   subpackages:
   - compute/metadata
   - internal
+- package: k8s.io/kubernetes
+  version: e4a1ca7105fa9a3075fbf13a3286f8dddd59a658
```
tomdee: Should this be a tagged version? Is there any significance behind this particular version?

tomdee:
So, this all looks very cool to me and I support it!

The upside:
-  It makes flannel easier to deploy on kubernetes which is a really good thing.
-  It's a small amount of code (ignoring deps), and it's doesn't interact with any existing function - there's no big refactor required or anything like that.

Downsides are:
-  it makes the vendor directory much bigger and harder to manage
-  it make the flanneld binary _much_ bigger (20 MB -> 60 MB)
-  (Both of the above concerns could go away as kubernetes get their client libraries sorted out)
-  It adds a _little_ more code to maintain.
-  It adds a new way of deploying flannel which may need testing every release

To me, the upsides outweigh the downsides so it gets a "looks good in principle" from me.

Before it gets merged to master
* there are a few comments and questions to address
* the CNI related parts need rethinking (and probably removing from the go code)
* it needs documentation

aaronlevy:
An option for removing the cni config write might just be a side car container with something like:

```
      containers:
      - name: install-cni
        image: quay.io/coreos/flannel
        command: [ "/bin/bash", "-c", "cp", "/etc/kube-flannel/cni-conf.json", "/etc/cni/net.d/10-flannel.conf", "while", "true;", "sleep 3600;", "done" ]
        volumeMounts:
        - name: cni
          mountPath: /etc/cni/net.d
        - name: flannel-cfg
          mountPath: /etc/kube-flannel/
      - name: kube-flannel
        image: quay.io/coreos/flannel
        command: [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr" ]
```

philips:
@aaronlevy side car seems OK to me.  Although, we shouldn't do a copy like that but instead we shoudl copy to .tmp-$random-config, then move

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,74 @@
...
+    {
+      "name": "cbr0",
+      "type": "flannel",
+      "delegate": {
+        "isDefaultGateway": true
+      }
+    }
+  net-conf.json: |
+    {
+      "Network": "10.244.0.0/16",
```
ddysher: What's the relationship between Network and cluster-cidr flag in controller manager?  Do they have to match each other?

mikedanese: They should match.


On Aug 27, 2016 2:50 AM, "Deyuan Deng" <notifications@github.com> wrote:

In Documentation/kube-flannel.yml
<https://github.com/coreos/flannel/pull/483#discussion_r76513697>:

> +  namespace: kube-system
> +  labels:
> +    tier: node
> +    app: flannel
> +data:
> +  cni-conf.json: |
> +    {
> +      "name": "cbr0",
> +      "type": "flannel",
> +      "delegate": {
> +        "isDefaultGateway": true
> +      }
> +    }
> +  net-conf.json: |
> +    {
> +      "Network": "10.244.0.0/16",

What's the relationship between Network and cluster-cidr flag in controller
manager? Do they have to match each other?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/coreos/flannel/pull/483/files/f1cf895dfd2433cd98b9316729c25a9693f5655b..f52fee7cfe1824e959e42bc439088af752dc06bd#r76513697>,
or mute the thread

subnet/kube/kube.go
```diff
@@ -0,0 +1,293 @@
...
+	if podName == "" || podNamespace == "" {
+		return nil, fmt.Errorf("env variables POD_NAME and POD_NAMESPACE must be set")
+	}
+
+	pod, err := c.Pods(podNamespace).Get(podName)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving pod spec for '%s/%s': %v", podNamespace, podName, err)
```
aaronlevy: This can be follow up / don't want to block on this, but: We may want to have this in retry loop. In my tests it's pretty common to see this crashloop a few times during bootstrap. Not a big deal because it recovers, but to an operator the pod restart count might signal something is wrong, when it's really not.

luxas:
@mikedanese Maybe we can use `kubernetes/client-go` for talking to the apiserver here?
Maybe client go is leaner than the current vendored packages.

I'd like to get a working version of this in ASAP, as I have to submit the manifests to hyperkube before the v1.4 release, so please come back to this soon :)

mikedanese:
@luxas I don't think we can until pkg/controller/framework is moved to a versioned client.

main.go
mikedanese: What should I call the default subnet manager?

Documentation/kube-flannel.yml
luxas: Can you update this to `10.2.0.0/16`?
We thought that would be the best flannel CIDR (discussed in the kubeadm PR)

Documentation/kube-flannel.yml
luxas: Maybe update to `quay.io/coreos/flannel:v0.7.0-amd64`? (although it isn't released yet)

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,80 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        volumeMounts:
+        - name: run
+          mountPath: /run
+        - name: flannel-cfg
+          mountPath: /etc/kube-flannel/
+      - name: install-cni
+        image: quay.io/coreos/flannel
```
luxas: busybox should be sufficent

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,80 @@
...
+        command: [ "/bin/bash", "-c", "set -e -x; cp /etc/kube-flannel/cni-conf.json /tmp/flannel.cfg; mv /tmp/flannel.cfg /cni/net.d/10-flannel.conf; sleep infinity;" ]
```
luxas: maybe include a while loop in order to be safe

main.go
luxas: just `etcd`?

Documentation/kube-flannel.yml
mikedanese: Where was the discussion?

main.go
mikedanese: This is not what we did with the remote subnet manager so the cli flag api might be a little consistent. We use the remote subnet manager if --remote is passed.

subnet/kube/kube.go
mikedanese: I don't know if I understand the suggestion. This is called from here https://github.com/coreos/flannel/blob/master/subnet/watch.go#L262 inside the backend

subnet/kube/kube.go
mikedanese: We already are. The store is maintained by a reflector which is setup inside the framework.Informer. The framework.Informer has a superset of the functionality.

Documentation/kube-flannel.yml
mikedanese: It's declared in the volumeMounts field in the flannel daemonset.

subnet/kube/kube.go
mikedanese: Done.

glide.yaml
mikedanese: We need a v1.4 release. I can wait to use a v1.4 release once one is tagged (right now only betas are out).

mikedanese: Side note: it took way to long to update vendor (multiple hours). I have no idea what I'm doing with glide. Any straightforward doc on updating deps I should look at?

Documentation/kube-flannel.yml
```diff
@@ -0,0 +1,80 @@
...
+            fieldRef:
+              fieldPath: metadata.namespace
+        volumeMounts:
+        - name: run
+          mountPath: /run
+        - name: flannel-cfg
+          mountPath: /etc/kube-flannel/
+      - name: install-cni
+        image: busybox
+        command: [ "/bin/sh", "-c", "set -e -x; while true; do cp /etc/kube-flannel/cni-conf.json /tmp/flannel.cfg; mv /tmp/flannel.cfg /cni/net.d/10-flannel.conf; sleep 10; done;" ]
```
aaronlevy: fwiw I believe @philips suggested `cp` then `mv` based on `mv`(rename) being atomic if moving from same device. In some cases /tmp may not be same device and kind of defeats purpose. Maybe:

```
cp /etc/kube-flannel/cni-conf.json /cni/net.d/.tmp.10-flannel.conf
mv /cni/net.d/.tmp.10-flannel.conf  /cni/net.d/10-flannel.conf`
```

aaronlevy: Also, do we want to continually re-copy this forever (vs copy once, sleep forever)?

glide.yaml
chancez: Use glide 0.12.x, it's faster/better. Also, what command are you running exactly?

If on glide 0.12.x+ it should be roughly:

```
glide update --strip-vendor --skip-test
```

Also, can you separate out changing `glide.yaml` and `glide.lock` from the changes in vendor? It's impossible to see what changed in the manifest without checking out locally. (But I will so I can look).

Documentation/kube-flannel.yml
mikedanese: I had it copying once and @luxas asked me to put it in a while. https://github.com/coreos/flannel/pull/483#discussion_r77900055

I am unopionated

mikedanese: It should not change at runtime

glide.yaml
mikedanese: Thanks, I will give that a try.

Documentation/kube-flannel.yml
aaronlevy: I believe we want to have a `restartPolicy=true` on the pod so I think we don't want this container to exit (otherwise it will be in a restart loop). But it does seem a little odd to keep copying repeatedly (or incorrect if it should never change at runtime).

```
cp cni-conf.json tmp-10-flannel.conf
mv tmp-10-flannel.conf 10-flannel.conf
while true; do
    sleep 3600
done
```

glide.yaml
chancez: Im getting this error which is new to me:
```
[ERROR]	Update failed for google.golang.org/cloud: The Remote does not match the VCS endpoint
```
Looking at https://github.com/GoogleCloudPlatform/gcloud-golang makes me thing the dependency should be 
`cloud.google.com/go` instead in glide.yaml rather than `google.golang.org/cloud`

chancez: Changing that import let me run glide. So I'll try to update imports and see if it compiles.

chancez: Okay, so I had to do the following, since it seems like the repo for the google cloud libraries may have moved and, not only did they move, but the import URL changed. This will keep the existing import URL. (Glide can't figure it out due to the import URL changing i think)
```
diff --git a/glide.yaml b/glide.yaml
index de8f578..2f93117 100644
--- a/glide.yaml
+++ b/glide.yaml
@@ -30,7 +30,7 @@ import:
   - activation
   - daemon
 - package: github.com/coreos/pkg
-  version: v2 
+  version: v2
   subpackages:
   - flagutil
 - package: github.com/golang/glog
@@ -68,7 +68,9 @@ import:
 - package: google.golang.org/appengine
   version: 6a436539be38c296a8075a871cc536686b458371
 - package: google.golang.org/cloud
-  version: 872c736f496c2ba12786bedbb8325576bbdb33cf
+  repo: https://github.com/GoogleCloudPlatform/gcloud-golang.git
+  vcs: git
+  version: eb47ba841d53d93506cfbfbc03927daf9cc48f88
   subpackages:
   - compute/metadata
   - internal
```

chancez:
So I was able to get glide working with the following changes applied to glide.yaml:

```
diff --git a/glide.yaml b/glide.yaml
index de8f578..2f93117 100644
--- a/glide.yaml
+++ b/glide.yaml
@@ -30,7 +30,7 @@ import:
   - activation
   - daemon
 - package: github.com/coreos/pkg
-  version: v2 
+  version: v2
   subpackages:
   - flagutil
 - package: github.com/golang/glog
@@ -68,7 +68,9 @@ import:
 - package: google.golang.org/appengine
   version: 6a436539be38c296a8075a871cc536686b458371
 - package: google.golang.org/cloud
-  version: 872c736f496c2ba12786bedbb8325576bbdb33cf
+  repo: https://github.com/GoogleCloudPlatform/gcloud-golang.git
+  vcs: git
+  version: eb47ba841d53d93506cfbfbc03927daf9cc48f88
   subpackages:
   - compute/metadata
   - internal

```

As mentioned in my comment above, the google cloud API import URL changed, and maybe the underlying repo changed too, but regardless, the import url changed from what everything expects, so we need to specify where to clone from so we can keep the import URL as `google.golang.org/cloud`.

Once I fixed that, `glide up -v --skip-test` took ~40 seconds (unsure of how much I had cached) using glide 0.12.1. Additionally, you might consider using https://github.com/sgotti/glide-vc/ to remove unused files/packages/dependencies from vendor.

Documentation/kube-flannel.yml
aaronlevy: I wasn't around for the discussion, but we have been using `10.2.0.0/16` for default pod network for a while in the coreos-kubernetes docs/tools. I don't feel strongly about this though - as it's easily changed by the user.

Documentation/kube-flannel.yml
mikedanese: ok reverted. busybox doesn't support `sleep infinity` and shell should have a pause builtin.

chancez:
Unsubscribing. LMK if you have issues w/ glide.

philips:
I know this sucks but any chance we could move this PR to the new client to slim dependencies? https://github.com/kubernetes/client-go

mikedanese:
@philips I don't think we can until pkg/controller/framework is moved to a versioned client.

mikedanese:
This should be able to use some of client-go once https://github.com/kubernetes/kubernetes/pull/32718/ merges to reduce the amount of vendord.

tomdee:
@mikedanese What do I need to do to help you get this merged (other than hit the merge button :smile:). Some thoughts:
* Remove the vendor directory changes - but make sure that the glide.yaml and glide.lock are correct. This will make it easier for people to review the code.
* Documentation - I don't think there is any at the moment - are you able to add a doc explaining briefly the motivation behind this, roughly how it works and how it can be configured/deployed?
* And finally, I'm happy for their to be an example of deploying this under kubernetes, but that example can't be the "master", i.e. it's not a deliverable of the flannel project and I don't want other projects relying on it. I need to be able to change/update the "example" without worrying about it breaking other projects.

kkirsche:
@tomdee wouldn't the fact that you need to be able to change and update an example be exactly why it should be in master? Examples should always represent the current state of the codebase not what used to be, which is what often happens when people end up storing their examples separate from repos they are for.

tomdee:
The example can be checked into the master branch, I just don't want other projects automatically pulling that example yaml file and using it in their deploys. There was some discussion about what should be in flannel and what shouldn't be. e.g. An _example_ script for deploying the flannel CNI plugin would be OK, but it shouldn't be the _actual_ script that any deployment relies on.

tomdee:
Merging - we'll add more docs in a followup PR.

bh016088:
Any word on documentation?  I would like to use this and work through the kube api but even the help description of "Contact the Kubernetes API for subnet assignement instead of etcd or flannel-server." doesn't tell me what to actually put as the argument for -kube-subnet-mgr.

tomdee:
Docs are being tracked in this issue - https://github.com/coreos/flannel/issues/587

A PR adding _any_ docs would be very welcome.

timchenxiaoyu:
why we need  kubernetes apiserver  backed?

tomdee:
So that flannel doesn't need to talk to etcd. It makes configuration easier

timchenxiaoyu:
ths

vdavidoff:
If I am parsing this correctly, this only works if flannel is run inside kubernetes - is that right? Or will there be a mode where you can point it at a kube-apiserver rather than it discovering incluster config things?

PR #489, opened 2016-08-16
Makefile: Make the ARCH part of the tag name not the image name and add pthread to busybox

tomdee:


tomdee:
Fixes #494

Makefile
```diff
@@ -101,6 +105,10 @@ dist/flanneld-$(ARCH):
 		mv dist/flanneld dist/flanneld-$(ARCH) && \
 		file dist/flanneld-$(ARCH)'
 
+## Busybox images are missing pthread. Pull it out of the kube-cross image
+dist/libpthread.so.0-$(ARCH):
+	docker run -ti --rm -v `pwd`:/host gcr.io/google_containers/kube-cross:$(KUBE_CROSS_TAG) cp /lib/$(LIB_DIR)/libpthread.so.0 /host/dist/libpthread.so.0-$(ARCH)
```
eyakubovich: Is there a way to not pull out a lib out of some other container? Can we work to get this included into busybox-glibc image?

tomdee: This is the container that flanneld is built in, so seems like a good choice to me.

luxas:
We've debated about the multiarch naming for kubernetes as well, and came to the conclusion that `binary-arch:version` is what we're using. See: https://github.com/kubernetes/kubernetes/pull/26863 and https://github.com/kubernetes/kubernetes/pull/23059#issuecomment-201924090

Of course you may change to `binary:version-arch`, but I strongly prefer the current naming.

Another thing: earlier you've released flannel without the "v" prefix (current release is `flannel:0.5.5`).
With this new build script, it seems like the v has appeared in the tag; which is a breaking change.
I'm not opposed to it; v is good, but you should be consistent with other coreos images.

Adding pthread LGTM

luxas:
You should get this merged and release v0.6.1

PR #490, opened 2016-08-16
Add functional (end-to-end) testing

eyakubovich:
This script is the first cut at getting end-to-end testing
for flannel. It brings up etcd in a container and then 2 flannel
instances (also in containers). It then pings one flannel instance
from another. It does this for each of the three backends being
tested: udp, vxlan, host-gw.

To run the script, use "make e2e-test"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test cover gofmt gofmt-fix license-check clean tar.gz docker-push release docker-push-all
+.PHONY: test e2e-test cover gofmt gofmt-fix license-check clean tar.gz docker-push release docker-push-all
 
 # Registry used for publishing images
 REGISTRY?=quay.io/coreos
@@ -41,6 +41,9 @@ dist/flanneld: $(shell find . -type f  -name '*.go')
 test: license-check gofmt
 	go test -cover $(TEST_PACKAGES_EXPANDED)
 	cd dist; ./mk-docker-opts_tests.sh
+
+e2e-test: dist/flanneld-$(TAG)-$(ARCH).docker
+	cd dist; ./functional-test.sh $(REGISTRY)/flannel:$(TAG)-$(ARCH)
 
 cover:
 	# A single package must be given - e.g. 'PACKAGES=pkg/ip make cover'
--- a/dist/functional-test.sh
+++ b/dist/functional-test.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+
+
+ETCD_IMG="quay.io/coreos/etcd:v3.0.3"
+FLANNEL_NET="10.10.0.0/16"
+
+usage() {
+	echo "$0 FLANNEL-DOCKER-IMAGE"
+	echo
+	echo "Run end-to-end tests by bringing up two flannel instances"
+	echo "and having them ping each other"
+	echo
+	echo "NOTE: this script depends on Docker 1.9.0 or higher"
+	exit 1
+}
+
+version_check() {
+	required=$1
+	actual=$2
+	err_prefix=$3
+
+	req_maj=$(echo $required | cut -d . -f 1)
+	req_min=$(echo $required | cut -d . -f 2)
+	act_maj=$(echo $actual | cut -d . -f 1)
+	act_min=$(echo $actual | cut -d . -f 2)
+
+	if [ $act_maj -lt $req_maj ] || ( [ $act_maj -eq $req_maj ] && [ $act_min -lt $req_min ] ); then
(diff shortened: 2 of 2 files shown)


dist/functional-test.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
+
+
+ETCD_IMG="quay.io/coreos/etcd:v3.0.3"
```
tomdee: Is it better to us a fixed version here or to just take :latest? Tracking latest might give us an early heads-up if there's a problem using flannel with the latest etcd release?

dist/functional-test.sh
```diff
@@ -0,0 +1,155 @@
...
+	ver=$(docker version -f '{{.Server.Version}}')
+	version_check "1.9" $ver
+}
+
+run_test() {
+	backend=$1
+
+	while true; do
+		# etcd might take a bit to come up
+		docker run -it --name=flannel-test-etcdctl --entrypoint=/usr/local/bin/etcdctl $ETCD_IMG \
```
tomdee: Can't you just run this with --rm?

tomdee: then you can just do 
```
while ! docker run...;do
   sleep 1
done
```

dist/functional-test.sh
```diff
@@ -0,0 +1,155 @@
...
+		exit 1
+	fi
+
+	flannel2=$(docker run -d --privileged --entrypoint=/opt/bin/flanneld $flannel_img --etcd-endpoints=$etcd_endpt)
+	if [ $? -ne 0 ]; then
+		exit 1
+	fi
+
+	echo flannels running
+	sleep 5
```
tomdee: A little magical. Why 5? Can you add a comment explaining where 5 comes from (even if it's just arbitrary)

dist/functional-test.sh
```diff
@@ -0,0 +1,155 @@
...
+
+	# add a dummy interface with $FLANNEL_SUBNET so we have a known working IP to ping
+	ping_dest=$(docker "exec" --privileged $flannel1 /bin/sh -c '\
+		source /run/flannel/subnet.env && 
+		ip link add name dummy0 type dummy && \
+		ip addr add $FLANNEL_SUBNET dev dummy0 && \
+	       	ip link set dummy0 up && \
+		echo $FLANNEL_SUBNET | cut -f 1 -d "/" ')
+
+	docker "exec" -it --privileged $flannel2 /bin/ping -c 5 $ping_dest
```
tomdee: What's with the quoting?

dist/functional-test.sh
```diff
@@ -0,0 +1,155 @@
...
+		fi
+
+		sleep 5
+	done
+
+	echo flannel config written
+
+	flannel1=$(docker run -d --privileged --entrypoint=/opt/bin/flanneld $flannel_img --etcd-endpoints=$etcd_endpt)
+	if [ $? -ne 0 ]; then
+		exit 1
```
tomdee: If flannel doesn't come up, then exit is called which will leave a dangling etcd container. What about naming the etcd container when you create it, and always trying to delete it (ignore the rc) before trying to create it.

Makefile
```diff
@@ -38,6 +38,9 @@ test: license-check gofmt
 	go test -cover $(TEST_PACKAGES_EXPANDED)
 	cd dist; ./mk-docker-opts_tests.sh
 
+e2e-test: dist/flanneld-$(ARCH)-$(TAG).docker
+	cd dist; sudo -E ./functional-test.sh $(REGISTRY)/flannel-$(ARCH):$(TAG)
```
tomdee: I've made some changes to the naming of images - it's always flannel:$(TAG)-$(ARCH) now

tomdee:
Looks pretty nice to me. There are a few questions/comments to address. Also - have you tried running this on a CI system? Do you know if it works on travis?

dist/functional-test.sh
eyakubovich: Yes but on the other hand, you can start getting spurious failures since etcd is changing underneath you.

dist/functional-test.sh
eyakubovich: Great point!

dist/functional-test.sh
eyakubovich: I was thinking that `exec` is a bash keyword but it's just a builtin. I'll drop it.

tomdee:
Still no answer on the "does it work on travis" question, but merging anyway as I don't want it to languish any further.

PR #504, opened 2016-09-11
Updated deprecated google/golang docker container

qhenkart:
google/golang on dockerhub has been deprecated: https://hub.docker.com/r/google/golang/

The new official GO address is here: https://hub.docker.com/_/golang/

The flannel build fails unless it's run with modern versions of GO which is only supported in the second link


tomdee:
This container is no longer used so closing.

Issue #513, opened 2016-09-22
flannel v0.6.1 logging error


mischief:
from https://github.com/coreos/flannel/issues/491:

```
Sep 14 08:54:12 ip-10-10-4-76.ec2.internal rkt[2704]: log: exiting because of error: log: cannot create log: open /var/tmp/flanneld.ip-10-10-4-76.root.log.INFO.20160914-085412.2704: no such file or directory
```

the issue here is that backend.Register is called before main() sets `logtostderr`, and the new base images (busybox) lack /var/tmp which is where glog will try to write files by default.

the solution seems to be to simply not log during init.

mischief:
fixed by #514

PR #514, opened 2016-09-22
backend: do not log in Register

mischief:
various backends call backend.Register, which will log a message.
because Register is called in init() functions before main() runs,
glog's 'logtostderr' flag is not set yet, so glog attemps to create log
files in $TMPDIR. this directory doesn't exist in the new busybox base
image, and we don't want to write log files anyway.

### DIFF
--- a/backend/manager.go
+++ b/backend/manager.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


mischief:
fixes #513

tomdee:
LGTM

PR #518, opened 2016-09-25
Fix the kube-flannel.yaml in order to work

luxas:
@tomdee This makes it possible to run this flannel daemonset on other architectures as well, thanks!
Just `curl -sSL <url> | sed "s/amd64/arm/" | kubectl apply -f -` and you're done!

Can you merge this please so we can use it with kubeadm please?

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -40,9 +40,11 @@ spec:
         app: flannel
     spec:
       hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/arch: amd64
       containers:
       - name: kube-flannel
-        image: quay.io/coreos/flannel-git:latest
+        image: quay.io/coreos/flannel-git:v0.6.1-28-g5dde68d-amd64
         command: [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr" ]
         securityContext:
           privileged: true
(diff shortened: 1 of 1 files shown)


tomdee:
@luxas I was trying to avoid the situation where external projects rely on this "sample" config file. I don't believe that the flannel project is the right place for the canonical version of this file. See my comments https://github.com/coreos/flannel/pull/483#issuecomment-247160212 and https://github.com/coreos/flannel/pull/483#issuecomment-247389118

Documentation/kube-flannel.yml
```diff
@@ -18,7 +18,7 @@ data:
     }
   net-conf.json: |
     {
-      "Network": "10.244.0.0/16",
+      "Network": "10.2.0.0/16",
```
tomdee: What's this change for?

Documentation/kube-flannel.yml
```diff
@@ -61,8 +63,8 @@ spec:
         - name: flannel-cfg
           mountPath: /etc/kube-flannel/
       - name: install-cni
-        image: busybox
-        command: [ "/bin/sh", "-c", "set -e -x; TMP=/etc/cni/net.d/.tmp-flannel-cfg; cp /etc/kube-flannel/cni-conf.json ${TMP}; mv ${TMP} /etc/cni/net.d/10-flannel.conf; while :; do sleep 3600; done" ]
+        image: quay.io/coreos/flannel-git:v0.6.1-28-g5dde68d-amd64
+        command: [ "/bin/sh", "-c", "set -e -x; cp -f /etc/kube-flannel/cni-conf.json /etc/cni/net.d/10-flannel.conf; while true; do sleep 3600; done" ]
```
tomdee: What's this change for?

luxas:
Okay, the only thing I want here is that this manifest will be multiarch, so I can easily can run kubectl apply on it.

Please merge this as soon as possible. I removed the strange parts from this PR now

If you want to create a separate project for kubeadm integration, feel free to.
If you do, it would be the same as weave has done, but a separate directory of the flannel project is also fine for me.

At least, please merge and I'll add this to http://kubernetes.io/docs/admin/addons/, it will be good for users.

luxas:
ping @tomdee

tomdee:
@luxas I still don't really understand this change. Do you need the version pinning stuff or just the extra label?

luxas:
a) We need somewhere to point users to regarding flannel and kubeadm
b) The manifest has to point to a specific `-ARCH` tag (`-amd64` by default) in order to be able to sed it to arm or arm64 easily

See https://github.com/kubernetes/kubernetes/pull/30356 for reference

Thanks, and please let this though!

PR #524, opened 2016-10-02
aws-vpc: remove "blackholes"

### LINKED ISSUE #509: Flannel should cleanup blackhole/unused routes
Not much background on this, but there are a lot of Black Hole routes from instances that were recycled/terminated, shouldn't they be cleaned up on a hourly/daily basis?
Screenshot: https://puu.sh/rgF6B/6679679412.png

Bests

monder:
Fixes #509

It removes the invalid routes from the routetable that belong to `flannel` network.

I don't think we need to cleanup this regularly (not without client-server architecture at least) since the it is created only on instance startup.

### DIFF
--- a/backend/awsvpc/awsvpc.go
+++ b/backend/awsvpc/awsvpc.go
@@ -17,6 +17,7 @@ package awsvpc
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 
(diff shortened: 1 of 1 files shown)


tomdee:
LGTM

PR #534, opened 2016-10-21
Allow setting masterUrl and Kubeconfig for kube-flannel mode.

TamalSaha:
We have a case where we need to set public_ip for flannel daemon. This can't be done using DaemonSet. We are planning to run it as static pods. This change will allow us to set masterUrl and Kubeconfig for flannel.

I ran the following commands to revendor:

``` bash
glide get github.com/go-ini/ini
glide install -v
```


cristifalcas:
I think you have to check your PR, because it says you are modifying more than 10000 files. It looks like you are deleting everything?

TamalSaha:
@cristifalcas, the PR is ok. Travis build is working. I need to add dependency on `k8s.io/kubernetes/pkg/client/unversioned/clientcmd`

main.go
```diff
@@ -76,6 +78,8 @@ func init() {
 	flag.StringVar(&opts.remoteCertfile, "remote-certfile", "", "SSL certification file used to secure client/server communication")
 	flag.StringVar(&opts.remoteCAFile, "remote-cafile", "", "SSL Certificate Authority file used to secure client/server communication")
 	flag.BoolVar(&opts.kubeSubnetMgr, "kube-subnet-mgr", false, "Contact the Kubernetes API for subnet assignement instead of etcd or flannel-server.")
+	flag.StringVar(&opts.kubeMaster, "master", "", "The address of the Kubernetes API server (overrides any value in kubeconfig)")
```
tomdee: I think this option name is too generic, including `kube` would help, e.g. kube-api-master

tomdee:
I think you're deleting too much in the vendor dir too

TamalSaha:
@tomdee , I have made the changes as suggested. PTAL.

TamalSaha:
@tomdee, can this pull request be merged?

TamalSaha:
@aaronlevy do you mind taking a look at this pr?

aaronlevy:
@TamalSaha can you give a little more info into why you need this? I'm not very clear on what "we need to set public_ip for flannel daemon" means.

If this is running "in-cluster", which is the expected design, then you should be able to use the internal service account (api-server location as well assets to securely contact api-server).

TamalSaha:
@aaronlevy, we are using Flannel to setup networking for our kube cluster in Linode. Linode assigns 2 IPs (one public and one private) to the same eth0 interface. We would like to route all cluster pod traffic via the private IPs. So, we run flannels using static pods and set the -publicIP=<instances_private_ip>. In that scenario, we need to pass a kubeconfig to flanneld so that it can connect to master.

aaronlevy:
Is there any reason you can't tell the api-server to advertise itself on the private ip (if that's the one you want to use?), then I would think it should work out of box with no changes:

on api-server:
`--advertise-address=<private-ip>`

in the flannel daemonset command:
`--iface=<private-ip>` or `--public-ip=<private-ip>`

aaronlevy:
Also, to address the concern about daemonsets, you can use the downward api in the pod template to extract the local node's podIP. An example of this is in bootkube:

https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L585
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L597-L600

TamalSaha:
I have already set --advertise-address=<private-ip-of-master>.

From the flannel docs, I read:
```
--public-ip="": IP accessible by other nodes for inter-host communication. Defaults to the IP of the interface being used for communication.
--iface="": interface to use (IP or name) for inter-host communication. Defaults to the interface for the default route on the machine.
```
My understanding was that this `--public-ip` is the Ip used by flannel to route traffic.

> Also, to address the concern about daemonsets, you can use the downward api in the pod template to extract the local node's podIP. An example of this is in bootkube:

I think I see your point. Since the Daemonset has to be `HostNetwork: true`, PodIP = Instance's IP. If I configure Kubelet to use the <instance-private-ip> as nodeIP, then PodIP of flannel daemon will be <instance-privte-ip>. I think this will work. Did I understand you correctly?

aaronlevy:
To be honest, without looking at the code I'm not entirely sure the full interaction between `--public-ip` and `--iface` -- maybe @tomdee has a bit more info.

I guess if your interface has two IPs you could set both flags if the default route on that interface may not be the IP you want (e.g. --iface is defaulting to the wrong IP)? Just a guess (networking is not my strongpoint... PRs to this repo notwithstanding...)

aaronlevy:
@TamalSaha also yes, your understanding of the nodeIP -> podIP in the daemonset is correct.

TamalSaha:
@aaronlevy Thanks a lot for your help! I will test with DaemonSet and reopen if the problem exists.

tomdee:
I changed the logging so hopefully the behaviour around the --iface and --public-ip options should be a little clearer from the logging output at least.

PR #540, opened 2016-11-02
AWS VPC backend fixes and cleanups

### LINKED ISSUE #447: Disabling SourceDestCheck fails on instances with more than one ENI
Flannel fails to disable `SourceDestCheck` on instances with more than one network interface. SourceDestCheck attribute is actually an ENI attribute, but weirdly AWS exposes it as an instance attribute if it only has one network interface.

https://github.com/coreos/flannel/blob/6c95ca9086cf7f5fc6e1d415fb23c50aeec34c04/backend/awsvpc/awsvpc.go#L182

I could do a PR to fix that, if people agree? It'd be something similar to this: https://github.com/UKHomeOffice/smilodon/blob/master/aws.go#L260

Another way this could be fixed is, first we check which network interface flannel is going to use and only disable SourceDestCheck on that one.

vaijab:
- Update aws-sdk-go to latest stable, required for `aws/session` package
- Fix AWS API rate limits, flannel would go into a loop and will never recover because AWS API rate limits weren't handled properly
- Now flannel works on instances with more than one ENI, see #447 
- Improved logging and other clean ups

Fixes #447


vaijab:
CI job needs to be restarted because www.netfilter.org was down.

vaijab:
Any chance of reviewing this? /cc @tomdee 

We're using a custom built flannel version which includes this patch without any issues so far.

tomdee:
Thanks @vaijab looks good. Merging.

PR #547, opened 2016-11-16
Modifying pkg/ip/ipnet.go to fix endian problem

salamaniibm:
During test case execution, we came across all test case failues due to invalid IP.s generation for s390x system. On checking git issues , we came across the link  https://github.com/coreos/flannel/issues/526, which logs similar issue for bigEndiane during IP.s generation.
      On analysis we came across littleEndian check, where code is manipulating IP.s for bigEndianess, which is unnecessary because changes are required only at bit-level not for IP. We are removing the littleEndian check where the IP address are involved.
      For example subnetSize := ip.IP4(1 << (32 - cfg.SubnetLen)), assume cfg.SubnetLen=24 (by default) then, value of subnetSize should be same regardless of endianess, which is achieved after endiness check is removed.

### DIFF
--- a/pkg/ip/ipnet.go
+++ b/pkg/ip/ipnet.go
@@ -51,7 +51,7 @@ func MustParseIP4(s string) IP4 {
 }
 
 func (ip IP4) Octets() (a, b, c, d byte) {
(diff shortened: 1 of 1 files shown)


tomdee:
LGTM

Issue #427, opened 2016-04-07
L3 miss and Route not found loop


cpg1111:
Hi there, 
I'm running Flannel v0.5.5 in vxlan mode as part of a Kubernetes cluster.  I seem to keep getting this repeating:

```
<Timestamp> <host> flanneld[873]: I0407 18:36:51.705743 00873 vxlan.go:345] L3 miss: <Service's IP>
<Timestamp> <host> flanneld[873]: I0407 18:36:51.705865 00873 vxlan.go:349] Route for <Service's IP> not found
```

It only happens to one node in the cluster at a time.  It would seem like it's not aware a route has changed/removed and it is still looking for it. 

It all was working fine for awhile, then I updated one service and then I keep getting this.

iT2afL0rd:
This happens to me also.

cpg1111:
It seems Flannel changes its subnet right before this happens, making each service's IPs invalid, I assume the same applies to the pods' IPs.  Anyone have insight as to what could cause this?

hermanjunge:
Got the same problem.

Rastusik:
seems like I run into this as well

rvadim:
Got the same problem. Cross node network not work due to docker load **flannel_docker_opts.env**, but in my case flunnel starts after docker, so restart docker fix this problem.
```
$ systemctl cat docker
# /usr/lib64/systemd/system/docker.service
[Unit]
Description=Docker Application Container Engine
Documentation=http://docs.docker.com
After=docker.socket early-docker.target network.target
Requires=docker.socket early-docker.target

[Service]
EnvironmentFile=-/run/flannel_docker_opts.env
ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ

[Install]
WantedBy=multi-user.target
```
Add below config to avoid problem in future (cloud-config or something)
```
# /etc/systemd/system/docker.service.d/40-flannel.conf
[Unit]
Requires=flanneld.service
After=flanneld.service
```

tomdee:
I'm a little surprised to see service IPs in the flannel logs, shouldn't the kube-proxy be rewriting them?

cpg1111:
[...]

PR #555, opened 2016-11-24
vxlan logging verbosity

jsoriano:
As discussed in #427, vxlan logs can be too verbose. With current glog usage, it's not possible to reduce log level, it logs to stderr, and when logging to stderr, everything is logged. Everything is logged unless verbosity levels are used. This PR adds verbosity levels to vxlan Info logs, so they can be enabled when debugging some problem.


tomdee:
LGTM

jsoriano:
Thanks!

PR #559, opened 2016-11-30
Add a flag to configure the subnet lease renewal margin.

kryzthov:
This adds the command-line flag `--subnet-lease-renew-margin`, specified as
a number of minutes, instead of relying on the hard-coded default of 1h.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ This is the only mandatory key.
    The list of available backends and the keys that can be put into the this dictionary are listed below.
    Defaults to "udp" backend.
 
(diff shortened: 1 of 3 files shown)


tomdee:
Could you add some documentation for this? At least a small update to the README but the more the better.

kryzthov:
Hey Tom! Thanks for the feedback. I added a line for the new flag and a small section on the subnet lease duration and timing. Let me know if you think it should be expanded.

tomdee:
LGTM - merging

Issue #533, opened 2016-10-19
kube-flannel: host->remote-pod networking does not work on first node


aaronlevy:
I've been trying to use kube-flannel as part of the bootkube, but I'm still seeing an issue on bootstrap.

What I'm seeing is that on some nodes host-network --> remote pod (seemingly) hits some kind of race condition when bootstrapping using kube-flannel.

pod-to-pod (on same node and to remote nodes) works fine, this just affects host to remote pod, and it begins working after restarting kube-flannel on the remote node.
### Launch a cluster:

```
go get github.com/kubernetes-incubator/bootkube
git remote add aaronlevy https://github.com/aaronlevy/bootkube
cd $GOPATH/src/github.com/kubernetes-incubator/bootkube
git fetch aaronlevy
git checkout -b v1.4.0+flannel aaronlevy/v1.4.0+flannel
make clean all
cd hack/multi-node
./bootkube-up
```
### Start nginx pods for testing:

```
kubectl --kubeconfig=cluster/auth/kubeconfig run n1 --image=nginx --replicas=2
```

You should will have one nginx pod on each node, but can verify by checking nodeIP of each pod:

```
kubectl --kubeconfig=cluster/auth/kubeconfig get pods -owide
```

Make note of the podIP for each of the pods above, and which node they are assigned to. E.g.:

```
NAMESPACE     NAME                                       READY     STATUS    RESTARTS   AGE       IP             NODE
default       n1-1110676790-e0w0c                        1/1       Running   0          12m       10.2.1.2       172.17.4.201
default       n1-1110676790-s3qfi                        1/1       Running   0          12m       10.2.0.5       172.17.4.101
```
- The 10.2.0.5 is assigned to the "controller" (c1) node (172.17.4.101)
- The 10.2.1.2 is assinged to the "worker" (w1) node (172.17.4.201)
### Test the routability from each host:
- note: This seems to fail in different ways sometimes (e.g. both sides, or just one side). Typically from what I've seen the common case is master (host) --> worker (pod) fails.

From controller node (route to remote pod likely doesn't work):

```
vagrant ssh c1
$ curl 10.2.0.5 # same node, should work
$ curl 10.2.1.2 # different node, probably doesn't work
```

From worker node (all routes should work):

```
vagrant ssh w1
$ curl 10.2.0.5 # different node, probably works
$ curl 10.2.1.2 # same node, should work
```
### Now to resolve the issue:

[...]

PR #560, opened 2016-12-02
Improvements to kubernetes lease management & vxlan backend

aaronlevy:
fixes https://github.com/coreos/flannel/issues/533
fixes https://github.com/coreos/flannel/issues/535

The changes to `backend/vxlan/device.go` allow the use of the network address as a subnet assigned to a node. (e.g `network=10.1.0.0/16`, and a node is assigned `10.1.0.0/24`)

This is necessary because the kubernetes subnet lease assignment allows the use of the first subnet (whereas flannel previously did not). The fix itself is to remove all existing routes for the interface, and explicitly re-add a route.

In `subnet/kube/kube.go` the change is to emit lease events, rather than continually resetting all leases via a snapshot. We were previously ignoring updates to vxlan leases, because when evaluating the snapshots - we did not check if the mac address on the lease had changed (instead, we were only checking the subnet).

/cc @philips @bison @tomdee @mikedanese

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -128,6 +128,34 @@ func (dev *vxlanDevice) Configure(ipn ip.IP4Net) error {
 		return fmt.Errorf("failed to set interface %s to UP state: %s", dev.link.Attrs().Name, err)
 	}
 
+	// Remove all existing routes for this interface
+	mainFilter := &netlink.Route{
+		LinkIndex: dev.link.Attrs().Index,
+		Table:     syscall.RT_TABLE_MAIN,
+	}
+
+	localFilter := &netlink.Route{
+		LinkIndex: dev.link.Attrs().Index,
+		Table:     syscall.RT_TABLE_LOCAL,
+	}
+
+	mainRoutes, err := netlink.RouteListFiltered(netlink.FAMILY_ALL, mainFilter, netlink.RT_FILTER_OIF|netlink.RT_FILTER_TABLE)
+	if err != nil {
+		return fmt.Errorf("Failed to list routes: %v", err)
+	}
+
+	localRoutes, err := netlink.RouteListFiltered(netlink.FAMILY_ALL, localFilter, netlink.RT_FILTER_OIF|netlink.RT_FILTER_TABLE)
+	if err != nil {
+		return fmt.Errorf("Failed to list routes: %v", err)
+	}
+
+	for _, er := range append(mainRoutes, localRoutes...) {
+		log.Infof("Removing route: %s", er.String())
+		if err := netlink.RouteDel(&er); err != nil {
+			return fmt.Errorf("Failed to delete route: %v", err)
+		}
+	}
+
 	// explicitly add a route since there might be a route for a subnet already
 	// installed by Docker and then it won't get auto added
 	route := netlink.Route{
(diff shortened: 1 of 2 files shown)


aaronlevy:
In my tests, after rebooting a node it can take a few seconds for a network request to go through the first time, as the remote destination will still have stale arp table entries. After a few seconds it will see the L3 miss and update the table. As an optimization, we might be able to flush the cache when we see that the lease has changed (but I'd leave that to a follow up).

lxpollitt:
cc @tomdee

subnet/kube/kube.go
```diff
@@ -120,11 +124,59 @@ func newKubeSubnetManager(c clientset.Interface, sc *subnet.Config, nodeName str
...
+	}
+}
+
+func (ksm *kubeSubnetManager) handleUpdateLeaseEvent(oldObj, newObj interface{}) {
+	o := oldObj.(*api.Node)
+	n := newObj.(*api.Node)
+	if s, ok := n.Annotations[subnetKubeManagedAnnotation]; !ok || s != "true" {
+		return
+	}
+	if o.Annotations[backendDataAnnotation] == n.Annotations[backendDataAnnotation] &&
```
philips: nit: newline above would be nice.

tomdee:
@aaronlevy Awesome work on getting this fixed, looks like it was much more complicated than we initially thought (and thanks @bison for the assist) I'm still getting my head around the issue and your fix but here are some initial thoughts:
1) Have you looked into the failing tests?
2) How confident are you that it's safe to use the first subnet (given the slightly unhelpful comment https://github.com/coreos/flannel/blob/master/subnet/config.go#L73-L80)
3) Given there was some concern about this also being a problem for the UDP backend, do you think it's worth investigating/fixing that too?

aaronlevy:
1.) The failing test looks unrelated (invalid `docker logs` parameter). I'll try to verify locally.

2.) It seems from my brief internet reading, it is valid to use (but for some legacy reasons some do not). However, for any cases where kubernetes is assigning the pod cidrs - this is going to be used (so for anywhere that an overlay tool is not being used: aws/gce/etc.)

Additionally, the change is just for an external lease assignment - the existing flannel functionality doesn't change. What does change, is blowing away all previous routes for the vxlan interface. As long as this is acceptable -- that would be the only change to existing functionality.

Maybe someone with more network-chops sign off on the use of the network address / safety of removing all existing routes when creating interface?

3.) There were actually two issues at play here, and I believe only one of them affected the UDP backend (using the network address) and this is also resolved (but, I haven't explicitly tested this yet). The other issue had to do with only determining a lease change if there was a change to the subnet -- which is true for UDP -- but for VXLAN the mac address of the vxlan device could also change.

The latter issue existed for vxlan anyway, but the window for it occurring outside the kubernetes implementation was very small: you had to have a failed watch from etcd, which triggers a reset of all leases and in that window a mac address of one of the leases would need to change. I can add a TODO about this if we want: https://github.com/coreos/flannel/blob/808b3061e1dc69fe08af6b36e75e3c155b9e91d7/subnet/watch.go#L79 -- really this check should be backend specific (e.g. what constitutes a lease change for a particular backend).

backend/vxlan/device.go
```diff
@@ -128,14 +128,42 @@ func (dev *vxlanDevice) Configure(ipn ip.IP4Net) error {
...
+	}
+
+	for _, er := range append(mainRoutes, localRoutes...) {
+		log.Infof("Removing route: %s", er.String())
+		if err := netlink.RouteDel(&er); err != nil {
+			return fmt.Errorf("Failed to delete route: %v", err)
+		}
+	}
+
 	// explicitly add a route since there might be a route for a subnet already
```
tomdee: This comment isn't right any more

tomdee:
As discussed offline, do you want to split the vxlan bits out of this and make this PR just about the Kubernetes fixes. Then we can continue discussion of the vxlan fix back over on #533

aaronlevy:
Going to close this PR in favor of https://github.com/coreos/flannel/pull/564 which just pulls out the subnet/kube changes. Discussion of the remaining issue should probably stay in https://github.com/coreos/flannel/pull/564 -- let me know if this should be re-opened instead.

PR #561, opened 2016-12-02
backend/aws-vpc: add support for multiple route tables

pingles:
We're setting up a cluster in AWS that spans across availability zones. Because we're using private + public networks we have 3 route tables (1 for each AZ). This change adds support to Flannel for configuring more than one route table.

The implementation is pretty naive: all route tables will be configured with the same routes (and with blackholes removed etc.). 

I'm not an expert in AWS networking so any comments would be appreciated. Likewise, if there's a good way to write some tests to prove the behaviour, or ways to improve the code then please let me know!

Thanks so much for the project!

### DIFF
--- a/backend/awsvpc/awsvpc.go
+++ b/backend/awsvpc/awsvpc.go
@@ -53,13 +53,42 @@ func (be *AwsVpcBackend) Run(ctx context.Context) {
 	<-ctx.Done()
 }
 
+type backendConfig struct {
+	RouteTableID interface{} `json:"RouteTableID"`
+}
+
+func (conf *backendConfig) routeTables() ([]string, error) {
+	if table, ok := conf.RouteTableID.(string); ok {
+		log.Info(fmt.Sprintf("RouteTableID configured as string: %s", table))
+		return []string{table}, nil
+	}
+	if rawTables, ok := conf.RouteTableID.([]interface{}); ok {
+		log.Info(fmt.Sprintf("RouteTableID configured as slice: %+v", rawTables))
+		tables := make([]string, len(rawTables))
+		for idx, t := range rawTables {
+			table, ok := t.(string)
+			if !ok {
+				return nil, fmt.Errorf("Unexpected type in RouteTableID slice. Must be strings.")
+			}
+			tables[idx] = table
+		}
+		return tables, nil
+	}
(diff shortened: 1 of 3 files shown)


vaijab:
Does that mean that flannel adds private subnet routes to public network routing tables?

pingles:
@vaijab: nope. we have our VPC networking configured with a private + public subnet per-AZ, for example: `eu-west-1a-public`, `eu-west-1a-private` etc. The public subnets have routes to the internet gateway, the private networks have a `0.0.0.0` route pointing to nat gateways in the equivalent public subnet.

So, for Flannel we configure it to manage routes in the private route tables.

Hope that helps? I'm not really an AWS networking guru so I may have messed up some terminology.

t0mmyt:
To expand, if you want to have non internet facing subnets (in an AWS VPC) that have internet access, each subnet requires its own NAT gateway for HA.  This means that each subnet must have its own route table (as each has a different default route) but each route table still needs the full set of routes set by flannel.  Therefore the fix is to support propagating to multiple route tables.

vaijab:
@t0mmyt @pingles thanks. I think we wanted to have a similar AWS networking setup for our clusters, but flannel was the show stopper back then.

lokesh-shekar:
This PR merge will help us greatly

tomdee:
@pingles Are you able to rebase this PR and fix the conflict?

pingles:
Yep, let me take a look now.

pingles:
So it looks a little more involved than a few straight fixes- it's pretty late here now so I'll take a look at work tomorrow and push an update.

tomdee:
Great, thanks @pingles

pingles:
Really sorry but I've not had much time to do this today. I'm also away for the next week on holiday so I _probably_ won't look at it until I get back. If someone else wants to give it a try please do- otherwise I'll do it asap.

t0mmyt:
Rebase @ #717

pingles:
Thanks for doing that @t0mmyt! @tomdee any suggestions on how we could improve to be merge-worthy? :)

gunjan5:
@pingles @t0mmyt I've added review comments over at #717

gunjan5:
#717 is merged, so closing this one. Thanks @pingles and @t0mmyt!

PR #564, opened 2016-12-05
Improve kube subnet lease handling

aaronlevy:
- Emit add/remove events to the subnet handler rather than using periodic snapshots.
    - This resolves an issue where snapshots would not evaluate changes to the MAC address in the lease. This remains an issue, but we do not need to use snapshots initially, and this PR changes to using events. Filed issue: https://github.com/coreos/flannel/issues/563 to track - as we might want to use snapshots eventually if we want to reset the state periodicially.
- Wait for informer cache to sync so we're not acting on partial/empty state
- We were updating a cached node object in-place, then calling an update via API. If the update failed, it would have left the cached object mismatched from actual state in API. Instead, operate on a copy of the cached object to remove this concern.

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -45,7 +45,8 @@ var (
 )
 
 const (
-	resyncPeriod = 5 * time.Minute
+	resyncPeriod              = 5 * time.Minute
+	nodeControllerSyncTimeout = 10 * time.Minute
 
 	subnetKubeManagedAnnotation = "flannel.alpha.coreos.com/kube-subnet-manager"
 	backendDataAnnotation       = "flannel.alpha.coreos.com/backend-data"
@@ -105,12 +106,14 @@ func NewSubnetManager() (subnet.Manager, error) {
 	}
 	go sm.Run(context.Background())
 
-	err = wait.Poll(time.Second, 10*time.Second, func() (bool, error) {
+	glog.Infof("Waiting %s for node controller to sync", nodeControllerSyncTimeout)
+	err = wait.Poll(time.Second, nodeControllerSyncTimeout, func() (bool, error) {
 		return sm.nodeController.HasSynced(), nil
 	})
 	if err != nil {
 		return nil, fmt.Errorf("error waiting for nodeController to sync state: %v", err)
 	}
+	glog.Infof("Node controller sync successful")
 
 	return sm, nil
 }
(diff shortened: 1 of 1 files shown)


aaronlevy:
I will do manual testing of this a bit later today, just wanted to get it up for review.

/cc @tomdee 

Related discussion: https://github.com/coreos/flannel/issues/533
Replaces part of: https://github.com/coreos/flannel/pull/560

subnet/kube/kube.go
```diff
@@ -103,7 +104,15 @@ func NewSubnetManager() (subnet.Manager, error) {
 		return nil, fmt.Errorf("error creating network manager: %s", err)
 	}
 	go sm.Run(context.Background())
-	return sm, err
+
+	err = wait.Poll(time.Second, 10*time.Second, func() (bool, error) {
+		return sm.nodeController.HasSynced(), nil
```
philips: log output would be nice here.

subnet/kube/kube.go
```diff
@@ -103,7 +104,15 @@ func NewSubnetManager() (subnet.Manager, error) {
 		return nil, fmt.Errorf("error creating network manager: %s", err)
 	}
 	go sm.Run(context.Background())
-	return sm, err
+
+	err = wait.Poll(time.Second, 10*time.Second, func() (bool, error) {
```
philips: why 10 seconds?

philips:
Overall, LGTM

subnet/kube/kube.go
aaronlevy: Somewhat arbitrary. If this is not able to sync, the pod will fail and we'll see a restart count in `kubectl` output. Seemed that would at least make some debugging easier than waiting forever (e.g. in the case it cannot contact apiserver). 10 seconds seemed like a reasonable amount of time to wait on this.

subnet/kube/kube.go
aaronlevy: ack, will add.

aaronlevy:
Prior to this merging I still need to do a manual test. Then I also want to test the renew lease logic (which is 24hrs right now, but I'll lower for testing).

aaronlevy:
Manual tests went fine.

The only thing I'm noticing is that after the mac address changes (reboot of machine or restart of flannel), the first request will take ~10seconds because we initially treat it as "not a miss" (eventually I'm assuming the arp table MAC entry is cleared and we then repopulate with correct info from the lease). Likely some optimizations we can make here like flushing the arp table entry on lease change (but I can open a separate issue for this).

@philips I responded to your questions, but wondering if you would prefer changes (also this got "hidden", but still unaddressed: https://github.com/coreos/flannel/pull/564#discussion_r91019413

If all looks well, should be ready for merge.

philips:
LGTM, merge it.

PR #565, opened 2016-12-05
network manager: Improve logging

tomdee:
Improve startup logging around determining external address and
interface to use


philips:
LGTM

nit: would have been easier to review if the rename and logs were in separate commits.

tomdee:
OK, I'll know for next time 😄

PR #566, opened 2016-12-06
backend/vxlan: Remove all existing routes when configuring interface

philips:


philips:
@tomdee @aaronlevy not removing local routes seems to make the tests pass. Need to investigate more.

aaronlevy:
In my testing it did not, however, resolve the pod routing issue.

tomdee:
replaced by #569

PR #569, opened 2016-12-09
[do-not-merge] backend/vxlan: Remove local broadcast route for network

aaronlevy:
Fixes https://github.com/coreos/flannel/issues/533
Fixes https://github.com/coreos/flannel/issues/535

This change will remove the broadcast route added to the local route table when the vxlan interface is setup. By removing the route entry, the node will be able to properly route traffic meant for the host assigned the network address within the in flannel network (see issue/533 for more info).

While this may seem a bit heavy-handed - if flannel is restarted, the broadcast routes will not be re-created anyway.

/cc @tomdee @philips

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -28,6 +28,8 @@ import (
 	"github.com/coreos/flannel/pkg/ip"
 )
 
+const RTN_BROADCAST = 3
+
 type vxlanDeviceAttrs struct {
 	vni       uint32
 	name      string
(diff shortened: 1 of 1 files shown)


philips:
Posted to netdev asking about why the Kernel does this: http://marc.info/?l=linux-netdev&m=148132848222491&w=2

backend/vxlan/device.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/coreos/flannel/pkg/ip"
 )
 
+const RTN_BROADCAST = 3
```
steveeJ: the `syscall` package defines this constant

aaronlevy:
Issue this PR was trying to address was resolved by: https://github.com/coreos/flannel/pull/576

PR #573, opened 2016-12-13
e2e: Allow the backend list to be overridden

tomdee:
e.g.
BACKEND=vxlan make e2e-test

to run just the vxlan tests


tomdee:
Very simple test only change so merging

PR #576, opened 2016-12-13
backend/vxlan: Set the netmask of the IP used for the vxlan device

tomdee:
Use a /32 netmask for the vxlan device. This avoids the kernel programming
broadcast routes (which won't work).

Unicast routes are already explicitly programmed so there's no need to
use a broader netmask.

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -305,7 +305,10 @@ func setAddr4(link *netlink.Vxlan, ipn *net.IPNet) error {
 			return fmt.Errorf("failed to delete IPv4 addr %s from %s", addr.String(), link.Attrs().Name)
 		}
 	}
-
+	// Ensure that the device has a /32 address so that no broadcast routes are created.
+	// This IP is just used as a source address for host to workload traffic (so
+	// the return path for the traffic has a decent address to use as the destination)
+	ipn.Mask = net.CIDRMask(32, 32)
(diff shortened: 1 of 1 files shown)


backend/vxlan/device.go
```diff
@@ -305,7 +305,10 @@ func setAddr4(link *netlink.Vxlan, ipn *net.IPNet) error {
 			return fmt.Errorf("failed to delete IPv4 addr %s from %s", addr.String(), link.Attrs().Name)
 		}
 	}
-
+	// Ensure that the device has a /32 address so that no broadcast routes are created.
+	// This IP is just used as a source address for host to workload traffic (so
+	// the return path for the traffic has a decent address to use as the destination)
+	ipn.Mask = net.CIDRMask(32, 32)
```
aaronlevy: really minor nit: may want this in the caller if want to keep `setAddr4` generic. Or if this is the only use case (which I believe it is), may as well only require caller to provide an`IP` rather than `IPNet`.

aaronlevy:
@tomdee I did manual tests against the repo steps in https://github.com/coreos/flannel/issues/533 and all went well.

On 2 nodes tested:

host to pod in both directions
pod to pod in both directions

Rebooted each node and retried the above then rebooted the next node and tried again.

All worked!

tomdee:
Cool! Merging.

PR #579, opened 2016-12-19
Replacing the user id with group id.

### LINKED ISSUE #578: The flannel makefile should set correct group id
To build flanneld the Makefile uses docker containers running with a specific user id and group id. I believe the gid is being set incorrectly in a few places:
https://github.com/coreos/flannel/blob/master/Makefile#L102
https://github.com/coreos/flannel/blob/master/Makefile#L118

The docker run option takes a user id and group id:
`-u, --user=                Username or UID (format: <name|uid>[:<group|gid>])`

It should be `-u $(shell id -u):$(shell id -g)` in both cases.

mbruzek:
I found this bug with the Makefile's use of docker.

Fixes #578


tomdee:
Thanks, looks good.

PR #591, opened 2017-01-15
Add Kubernetes RBAC support

andrewrynhard:
This PR adds RBAC support.

@liggitt

### DIFF
--- a/Documentation/kube-flannel-rbac.yml
+++ b/Documentation/kube-flannel-rbac.yml
@@ -1,0 +1,26 @@
+# Create the clusterrole:
+# $ kubectl create -f kube-flannel-rbac.yml
+# Bind the flannel serviceaccount to the flannel clusterrole:
+# $ kubectl create clusterrolebinding flannel --clusterrole=flannel --serviceaccount=kube-system:flannel
+# Create the pod using the same namespace used by the flannel serviceaccount:
+# $ kubectl create --namespace kube-system -f kube-flannel.yml
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: flannel
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - get
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - list
+      - update
+      - watch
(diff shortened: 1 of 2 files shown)


Documentation/kube-flannel-rbac.yml
```diff
@@ -0,0 +1,123 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: flannel
+  namespace: kube-system
```
liggitt: rather than hardcoding the kube-system namespace and duplicating the config/pod spec from the main example, I would include the service account in the core kubernetes yml example (without the namespace), make the core example use a bespoke service account, limit this file to the clusterrole with the required permissions, and document the kubectl command to grant the clusterrole to the flannel service account in the desired namespace

andrewrynhard: Ahh yes. Will fix now.

andrewrynhard:
@liggitt updated. Is this what you had in mind?

Documentation/kube-flannel.yml
```diff
@@ -42,6 +69,7 @@ spec:
       hostNetwork: true
       nodeSelector:
         beta.kubernetes.io/arch: amd64
+      serviceAccount: flannel
```
liggitt: serviceAccountName is the non-deprecated field

liggitt:
Looks good to me

Documentation/kube-flannel-rbac.yml
```diff
@@ -0,0 +1,26 @@
+# Create the clusterrole:
+# $ kubectl create -f kube-flannel-rbac.yml
+# Bind the flannel serviceaccount to the flannel clusterrole:
+# $ kubectl create clusterrolebinding flannel --clusterrole=flannel --serviceaccount=kube-system:flannel
+# Create the pod using the same namespace used by the flannel serviceaccount:
+# $ kubectl create --namespace kube-system -f kube-flannel.yml
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1alpha1
```
liggitt: might want to switch this to v1beta1 for longevity

andrewrynhard: yep, fixing now

tomdee:
LGTM - merging

PR #595, opened 2017-01-21
[WIP] Initial prototyping of etcdv3 support

tomdee:
I'd be more than happy for someone to take this on and complete it


xiang90:
@tomdee Can split the vendoring and code part? I would like to give this a quick review :)

subnet/local_manager.go
```diff
@@ -68,7 +68,14 @@ func (c watchCursor) String() string {
 }
 
 func NewLocalManager(config *EtcdConfig) (Manager, error) {
-	r, err := newEtcdSubnetRegistry(config, nil)
+	var err error
+	var r Registry
+
+	if config.UseV3API {
```
tomdee: I haven't done the part to choose between v2 and v3 yet, so this branch always uses v3

tomdee:
@xiang90 Done. But please note that this really is far from complete - I've really done the bare minimum to get _something_ that works. Of course, any and all feedback is still appreciated!

philips:
I don't think we should have an etcd v3 backend. Most of the users of Flannel should be using the Kubernetes subnet manager.

styxlab:
@philips flannel reads its configuration from etcd. How does using the Kubernetes subnet manager change that?

PR #600, opened 2017-01-29
[doc][kube-flannel] set explicit arch on daemonset images

bamarni:
It would help for quick scripting on dev/test environments and other architectures.

For instance the [following instructions](https://github.com/luxas/kubernetes-on-arm#set-up-the-master) are currently broken since https://github.com/coreos/flannel/commit/ee52b531b0c3ce81b3a08bc3a5dccf4d31476ecd.


tomdee:
Thanks for spotting this. Merging.

PR #606, opened 2017-02-08
remote: Remove experimental remote support

tomdee:

PR #608, opened 2017-02-08
bugfix: json encode object with pointer

ZhuPeng:
The json encoded `Lease`, which BackendData (`json.RawMessage`) base64 encoded,  in `handleRenewLease` function was wrong.


philips:
We are going to remove the remote API in #606. So, I am going to close this issue.

PR #625, opened 2017-03-03
added production users and integrations pages

pgburt:
Production-users and Integrations pages, per the request from @philips


philips:
Instead of libraries and tools it should be integrations. So, for example Kubernetes should be listed.

pgburt:
Added some content and a link to the bottom of the README @philips

tomdee:
LGTM

PR #629, opened 2017-03-09
Backends: Remove Run() from interface as it's not used

tomdee:


backend/manager.go
```diff
@@ -73,7 +73,7 @@ func (bm *manager) GetBackend(backendType string) (Backend, error) {
 
 	bm.wg.Add(1)
 	go func() {
-		be.Run(bm.ctx)
+		<-bm.ctx.Done()
```
gunjan5: Thinking out loud here: shouldn't we be able to just remove `ctx` field from `manager` struct? Seems like we're not using it anywhere, just closing it here.

tomdee: I think it would be better to do this under a new PR, but thanks for spotting it.

PR #634, opened 2017-03-10
Adding flannel support for s390x 

salamani:
Hi
we are adding a dockerfile for s390x so that flanneld binary will be available for s390x platform.
In this PR we are upgrading to kube-cross:v1.7.4-1 because it supports s390x.
Busybox uses old version of libc.so.6 and ld64.so.1, so pulling it out of the kube-cross image for s390x.


Makefile
```diff
@@ -30,8 +30,17 @@ ifeq ($(ARCH),ppc64le)
...
 endif
+ifeq ($(ARCH),s390x)
+	LIB_DIR=s390x-linux-gnu
+	CC=s390x-linux-gnu-gcc
+endif
 GOARM=6
-KUBE_CROSS_TAG=v1.6.2-2
+ifeq ($(ARCH),s390x)
+	# kube-cross:v1.7.4-1 supports s390x
+	KUBE_CROSS_TAG=v1.7.4-1
```
tomdee: Can you think of a good reason not to just switch to this version of KUBE_CROSS in all cases?

salamaniibm: KUBE_CROSS:1.7.4-1 has go which supports all the architectures including s390x(The curent version of KUBE_CROSS has go which does not support s390x). Also this new image uses glibc version 2.23.
But the busybox uses the glibc version 2.19 which is not compatible with flannel binary generated.To handle this situation for s390x, I am pulling libc.so.6 and ld64.so.1 from kube-cross to busybox.
If we upgrade the KUBE_CROSS version then either we need to change glibc version used in busybox or like s390x we will have to pull libc.so.6 and ld64.so.1 from kube-cross and use in busybox for all the architectures.

salamaniibm: Did I answer your question or let me know if there is any comment

tomdee: You did :smile: I now need to make a decision about whether we want consistency across our platforms (at a cost of a bit of reduced stability) or whether to take this change as is but to have the extra complexity in the builds.

I don't think it needs to be decided right now, but should be before the next flannel release.

tomdee:
This might need a tweak now that #650 has gone in.

tomdee:
@Pensu If you could squash the commits down to one then I can merge this.

Pensu:
@tomdee Seems like this PR is owned by @salamani, I am not sure what am I supposed to do here?

tomdee:
@Pensu Sorry wrong person! @salamani Would you be able to squash the commits, then I can merge?

tomdee:
Closing since I've merged the commit from this PR to master already.

BUT - the build seems to be failing and I have no way to test it. @salamaniibm would you be able to take a look? I'm seeing the following in the travis output:
```
dist/iptables-s390x: ELF 64-bit MSB shared object, IBM S/390, version 1 (SYSV), dynamically linked, interpreter /lib/ld64.so.1, for GNU/Linux 3.2.0, BuildID[sha1]=8f1e00df428b34b16a92ccdb400e7520c40c5a69, not stripped
docker run --rm -v /home/travis/gopath/src/github.com/coreos/flannel:/host gcr.io/google_containers/kube-cross:v1.7.5-3 cp s390x-linux-gnu/libc-2.23.so /host/dist/libc.so.6-s390x
cp: cannot stat 's390x-linux-gnu/libc-2.23.so': No such file or directory
make[1]: *** [dist/libpthread.so.0-s390x] Error 1
make[1]: Leaving directory `/home/travis/gopath/src/github.com/coreos/flannel'
```

PR #662, opened 2017-04-05
Makefile: Use $(CURDIR)

glevand:
The current makefile variable 'PWD' is not defined in the makefile,
and leads to build errors.  Use $(shell pwd) instead.

Fixes errors like these:

  docker: Error response from daemon: Invalid volume spec ":/go/src/github.com/coreos/flannel:ro"


tomdee:
I think it would be better to change `${PWD}` to the make variable `$(CURDIR)`

glevand:
Rebased to latest, changed `$(shell pwd)` to `$(CURDIR`.

tomdee:
LGTM - merging

PR #677, opened 2017-04-13
update kube subnet manager to k8s.io/client-go

mikedanese:
@aaronlevy @philips


glide.lock
```diff
@@ -163,116 +187,146 @@ imports:
...
+  - pkg/util/sets
+  - pkg/util/validation
+  - pkg/util/validation/field
+  - pkg/util/wait
+  - pkg/util/yaml
+  - pkg/version
+  - pkg/watch
+  - third_party/forked/golang/reflect
+- name: k8s.io/client-go
+  version: 5b0e11b577b35539f05523c47e94ed96a17f992b
```
euank: Is there a reason to use master over a tagged version, e.g. 3.0?

mikedanese: I'm planning on using 3.0 but still trying to figure out glide. https://github.com/kubernetes/client-go#compatibility-matrix

mikedanese:
This diff seems absurdly large.

glide.lock
euank: Just setting `version` to `release-3.0` and then running `glide install -v` should work I think

euank:
The diff is the difference between `glide install` and `glide install --strip-vendor` I think.

There should probably be a makefile target or script so it's clear that flag's typically being used.

tomdee:
Yes `glide install --strip-vendor` should be used

tomdee:
@mikedanese Any special reason for bumping kube-cross? (Just trying to understand it so I can make a minimal patch release if needed)

mikedanese:
@tomdee "context" (added in go1.7) is required by k8s.io/client-go but the current kube-cross has go1.6.

tomdee:
I suspect the build errors are down to the kube-cross change
`/opt/bin/flanneld: relocation error: /lib/libpthread.so.0: symbol __libc_vfork, version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference`

tomdee:
@mikedanese The kube-cross upgrade is a bit tricky. You could rebase on top of #678 or just switch to using a different build container.

mikedanese:
I tested this by hand and it seems to work fine. @aaronlevy suggested a possibly smaller patch that would fix this.

aaronlevy:
I was thinking we might just switch to using `Patch()` rather than `Update()`, which we should eventually do anyway. Then wait on changing vendor to client-go (also client-go v3 is still technically beta - so we might need to bump this again soon anyway)

Using PATCH was brought up in: https://github.com/coreos/flannel/issues/667#issuecomment-293150531 and an example https://github.com/kubernetes/kubernetes/blob/3b1d2343a84d3b8e3fd21554568e06b8594cdf26/pkg/controller/controller_utils.go#L956-L961

mikedanese:
We definitely want to do both. I do think moving to client-go sooner rather then later is safe. FWIW we are using the same packages in client-go that are used extensively in the controller-manager already. beta here means that the public API might change and you will compile errors when you rebase, but the client is tested extensively. I can look at that patch later today.

diegs:
I built flannel using this PR and did a fair bit of testing, both in host-pod and host-host networking as well as making sure that flannel correctly preserved the taints. Everything LGTM.

philips:
Gah. Needs a rebase now.

@tomdee does this LGTY?
@mikedanese can you rebase?

PR #678, opened 2017-04-13
Makefile: Switch to a new version of kube-cross

tomdee:
THis means we're now using a newer version of go and
arm-linux-gnueabihf-gcc instead of arm-linux-gnueabi-gcc


tomdee:
Includes #634

PR #679, opened 2017-04-13
flannel-docs: Updating flannel Documents files.

zbwright:
Please add content as necessary. In particular, there's some placeholder text in running - "Zero-downtime restarts - moves to an upgrade section (or standalone doc) with extra k8s content"

### DIFF
--- a/Documentation/backends.md
+++ b/Documentation/backends.md
@@ -1,51 +1,106 @@
-### Backends
-* udp: use UDP to encapsulate the packets.
-  * `Type` (string): `udp`
-  * `Port` (number): UDP port to use for sending encapsulated packets. Defaults to 8285.
+# Backends
 
-* vxlan: use in-kernel VXLAN to encapsulate the packets.
-  * `Type` (string): `vxlan`
-  * `VNI`  (number): VXLAN Identifier (VNI) to be used. Defaults to 1.
-  * `Port` (number): UDP port to use for sending encapsulated packets. Defaults to kernel default, currently 8472.
-  * `GBP` (boolean): Enable [VXLAN Group Based Policy](https://github.com/torvalds/linux/commit/3511494ce2f3d3b77544c79b87511a4ddb61dc89).  Defaults to false.
+Flannel may be paired with several different backends. Once set, the backend should not be changed at runtime.
 
-* host-gw: create IP routes to subnets via remote machine IPs.
-  Note that this requires direct layer2 connectivity between hosts running flannel.
-  * `Type` (string): `host-gw`
+VXLAN is the recommended choice. host-gw is recommended for more experienced users who want the performance improvement and whose infrastructure support it (typically it can't be used in cloud environments). UDP is suggested for debugging only or for very old kernels that don't support VXLAN.
 
-* aws-vpc: create IP routes in an [Amazon VPC route table](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Route_Tables.html).
-  * Requirements:
-	* Running on an EC2 instance that is in an Amazon VPC.
-	* Permissions required: `CreateRoute`, `DeleteRoute`,`DescribeRouteTables`, `ModifyInstanceAttribute`, `DescribeInstances [optional]`
-  * `Type` (string): `aws-vpc`
-  * `RouteTableID` (string): [optional] The ID of the VPC route table to add routes to.
-     The route table must be in the same region as the EC2 instance that flannel is running on.
-     flannel can automatically detect the id of the route table if the optional `DescribeInstances` is granted to the EC2 instance.
+AWS, GCE, and AliVPC are experimental and unsupported. Proceed at your own risk.
 
-  Authentication is handled via either environment variables or the node's IAM role.
-  If the node has insufficient privileges to modify the VPC routing table specified, ensure that appropriate `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SECURITY_TOKEN` environment variables are set when running the flanneld process.
+For more information on configuration options for cloud components, see:
+* [AliCloud VPC Backend for Flannel][alicloud-vpc]
+* [Amazon VPC Backend for Flannel][amazon-vpc]
+* [GCE Backend for Flannel][gce-backend]
 
-  Note: Currently, AWS [limits](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Appendix_Limits.html) the number of entries per route table to 50.
+## Recommended backends
 
-* gce: create IP routes in a [Google Compute Engine Network](https://cloud.google.com/compute/docs/networking#networks)
-  * Requirements:
-    * [Enable IP forwarding for the instances](https://cloud.google.com/compute/docs/networking#canipforward).
-    * [Instance service account](https://cloud.google.com/compute/docs/authentication#using) with read-write compute permissions.
-  * `Type` (string): `gce`
+### VXLAN
 
-  Command to create a compute instance with the correct permissions and IP forwarding enabled:
-  `$ gcloud compute instances create INSTANCE --can-ip-forward --scopes compute-rw`
+Use in-kernel VXLAN to encapsulate the packets.
 
-  Note: Currently, GCE [limits](https://cloud.google.com/compute/docs/resource-quotas) the number of routes for every *project* to 100.
+Type and options:
+* `Type` (string): `vxlan`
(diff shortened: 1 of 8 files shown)


Documentation/configuration.md
```diff
@@ -1,34 +1,36 @@
-## Configuration
+# Configuration
```
zbwright: We need to add a note here that flannel ships with etcd2 - for etcd3 go to xxx.

tomdee:
@zbwright It would be better to split this into smaller PRs (if the changes don't overlap) as that makes it much easier for me to review and merge. You could even go as far as one PR per file/document.

tomdee:
So far I've looked at `Documentation/alicloud-vpc-backend.md` and `Documentation/aws-vpc-backend.md` and both look fine

zbwright:
@tomdee sorry about that - still learning how this works. I will certainly do smaller PRs next time. Would you like me to figure out how to do it to this one as well?

Documentation/backends.md
```diff
@@ -1,51 +1,106 @@
...
-  * Requirements:
-    * Running on an ECS instance that is in an Alicloud VPC.
-    * Permission require accessid and keysecret
-  * `Type` (string): `ali-vpc`
-  * `AccessKeyID` (string): api access key id. can also be configure with environment ACCESS_KEY_ID
-  * `AccessKeySecret` (string): api access key secret.can also be configure with environment ACCESS_KEY_SECRET
-  Note: Currently, AliVPC limit the number of entries per route table to 50.
+# Backends
+
+Flannel may be paired with several different backends. Once set, the backend cannot be changed at runtime.
```
tomdee: Maybe it should say "should not" instead of "cannot". People _can_ change it and if they *really* know what they're doing  it might be OK, but we don't support or recommend it.

Documentation/backends.md
```diff
@@ -1,51 +1,106 @@
...
+
+VXLAN is the default choice. host-gw is recommended for more experienced users who want the performance improvement. UDP is suggested for debugging only.
```
tomdee: This recommendation is really important to get right so apologies if this seems nit-picky! 
* In the code, UDP is still the default (for historical and maybe compatibility reasons).
* host-gw should mention the infrastructure restriction too. 

So, maybe something like this
>VXLAN is the recommended choice. host-gw is recommended for more experienced users who want the performance improvement and who's infrastructure support it (typically it can't be used in cloud environments). UDP is suggested for debugging only or for very old kernels that don't support vxlan.

Documentation/backends.md
```diff
@@ -1,51 +1,106 @@
...
+* `Type` (string): `vxlan`
+* `VNI` (number): VXLAN Identifier (VNI) to be used. Defaults to 1.
+* `Port` (number): UDP port to use for sending encapsulated packets. Defaults to kernel default, currently 8472.
+* `GBP` (Boolean): Enable [VXLAN Group Based Policy](https://github.com/torvalds/linux/commit/3511494ce2f3d3b77544c79b87511a4ddb61dc89).  Defaults to `false`.
+
+### host-gw
+
+Use host-gw to create IP routes to subnets via remote machine IPs. Requires direct layer2 connectivity between hosts running flannel.
+
+host-gw provides good performance, with few dependencies, and easy set up. http://machinezone.github.io/research/networking-solutions-for-kubernetes/
```
tomdee: I would probably not include this machinezone link as it doesn't give a very clear view of current vxlan performance

Documentation/backends.md
```diff
@@ -1,51 +1,106 @@
...
+
+### UDP
+
+Use UDP only for debugging if your network and kernel prevent you from using VXLAN or host-gw.
+
+Type and options:
+* `Type` (string): `udp`
+* `Port` (number): UDP port to use for sending encapsulated packets. Defaults to 8285.
+
+### AliVPC
```
tomdee: This one should be moved under experimental too

Documentation/configuration.md
tomdee: I'm not sure I understand this comment? We should have info on setting kubernetes configuration, but etcd3 support isn't something we have at the moment,

Documentation/configuration.md
```diff
@@ -1,34 +1,36 @@
...
+   Defaults to the first subnet of `Network`.
 
 * `SubnetMax` (string): The end of the IP range at which the subnet allocation should end with.
-   Defaults to the last subnet of Network.
+   Defaults to the last subnet of `Network`.
 
 * `Backend` (dictionary): Type of backend to use and specific configurations for that backend.
    The list of available backends and the keys that can be put into the this dictionary are listed below.
-   Defaults to "udp" backend.
+   Defaults to `udp` backend.
```
tomdee: Would it be worth going through the docs and making sure that backends are consistently `backticked`

Documentation/reporting_bugs.md
```diff
@@ -0,0 +1,36 @@
+# Reporting bugs
```
tomdee: Awesome, thanks for adding this.

Documentation/running.md
```diff
@@ -1,40 +1,100 @@
...
+* An interface (used for MTU detection and selecting the VTEP MAC in VXLAN).
+* An IP address for that interface.
+* A public IP that can be used for reaching this node. In `host-gw` it should match the interface address.
+
+### Making changes at runtime
+
+Please be aware of the following flannel runtime limitations.
+* The datastore type cannot be changed.
+* The backend type cannot be changed. (It can be changed if you stop all workloads and restart all flannel daemons.)
+* You can change the the subnetlen/subnetmin/subnetmax with a daemon restart. (Subnets can be changed with caution. If pods are already using IP addresses outside the new range they will stop working.)
```
tomdee: the the

Documentation/running.md
```diff
@@ -1,40 +1,100 @@
...
 Vagrant has a tendency to give the default interface (one with the default route) a non-unique IP (often 10.0.2.15).
+
 This causes flannel to register multiple nodes with the same IP.
+
 To work around this issue, use `--iface` option to specify the interface that has a unique IP.
-If you're running on CoreOS, use cloud-config to set `coreos.flannel.interface` to `$public_ipv4`.
+
+If you're running on CoreOS, use `cloud-config` to set `coreos.flannel.interface` to `$public_ipv4`.
+
+## Zero-downtime restarts - moves to an upgrade section (or standalone doc) with extra k8s content
```
tomdee: Ooops, this shouldn't be in the title

Documentation/configuration.md
zbwright: ok - I was overly worried about the etcd2 > 3 stuff.

Documentation/backends.md
zbwright: nit picky and precise is good.

Documentation/configuration.md
zbwright: I think I've done it, but will do again.

zbwright: And I don't know why, but I cannot reply to your previous comment. The etcd comment must have been original to the doc. Where can I find info on setting up k8s config?

Documentation/running.md
zbwright: removing comment for now - we can address later.

Documentation/configuration.md
joshix: by way of explanation, @zbwright, etcd clients like flannel can use the etcd v3 _or_ v2 apis when communicating with an etcd v3 service. So and etcd v3 binary can serve either the v2 http api and the v3 grpc api to a client. The two apis cannot be used interchangeably and they have separate data stores.

(No change required for this comment)

Documentation/running.md
```diff
@@ -18,32 +19,91 @@ flanneld -subnet-file /vxlan.env -etcd-prefix=/vxlan/network
...
+When running with a backend other than `udp`, the kernel is providing the data path with `flanneld` acting as the control plane.
+
+As such, `flanneld` can be restarted (even to do an upgrade) without disturbing existing flows.
+
+However in the case of `vxlan` backend, this needs to be done within a few seconds as ARP entries can start to timeout requiring the flannel daemon to refresh them.
+
+Also, to avoid interruptions during restart, the configuration must not be changed (e.g. VNI, --iface values).
+
+
+[coreos-etcd]: https://github.com/coreos/etcd/releases/latest
```
joshix: prefer https://github.com/coreos/etcd/blob/master/Documentation/dev-guide/local_cluster.md

Documentation/configuration.md
tomdee: 👍 So I think this comment is resolved right?

zbwright: yes. resolved.

PR #681, opened 2017-04-14
switch kube subnet manager to PATCH

mikedanese:
This works too.

@aaronlevy

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -29,12 +29,14 @@ import (
 	"github.com/golang/glog"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/client/cache"
 	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/internalclientset"
 	"k8s.io/kubernetes/pkg/client/restclient"
 	"k8s.io/kubernetes/pkg/controller/framework"
 	"k8s.io/kubernetes/pkg/runtime"
 	utilruntime "k8s.io/kubernetes/pkg/util/runtime"
+	"k8s.io/kubernetes/pkg/util/strategicpatch"
 	"k8s.io/kubernetes/pkg/util/wait"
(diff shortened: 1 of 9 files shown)


philips:
This LGTM but I haven't tested it.

subnet/kube/kube.go
```diff
@@ -231,7 +233,31 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, network string,
...
+		if err != nil {
+			return nil, err
+		}
+
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, v1.Node{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create patch for node %q: %v", ksm.nodeName, err)
+		}
+
+		_, err = ksm.client.Core().Nodes().Patch(ksm.nodeName, api.StrategicMergePatchType, patchBytes, "status")
```
aaronlevy: What does providing the "status" subresource do/mean?

mikedanese: @aaronlevy it's a subresource under node that only allows updates to .status fields and annotations. It's less privileged then the node resource and can be acl'd seperately. This code was written pre-rbac and I'm trying to reduce the privilege required by kube-flannel.

https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/node/strategy.go#L122 status strategy zeros out changes to spec.

mikedanese: Also see changes to rbac file.

aaronlevy: ah k. sgtm

aaronlevy:
@diegs and/or I will try and start testing this and https://github.com/coreos/flannel/pull/677 today.

mikedanese:
I tested this by hand on friday. We should probably merge this, cut a release then consider merging #677.

aaronlevy:
The changes themselves LGTM to me then.

mikedanese:
(more hand testing is always appreciated)

diegs:
Also hand-tested this one: launched a node using `--register-with-taints=key=value:PreferNoSchedule`. Verified that after flannel configured the node the taint was still present. Did a bunch of host-pod and pod-pod network checks and all looked good.

PR #687, opened 2017-04-19
vendor: remove files not needed for building

heyitsanthony:
Ran `glide vc --only-code --no-tests` over the repo. Includes a makefile rule for future use.


tomdee:
Ahhh, what a great PR - removing 5 million lines of code!

Does this need a particular (recent) version of glide? If so, could you add a comment stating that in the Makefile?

heyitsanthony:
@tomdee I don't think it needs a bleeding edge version of glide; I suspect the makefile will fail if it's not up to date. I added comments about where to get `glide` and `glide vc` so there's a hint for resolving any breakage from not having it.

etcd's [deps script](https://github.com/coreos/etcd/blob/master/scripts/updatedep.sh) goes so far as to force an exact repo revision, but that seemed excessive for what this PR is trying to do.

PR #702, opened 2017-05-03
bill-of-materials: initial commit

philips:
CoreOS projects are adding a machine parsable license file. The tool is
from https://github.com/coreos/license-bill-of-materials


philips:
This is used to generate https://coreos.com/legal/open-source/

PR #632, opened 2017-03-09
flanneld: adds basic health-check endpoint

genevievelesperance:
This begins to address #616.

We can follow up later with a richer health-check that captures connectivity to etcd.

cc: @rosenhouse

PR #722, opened 2017-05-17
add healthz

### LINKED ISSUE #616: Healthcheck endpoint
As an operator, I would like to be alerted if my `flanneld` is unable to connect to etcd.

Currently, if flannel loses its connection to etcd, it prints error messages to stderr but does not fail.

I could write some tooling to process those logs and alert me if certain strings are printed.

But I'd rather have some kind of healthcheck, e.g. I set up a script to periodically `curl` a special endpoint on `localhost`.  If I get back a `200 OK` then I know `flanneld` is healthy.  Otherwise, I can raise an alert, maybe restart the VM, etc.

I could imagine the healthcheck returning some basic information, like when it last renewed its lease with etcd.  But the important thing is a simple status code that could be easily interpreted and acted on.

A CNI plugin could even probe this healthcheck during the `ADD` action.  That way it could ensure that `flanneld` is alive, and that the `subnet.env` file on disk is up-to-date.

Would you be open to a PR like this?

cc: @rusha19 @mcwumbly @jaydunk

andyxning:
Fix #616 

This PR will add a `healthz` endpoint in http for flanneld by default in `0.0.0.0:8285`.

Note: This PR is based on #632 .

/cc @eyakubovich @tomdee @zbwright

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -58,6 +58,8 @@ The following configuration illustrates the use of most options with `udp` backe
 --subnet-lease-renew-margin=60: subnet lease renewal margin, in minutes.
 --ip-masq=false: setup IP masquerade for traffic destined for outside the flannel network. Flannel assumes that the default policy is ACCEPT in the NAT POSTROUTING chain.
 -v=0: log level for V logs. Set to 1 to see messages related to data path.
+--healthz-ip="0.0.0.0": The IP address for healthz server to listen (default "0.0.0.0")
+--healthz-port=0: The port for healthz server to listen(0 to disable)
 --version: print version and exit
 ```
 
@@ -68,3 +70,9 @@ MTU is calculated and set automatically by flannel. It then reports that value i
 The command line options outlined above can also be specified via environment variables.
 For example `--etcd-endpoints=http://10.0.0.2:2379` is equivalent to `FLANNELD_ETCD_ENDPOINTS=http://10.0.0.2:2379` environment variable.
 Any command line option can be turned into an environment variable by prefixing it with `FLANNELD_`, stripping leading dashes, converting to uppercase and replacing all other dashes to underscores.
+
+## Health Check
+
+Flannel provides a health check http endpoint `healthz`. Currently this endpoint will blindly
+return http status ok(i.e. 200) when flannel is running. This feature is by default disabled.
+Set `healthz-port` to a non-zero value will enable a healthz server for flannel.
--- a/main.go
+++ b/main.go
@@ -19,11 +19,13 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"strings"
 	"syscall"
+	"strconv"
 
 	"github.com/coreos/pkg/flagutil"
 	log "github.com/golang/glog"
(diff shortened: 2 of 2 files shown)


andyxning:
@eyakubovich @tomdee @zbwright Please take a look at the CI error. It seems nothing is related with the change. Seems there are more than one flannel instance is running simultaneous in `./dist/functional-test.sh`.

tomdee:
@andyxning I'm not sure I see the benefit of this PR - what's it actually checking (except maybe that the Go runtime hasn't crashed)

andyxning:
@tomdee Currently, flannel listen on udp port 8285 by default, we have no methods to do a health check about the status of flanneld daemon, i.e., whether it is running or stopped accidentally. We have encountered a situation where flanneld has been stopped accidentally for more than two days without any notice, and when we found this and restart it, the newly allocated subnet is different from the original cause we have not renew the lease about the original subnet. This makes it possible that we have two nodes with the same subnet. 

So, the original thought is we need a way to check whether flanneld is running or not. No much other logic. :)

tomdee:
I'm not totally against this, but before I can merge it you need to add some documentation for it and ensure the tests pass. I also think it should be an optional feature - and probably off by default.

andyxning:
@tomdee Will Do.

andyxning:
@tomdee Done. PTAL.

andyxning:
Friendly ping @tomdee

andyxning:
@tomdee  This health check is disable by default. I have not test whether it can pass all the tests when enabled by default. :)

andyxning:
Friendly Ping @tomdee :)

Documentation/configuration.md
```diff
@@ -68,3 +70,9 @@ MTU is calculated and set automatically by flannel. It then reports that value i
 The command line options outlined above can also be specified via environment variables.
 For example `--etcd-endpoints=http://10.0.0.2:2379` is equivalent to `FLANNELD_ETCD_ENDPOINTS=http://10.0.0.2:2379` environment variable.
 Any command line option can be turned into an environment variable by prefixing it with `FLANNELD_`, stripping leading dashes, converting to uppercase and replacing all other dashes to underscores.
+
+## Health Check
+
+Flannel provides an health check http endpoint `healthz`. Currently this endpoint will blindly
```
andyxning: Will do.

main.go
```diff
@@ -378,3 +388,18 @@ func WriteSubnetFile(path string, nw ip.IP4Net, ipMasq bool, bn backend.Network)
...
+	log.Infof("Start healthz server on %s", address)
+
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("flanneld is running"))
+	})
+
+	if err := http.ListenAndServe(address, nil); err != nil {
+		log.Errorf("Start healthz server error. %v", err)
+		panic(err)
```
andyxning: Nope. `panic` because this method is called `mustRunHealthz`. According to golang convention, we should panic when we can not do some thing that must be done.

The reason for why we use `mustRunHealthz` is because is think we should must start the health check server once we enable it. If not, we should panic or stop flannel to give user more clear result. 

@tomdee WDYT.

andyxning:
@tomdee Many thanks. 👏

PR #726, opened 2017-05-17
Clean up kube subnet manager

tomdee:


subnet/kube/kube.go
```diff
@@ -300,18 +270,28 @@ func nodeToLease(n v1.Node) (l subnet.Lease, err error) {
...
 	l.Attrs.BackendType = n.Annotations[backendTypeAnnotation]
 	l.Attrs.BackendData = json.RawMessage(n.Annotations[backendDataAnnotation])
+
 	_, cidr, err := net.ParseCIDR(n.Spec.PodCIDR)
 	if err != nil {
 		return l, err
 	}
+
 	l.Subnet = ip.FromIPNet(cidr)
-	l.Expiration = time.Now().Add(24 * time.Hour)
```
gunjan5: Just thinking out loud here, would there be an impact of not specifying the time here? if you leave it blank it will take the `zero-value` for `time.Time` which is `0001-01-01 00:00:00 +0000 UTC`

tomdee: I don't think there's an impact because nothing should be looking at this expiration.

main.go
```diff
@@ -203,8 +203,13 @@ func main() {
 
 	daemon.SdNotify(false, "READY=1")
 
-	// Block waiting to renew the lease
-	_ = MonitorLease(ctx, sm, bn)
+	if opts.kubeSubnetMgr {
```
tomdee: Good point, I'll add one.

PR #737, opened 2017-05-30
Add route table fetching using filters

alexey-medvedchikov:
## Description

Add filter-based routing table search into AWS backend, i.e. we can add RouteTableFilter string list using `Name=Value` format to configuration and this filters will be used to get required routing tables.

## Todo
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/backend/awsvpc/awsvpc.go
+++ b/backend/awsvpc/awsvpc.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


tomdee:
Hi @alexey-medvedchikov thanks for this PR! Could you write some documentation outlining why this is an important feature from a user's perspective? 

It's also really useful for me to know what testing you've done. What's the risk that this will introduce a regression for people not using this feature?

alexey-medvedchikov:
@tomdee yeah, sure. I've just expected you to check out code for now.

We're using flannel in our proprietary K8S bundle with multiAZ AWS setup. We don't have direct communication between terraform and provisioner, so it's impossible to get RT IDs directly. Other option is to use awscli and pass RT IDs into flanneld configuration, but it requires awscli to be installed and some bash glue. From our perspective it's much easier to do this job inside flannel, since it introduces one simple function.

I've made simple tests with filters configured, checked connectivity. I think AWS e2e tests will be awesome addition.

I don't see any risk for people not using this feature, since missing sections in config not introducing ones and it's explicitly coded "if no filters set do the usual way".

tomdee:
Sounds great to me! If you can add some documentation and hopefully some tests I can get it merged

tomdee:
@alexey-medvedchikov Could you add some documentation to the `aws-vpc-backend.md` file and I can merge this.

tomdee:
I would love to take this but it looks like the original author has gone away. If anyone wants to take it on, please rebase, add some documentation and open another PR.

Issue #29, opened 2014-08-29
Handle subnet lease getting expired


eyakubovich:
Although flannel will start renewing the lease an hour prior to expiration, it could still get lost: e.g. VM getting suspended. Flannel should try to get the same subnet assignment if it's still available but fall back to a new lease and signal the fact.

macb:
Is there any work under way for this? It'd be incredibly useful as right now if a machine loses a lease and gets a new one it renders any containers on the machine with no network connectivity.

tomdee:
One implementation of this is in #610

tomdee:
Also see #520 for some good questions about how flannel handles this at the moment.

tomdee:
When fixing this, we should make sure this failure scenario is discussed clearly in the docs.

rosenhouse:
FWIW, the system design that we've converged on for Cloud Foundry is that hosts are preferentially assigned their prior lease, even if it "expired."  And if a new host appears, it is assigned a lease in the following priority order:
- prefer subnets that have never been given out before, or subnets which were explicitly relinquished by a cleanly-terminating host.
- if none of those exist, only then does the new host take over an expired lease, and in that case it chooses the oldest such lease.

This is meant to minimize the probability that a lease is "stolen" from a live, but partitioned, container host.  But if that does occur, once the partition heals and the "victim" host re-connects, it will discover that its lease is no longer valid.  In this case, the victim host falls into a special, noisy failure mode which will (1) prevent any new workloads from being scheduled and (2) trigger the orchestration system to evacuate any existing workloads.  Once the evacuation is complete, the host will clean up any leftover networking state (e.g. remove the VXLAN device), acquire a new lease for itself and begin accepting new workloads.

We **think** this is the right plan.   Feedback welcome.

tomdee:
This is now fixed in v0.8.0

PR #752, opened 2017-06-19
flannel reads from created subnet.env file on startup

### LINKED ISSUE #610: Feature idea: on startup, read subnet.env and attempt to acquire that lease
### Background

We are concerned about preserving network connectivity for containers in the face of etcd outages and data loss.  For example, during upgrades of etcd, Cloud Foundry users have encountered various "split-brain"-like scenarios.   These scenarios are most easily resolved by stopping all etcd nodes, removing all the data directories, and finally restarting all the nodes.  But CF users expect their containers to remain connected to the network throughout.  Therefore, we prefer that etcd clients like flannel be resilient to etcd outages and data loss. 

### Problem
In manual testing of etcd outages that included data loss, we've found that flannel networks retain connectivity *only* if the `flanneld` processes remain alive throughout the outage.

Consider this sequence:
1. deploy container hosts A & B, running `flanneld` connected to a remote etcd cluster.
2. launch some containers on each host, using flannel CNI plugin; verify network connectivity between containers on different hosts
3. etcd outage begins: stop all etcd nodes
4. restart flanneld on host A; leave it running on host B
5. etcd data loss: `rm -rf ` the etcd data directories on every etcd node
6. etcd outage ends: start all etcd nodes (but with empty data directories)
7. check network connectivity across hosts

In testing, we've found that after step 6, host B correctly re-acquires its existing subnet lease -- the one in use by its running containers.  But host A does not: instead it picks a new subnet at random, which invariably does not match the IP addresses of containers it is hosting.  As a result, the containers on host A cannot reach host B. 

### Proposed solution

On startup, `flanneld` reads from the `subnet.env` file, if it exists.  If it contains a subnet valid for the current configuration, then `flanneld` attempts to re-acquire that subnet from etcd.  In the case of a conflict, `flanneld` falls back to picking a new subnet at random (current behavior).

**What do y'all think?  Would you be open to a PR?**

### References
- [Issue 520: "What should flannel do when it loses a lease?"](https://github.com/coreos/flannel/issues/520)
- [etcd disaster recovery in Cloud Foundry](https://github.com/cloudfoundry-incubator/etcd-release#failed-deploys-upgrades-split-brain-scenarios-etc)
- [more guidance on etcd cluster failures in Cloud Foundry](https://discuss.pivotal.io/hc/en-us/articles/228109388-How-to-recover-ETCD-cluster-after-failure)
- [notes on our investigation](https://www.pivotaltracker.com/story/show/137627203)

cc: @genevievelesperance @jaydunk @rusha19

mgleung:
## Description
Added feature to allow flannel to restart in case of etcd failures and
still keep the same subnet address for the hosts.

Fixes #610 #29

## Testing
Added unit tests in subnet_test.go

Ran etcd locally and killed the process.  Made sure that restarting flannel after an etcd restart produced the same subnet address on the lease.

Ran etcd locally and killed the process.  When restarting etcd, changed the network configuration so the previous subnet address for the lease no longer matched the SubnetLen.  On running flannel, checked that a new subnet was allocated for the lease with the correct subnet length.

Ran etcd locally and killed the process. When restarting etcd, changed the network address so the previous subnet address did not fall within it.  On running flannel, checked that a new subnet was allocated for the lease in the appropriate range.


tomdee:
Can you include the documentation change in this PR too?

ironlung975:
Added some documentation.  Let me know if it is clear enough.

Documentation/reservations.md
```diff
@@ -25,6 +25,16 @@ This shows that there is a single lease (`10.5.34.0/24`) which will expire in 85
 The `"PublicIP"` value is how flannel knows to reuse this lease when restarted. 
 This means that if the public IP changes, then the flannel subnet will change too.
 
+In case something happens to etcd and it is either unavailable or loses the record of its leases, flannel will then attempt to renew the last lease that it has saved in its subnet config file (which, unless specified, is located at `/var/run/flannel/subnet.env`)
```
tomdee: I'm not sure about this wording - if etcd is unavailable then flannel can't start, and why would etcd lose the record of its leases?

Documentation/reservations.md
```diff
@@ -25,6 +25,16 @@ This shows that there is a single lease (`10.5.34.0/24`) which will expire in 85
...
+```bash
+cat /var/run/flannel/subnet.env
+FLANNEL_NETWORK=10.5.0.0/16
+FLANNEL_SUBNET=10.5.34.1/24
+FLANNEL_MTU=1450
+FLANNEL_IPMASQ=false
+```
+In this case, if flannel fails to retrieve a lease from etcd, it will attempt to renew lease specified in `FLANNEL_SUBNET` (`10.5.34.1/24`).  It will only renew this lease if the subnet specified is valid for the current etcd network configuration otherwise it will allocate a new lease.
```
tomdee: retrieve an *existing* lease from etcd,

Documentation/running.md
```diff
@@ -46,6 +46,7 @@ FLANNEL_SUBNET=10.5.72.1/24
 FLANNEL_MTU=1450
 FLANNEL_IPMASQ=false
 ```
+The `FLANNEL_SUBNET` value written in this subnet config file will be checked each time flannel starts and will be reused if the subnet still matches the values stored in the etcd config.
```
tomdee: This is a bit ambiguous so will be confusing for uses. Can you make it clearer and more explicit?

main.go
```diff
@@ -410,3 +414,17 @@ func mustRunHealthz() {
 		panic(err)
 	}
 }
+
+func ReadSubnetFromSubnetFile(path string, config *etcdv2.EtcdConfig) {
```
tomdee: If you're adding a new library for reading the file, why not use the same library for writing the file?

subnet/etcdv2/local_manager.go
```diff
@@ -33,7 +33,8 @@ const (
 )
 
 type LocalManager struct {
-	registry Registry
+	registry       Registry
+	previousSubnet ip.IP4Net
```
tomdee: What was the reasoning for choosing this mechanism for passing the previousSubnet through to the tryAcquireLease function?

subnet/etcdv2/registry.go
```diff
@@ -48,13 +48,14 @@ type Registry interface {
...
-	Username  string
-	Password  string
+	Endpoints      []string
+	Keyfile        string
+	Certfile       string
+	CAFile         string
+	Prefix         string
+	Username       string
+	Password       string
+	PreviousSubnet ip.IP4Net
```
tomdee: What's the reasoning behind adding this?

main.go
mgleung: godotenv only reads from env files sadly

subnet/etcdv2/local_manager.go
mgleung: I wanted to avoid changing the `AcquireLease` function to accept additional arguments since that would require the kube manager to also be updated when it does not need a subnet ip (since it is not affected by this).  I could change it but I wanted to avoid changing the interface signatures.

subnet/etcdv2/registry.go
mgleung: Since I was adding a variable to the local manager, I thought that since it was currently being built according to the values passed in from this `EtcdConfig` that this would be the best place to pass in that value.  Looking back at it, this config really only applies to values used to set up a connection to etcd so the value should be moved out for organizational reasons.  Downside of moving it out is it clutters up the `NewLocalManager` constructor a little.

tomdee:
@mgleung as discussed offline, this looks good to go but it's back with you to make the change in the EtcdConfig

Issue #620, opened 2017-02-24
Support more flexible --iface option to ease flannel use in Daemonset


tomdee:
We could allow the `--iface` option to appear multiple times or maybe even allowing a regex.

Originally raised by user on mailinglist
> How can I define a single definition of DaemonSet to create slightly different pods from it?
> What I mean by different pods is that pods created from different commands.
> I currently make two different DaemonSet objects to create two different kinds of pods.
> 
> Let me tell you the details.
> 
> I've been modifying kube-flannel.yml, especially the definition of DaemonSet.
> I want to add "--iface=<ethernet interface>" to flanneld [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr"]
> and I want to use two different ethernet interfaces based on node label. 
> 
> I have three different kinds of nodes:
> 1) a master node with two ethernet interfaces: eth0 for internal network and eth3 for external network 
> 2) slave nodes with one ethernet interfaces: eth0 for internal network
> 3) a newly added node with one ethernet interfaces: p2p1 for internal network 
> 
> When I had 1) and 2), I was okay with the following command:
> [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr", "--iface=--eth0"] 
> I have to add --iface option because otherwise flanneld on the master node automatically detects eth3 other than eth0, 
> which is the reason I manually add the iface option.
> 
> Now that I have 3),
> I have to make two DaemonSet definitions with different commands:
> [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr", "--iface=--eth0"]  for 1) and 2)
> , [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr", "--iface=--p2p1"] for 3) 
> Label selector is used to tell which DaemonSet is used for a node. 
> The entire definition is attached at the end. 
> 
> Is there a more elegant way of creating slightly different pods from a single DaemonSet definition?
> It would be nice to have only a single definition with a placeholder which can be substituted by the value of a specific label. 
> 
> Otherwise I can do this by setting flannel in a different way?
> It would be nice if we can give a list of ethernet interfaces to --iface option but my colleague found that only one interface/ip can be given to --iface option.
> 
> Thanks.
> 
```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: flannel
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: kube-flannel-cfg
  labels:
    tier: node
[...]

PR #754, opened 2017-06-20
Added iface-regex option

mgleung:
## Description
Allows for more flexibility when selecting the network interface that
flannel should be using.

Addresses #620

## Testing
Ran flannel locally and set the `iface-regex` flag to various regex expressions to check if they match (or do not match) with any of the network interfaces on the machine.

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -54,6 +54,7 @@ The following configuration illustrates the use of most options with `udp` backe
 --etcd-cafile="": SSL Certificate Authority file used to secure etcd communication.
 --kube-subnet-mgr: Contact the Kubernetes API for subnet assignment instead of etcd.
 --iface="": interface to use (IP or name) for inter-host communication. Defaults to the interface for the default route on the machine.
+--iface-regex="": regex expression to match the first interface to use (IP or name) for inter-host communication. If unspecified, will default to the interface for the default route on the machine. This option is superseded by the iface option and will not be used if the iface option is also specified.
 --subnet-file=/run/flannel/subnet.env: filename where env variables (subnet and MTU values) will be written to.
 --subnet-lease-renew-margin=60: subnet lease renewal margin, in minutes.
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -324,6 +327,49 @@ func LookupExtIface(ifname string) (*backend.ExternalInterface, error) {
 				return nil, fmt.Errorf("error looking up interface %s: %s", ifname, err)
 			}
 		}
+	} else if len(ifregex) > 0 {
+		// Use the regex if nothing matches an explicit name or ip
```
tomdee: This comment doesn't match the code

Documentation/configuration.md
```diff
@@ -54,6 +54,7 @@ The following configuration illustrates the use of most options with `udp` backe
 --etcd-cafile="": SSL Certificate Authority file used to secure etcd communication.
 --kube-subnet-mgr: Contact the Kubernetes API for subnet assignment instead of etcd.
 --iface="": interface to use (IP or name) for inter-host communication. Defaults to the interface for the default route on the machine.
+--iface-regex="": regex expression to match the first interface to use (IP or name) for inter-host communication. If unspecified, will default to the interface for the default route on the machine.
```
tomdee: This doesn't mention that it's only used if `--face` isn't specified

main.go
```diff
@@ -89,6 +91,7 @@ func init() {
 	flag.StringVar(&opts.etcdUsername, "etcd-username", "", "Username for BasicAuth to etcd")
 	flag.StringVar(&opts.etcdPassword, "etcd-password", "", "Password for BasicAuth to etcd")
 	flag.StringVar(&opts.iface, "iface", "", "interface to use (IP or name) for inter-host communication")
+	flag.StringVar(&opts.ifaceRegex, "iface-regex", "", "regex expression to match the first interface to use (IP or name) for inter-host communication")
```
tomdee: It's also worth saying that it's overridden by iface here too

tomdee:
One more small comment, otherwise LGTM.

PR #759, opened 2017-06-29
Allow kube subnet manager to run outside of kubernetes

tomdee:
This is really useful for dev and test work

This PRs also fixes some problems in the vendoring and adds a target to
the Makefile to make it easy to run.


Makefile
```diff
@@ -231,3 +231,22 @@ run-etcd: stop-etcd
...
+
+stop-k8s-apiserver:
+	@-docker rm -f calico-k8s-apiserver
+
+run-local-kube-flannel-with-prereqs: run-etcd run-k8s-apiserver dist/flanneld
+	while ! kubectl apply -f dist/fake-node.yaml; do sleep 1; done
+	$(MAKE) run-local-kube-flannel
+
+run-local-kube-flannel:
+	# Current requires the netconf to be in /etc/kube-flannel/net-conf.json
```
gunjan5: do you mean `"Currently this..."` here?

main.go
```diff
@@ -67,6 +67,8 @@ type CmdLineOpts struct {
 	help                   bool
 	version                bool
 	kubeSubnetMgr          bool
+	kubeApiUrl             string
```
gunjan5: nit: it's a golang best practice to keep all the acronym characters same case in a variable name, but not a big deal

main.go
```diff
@@ -99,6 +101,8 @@ func init() {
 	flag.IntVar(&opts.subnetLeaseRenewMargin, "subnet-lease-renew-margin", 60, "Subnet lease renewal margin, in minutes.")
 	flag.BoolVar(&opts.ipMasq, "ip-masq", false, "setup IP masquerade rule for traffic destined outside of overlay network")
 	flag.BoolVar(&opts.kubeSubnetMgr, "kube-subnet-mgr", false, "Contact the Kubernetes API for subnet assignment instead of etcd.")
+	flag.StringVar(&opts.kubeApiUrl, "kube-api-url", "", "Kubernetes API server URL. Does not need to be specified if flannel is running in a pod.")
+	flag.StringVar(&opts.kubeConfigFile, "kubeconfig-file", "", "kubeconfig file location. Does not need to be specified if flannel is running in a pod.")
```
gunjan5: nit: I see flag help strings have mixed capitalization. Some start uppercase some don't

main.go
```diff
@@ -107,7 +111,7 @@ func init() {
 
 func newSubnetManager() (subnet.Manager, error) {
 	if opts.kubeSubnetMgr {
-		return kube.NewSubnetManager()
+		return kube.NewSubnetManager(opts.kubeApiUrl, opts.kubeConfigFile)
```
gunjan5: I notice we are not validating the URL/File path before passing it to the subnet mgr, might want to add a quick validation using `url.ParseRequestURI` and `os.Stat` so the request fails quickly on the client side instead of going higher up the stack

subnet/kube/kube.go
```diff
@@ -67,29 +68,44 @@ type kubeSubnetManager struct {
...
-func NewSubnetManager() (subnet.Manager, error) {
-	cfg, err := rest.InClusterConfig()
-	if err != nil {
-		return nil, fmt.Errorf("unable to initialize inclusterconfig: %v", err)
+func NewSubnetManager(apiUrl, kubeconfig string) (subnet.Manager, error) {
+
+	var cfg *rest.Config
+	var err error
+	// Use out of cluster config if the URL or kubeconfig have been specified. Otherwise use incluster config.
+	if apiUrl != "" || kubeconfig != "" {
```
gunjan5: We might not need to check if the flags are passed or do an InCluster if they're not, `BuildConfigFromFlags` takes care of that:
```go

func BuildConfigFromFlags(masterUrl, kubeconfigPath string) (*restclient.Config, error) {
	if kubeconfigPath == "" && masterUrl == "" {
		glog.Warningf("Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.")
		kubeconfig, err := restclient.InClusterConfig()
		if err == nil {
			return kubeconfig, nil
		}
		glog.Warning("error creating inClusterConfig, falling back to default config: ", err)
	}
	return NewNonInteractiveDeferredLoadingClientConfig(
		&ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
		&ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: masterUrl}}).ClientConfig()
}
```

So the whole `else` block can also go I think

subnet/kube/kube.go
```diff
@@ -67,29 +68,44 @@ type kubeSubnetManager struct {
...
-
-	pod, err := c.Pods(podNamespace).Get(podName, metav1.GetOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("error retrieving pod spec for '%s/%s': %v", podNamespace, podName, err)
-	}
-	nodeName := pod.Spec.NodeName
+	nodeName := os.Getenv("NODE_NAME")
 	if nodeName == "" {
-		return nil, fmt.Errorf("node name not present in pod spec '%s/%s'", podNamespace, podName)
+		podName := os.Getenv("POD_NAME")
```
gunjan5: I might be misunderstanding this, but isn't the whole point of this so flannel can run outside the cluster? How would it get `POD_NAME` and `POD_NAMESPACE` env var populated if it's not running in a k8s pod? Or should these be manually set while starting flannel outside a k8s pod? I think Line 93/94 could use a short comment explaining this

main.go
tomdee: I changed all to lowercase

main.go
tomdee: I don't think it's needed since this is a dev/test focused feature, it's more important to keep the code clean and small here.

subnet/kube/kube.go
tomdee: I chose to do it the way I did to maintain existing behavior. I saw the code you mention but that would change the log messages and it would fall back to using a default config if incluster fails, which isn't what we want.

subnet/kube/kube.go
tomdee: Added a comment

Makefile
```diff
@@ -231,3 +231,22 @@ run-etcd: stop-etcd
 
 stop-etcd:
 	@-docker rm -f flannel-etcd
+
+K8S_VERSION=1.6.1
```
gunjan5: actually k8s v1.7.0 was just released if you want to use that instead :)

PR #762, opened 2017-06-29
main.go: Fix logging options

tomdee:
Only display the logging options that actually work

### DIFF
--- a/main.go
+++ b/main.go
@@ -82,27 +82,53 @@ var (
 	opts           CmdLineOpts
 	errInterrupted = errors.New("interrupted")
 	errCanceled    = errors.New("canceled")
+	flannelFlags   = flag.NewFlagSet("flannel", flag.ExitOnError)
 )
 
 func init() {
-	flag.StringVar(&opts.etcdEndpoints, "etcd-endpoints", "http://127.0.0.1:4001,http://127.0.0.1:2379", "a comma-delimited list of etcd endpoints")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -82,27 +82,54 @@ var (
...
+	flannelFlags.BoolVar(&opts.ipMasq, "ip-masq", false, "setup IP masquerade rule for traffic destined outside of overlay network")
+	flannelFlags.BoolVar(&opts.kubeSubnetMgr, "kube-subnet-mgr", false, "Contact the Kubernetes API for subnet assignment instead of etcd.")
+	flannelFlags.BoolVar(&opts.version, "version", false, "print version and exit")
+	flannelFlags.StringVar(&opts.healthzIP, "healthz-ip", "0.0.0.0", "The IP address for healthz server to listen")
+	flannelFlags.IntVar(&opts.healthzPort, "healthz-port", 0, "The port for healthz server to listen(0 to disable)")
+
+	// glog will log to tmp files by default. override so all entries
+	// can flow into journald (if running under systemd)
+	flag.Set("logtostderr", "true")
+	flag.Set("stderrthreshold", "FATAL")
```
gunjan5: why is the threshold set to FATAL? Why not ERROR?

tomdee: Ooops, actually that line isn't needed.

tomdee:
I've tested that -v=10 gives me more logs, and that -log_backtrace_at produces a backtrace. Also checked that the usage text looks right.

PR #776, opened 2017-07-14
Remove RevokeLease from manager interface

tomdee:


gunjan5:
@tomdee looks mostly good code wise, my only concern is what happens to the subnet entry in etcd if someone upgrades flannel/

tomdee:
@gunjan5 I don't follow. How does this code change affect the subnet entry in etcd?

tomdee:
Discussed offline and @gunjan5 says SGTM.

PR #777, opened 2017-07-14
main.go: Remove sleep during shutdown

tomdee:


main.go
```diff
@@ -280,42 +299,44 @@ func main() {
...
 
 	// Kube subnet mgr doesn't lease the subnet for this node - it just uses the podCidr that's already assigned.
-	if opts.kubeSubnetMgr {
-		// Wait for the shutdown to be signalled
-		<-ctx.Done()
-	} else {
-		// Block waiting to renew the lease
-		_ = MonitorLease(ctx, sm, bn)
+	if !opts.kubeSubnetMgr {
+		err = MonitorLease(ctx, sm, bn, wg)
```
gunjan5: shouldn't you be passing `&wg` to `MonitorLease` if you're passing the reference?

main.go
```diff
@@ -339,10 +360,16 @@ func getConfig(ctx context.Context, sm subnet.Manager) (*subnet.Config, error) {
 	}
 }
 
-func MonitorLease(ctx context.Context, sm subnet.Manager, bn backend.Network) error {
+func MonitorLease(ctx context.Context, sm subnet.Manager, bn backend.Network, wg sync.WaitGroup) error {
```
gunjan5: same here, I think it should be `wg *sync.WaitGroup`

tomdee:
I amended the commits with the markups (since they were small). PTAL @gunjan5

PR #781, opened 2017-07-17
use init container to install cni on flannel daemonset

rickypai:
it's odd having a container that only does copy and sleep. it made sense when there's no support for initializing pods, but since kubernetes 1.6, init containers became GA, so the flannel daemonset should be able to leverage this to make the pod more clean.

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -53,6 +53,20 @@ spec:
         operator: Exists
         effect: NoSchedule
       serviceAccountName: flannel
+      initContainers:
+      - name: install-cni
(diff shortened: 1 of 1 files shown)


tomdee:
Hi @rickypai thanks for the contribution! I'd like to merge this but I think that the documentation should get an update to say that flannel now requires K8s 1.6 because of this

rickypai:
thanks for merging this. sorry i sort of forgot about this PR. I can update https://github.com/coreos/flannel/blob/master/Documentation/Kubernetes.md to indicate installation requirements

tomdee:
I added some information in #826 but more would of course be welcomed.

PR #782, opened 2017-07-20
Get Flannel to build on windows and add placeholders

rakelkar:
## Description
As a first step towards adding windows support in Flannel, get it to build!
Also add placeholders for where we need Windows implementations
Commented out all backends except vxlan and hostgw

## Todos
- [ ] Need to run Tests to see if it works (fails) as expected


rakelkar:
Aargh - did not mean to raise a PR on coreos just yet!

PR #786, opened 2017-07-31
backend/udp: Remove the udp backend

tomdee:
Means that CGO can be disabled and we can have static flannel binaries


s-urbaniak:
Question: Although inefficient and prohibiting static binaries, the UDP backend turned out to be an escape hatch in some environments, where VXLAN or other methods of network encapsulation are unsupported/broken/restricted. It is a bit of a pity that this is simply removed. Can this backend at least still live in a separate build/repo?

/cc @crawford @squeed @sym3tri

sym3tri:
@s-urbaniak yes that was a handy escape hatch to have.

@aaronlevy do you have an opinion on this?

aaronlevy:
I also agree that it is somewhat of a universal baseline that we've used in the past. Feel like this might need to be discussed maybe in mailing list? I don't have a good sense of how many users rely on this.

tomdee:
Thanks the feedback. I definitely won't merge this PR without further discussion.

PR #787, opened 2017-07-31
backend/vxlan: Add support for "direct routing"

tomdee:
This skips vxlan encapsulation if the hosts are on the same subnet

### DIFF
--- a/backend/vxlan/vxlan_network.go
+++ b/backend/vxlan/vxlan_network.go
@@ -85,78 +85,112 @@ type vxlanLeaseAttrs struct {
 
 func (nw *network) handleSubnetEvents(batch []subnet.Event) {
 	for _, event := range batch {
-		if event.Lease.Attrs.BackendType != "vxlan" {
-			log.Warningf("ignoring non-vxlan subnet(%s): type=%v", event.Lease.Subnet, event.Lease.Attrs.BackendType)
+		sn := event.Lease.Subnet
+		attrs := event.Lease.Attrs
+		if attrs.BackendType != "vxlan" {
+			log.Warningf("ignoring non-vxlan subnet(%s): type=%v", sn, attrs.BackendType)
 			continue
 		}
 
-		var attrs vxlanLeaseAttrs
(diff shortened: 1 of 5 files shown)


backend/vxlan/vxlan_network.go
```diff
@@ -85,78 +85,112 @@ type vxlanLeaseAttrs struct {
...
+			Gw:  attrs.PublicIP.ToIP(),
+		}
+		var directRoutingOK = false
+		if nw.dev.directRouting {
+			routes, err := netlink.RouteGet(attrs.PublicIP.ToIP())
+			if err != nil {
+				log.Errorf("Couldn't lookup route to %v: %v", attrs.PublicIP, err)
 				continue
 			}
+			if len(routes) == 1 && routes[0].Gw == nil {
```
gunjan5: how would this condition set `directRoutingOK` to `true` for `subnet.EventRemoved` event? maybe I'm missing something, but when we get the `EventAdded` we program the `directRoute` which has `Gw:  attrs.PublicIP.ToIP(),` then we get `EventRemoved` for the same one and if we check this condition where we see if `Gw==nil`, so the clean up in L173 won't happen?

tomdee: Isn't that what I'm doing with my `.Gw == nil` test? 

I could also check to see if the `FLAG_ONLINK` flag is set on the `flags` attribute if you think that would be better?

tomdee: The directRoutingOK check is seeing if there's a direct route to the public IP address of other host. If there is, then we add a route to the IP range for that host using the public IP address as the gateway. The "route" to the public IP address for that host is unchanged; it's still directly connected. 

So when the remove comes in, the public IP is still directly connected and we can remove the route to the network which uses the public IP as the gateway. Make sense?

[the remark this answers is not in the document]
gunjan5: that would be nice but not necessary since your comments are pretty clear

gunjan5: yeah that makes sense, I misunderstood the direct link checking logic and thought directRoutingOk flag will stay false for delete, but it doesn't, so all good now :)

PR #791, opened 2017-08-04
add ca-certificates

oilbeater:
fix: add ca-certificates

Backend like aws-vpc or ali-vpc call https IaaS API will failed without ca-certificates.


tomdee:
This is only going to work for the AMD64 images, but I assume that's all that works on AWS/ALI so I think it's OK to merge. I'm just slightly nervous about having these differences on the different platforms

PR #800, opened 2017-08-27
Add flag validation for subnet-lease-renew-margin

ruoshan:
when the subnet-lease-renew-margin is large than 1440(minutes), etcdv2
subnet manager would keep renew lease, causing high CPU load to flanneld
and etcd.


tomdee:
Looks good, thanks @ruoshan

Lincoln-audi:
this flag is mean  "subnet-lease-renew-margin " key in etcd ttl max 1440 min?

PR #803, opened 2017-08-31
Fix route deletion when replacing route in hostgw backend

### LINKED ISSUE #801: hostgw backend fails to replace old route table entries
In our cluster, flannel is failing to replace existing routes in the route table with new routes. Here's a log of the failure message

```
{"log":"I0829 17:00:21.967998       1 network.go:83] Subnet added: 10.32.10.0/24 via 10.68.29.72\n","stream":"stderr","time":"2017-08-29T17:00:21.968055987Z"}
{"log":"W0829 17:00:21.968144       1 network.go:106] Replacing existing route to 10.32.10.0/24 via 10.68.26.131 with 10.32.10.0/24 via 10.68.29.72.\n","stream":"stderr","time":"2017-08-29T17:00:21.968211104Z"}
{"log":"E0829 17:00:21.968207       1 network.go:108] Error deleting route to 10.32.10.0/24: no such process\n","stream":"stderr","time":"2017-08-29T17:00:21.96826321Z"}
```

Basically for some reason when flannel sends a message to the netlink socket asking the kernel to update the route table, the kernel returns a "no such process" error. I'm able to delete routes with `sudo ip route delete`, and when I delete the routes manually, Flannel is able to create the new routes correctly.

As a result of this the route table ends up being pretty badly misconfigured and what happens is that no packets can be sent through the cni0 bridge (all network connections from a container fail with `no route to host`).

## Expected Behavior

The route table should get updated when the subnet -> IP address mapping changes

## Current Behavior

New routes are added, but any update that requires a deletion fails.

Terminate nodes in our Kubernetes cluster and bring up new nodes (with new IP addresses)

## Your Environment

* Flannel version: 0.7.1 (none of the hostgw code has changed since then, though)
* Backend used: hostgw
* Kubernetes version (if used): 1.7.3
* Operating System and version: Ubuntu 16.04
* Link to your project (optional):

julia-stripe:
## Description

Bug fix for the hostgw backend.

We noticed (#801) that the Flannel was not replacing routes when new nodes were added to the cluster: we saw these error messages in our production cluster

```
[network.go:83] Subnet added: 10.32.10.0/24 via 10.68.29.72\n","stream":"stderr","time":"2017-08-29T17:00:21.968055987Z"}
[network.go:106] Replacing existing route to 10.32.10.0/24 via 10.68.26.131 with 10.32.10.0/24 via 10.68.29.72.\n","stream":"stderr","time":"2017-08-29T17:00:21.968211104Z"}
[network.go:108] Error deleting route to 10.32.10.0/24: no such process\n","stream":"stderr","time":"2017-08-29T17:00:21.96826321Z"}
```

This was happening when an existing route was being replaced with a new route.

This turns out to be for 2 reasons:

1. The wrong LinkIndex was being set on the netlink messages: the LinkIndex was always set to 0, when it should be set to the index of the external network interface. (`extIface.Iface.Index`)
2. Instead of deleting the **old** route, this code path was trying to delete the **new** route. Since the new route doesn't exist, the deletion fails. 

I tested Flannel with this patch in our cluster and now replacing routes works correctly.

Fixes #801.

### DIFF
--- a/backend/hostgw/hostgw_network.go
+++ b/backend/hostgw/hostgw_network.go
@@ -29,12 +29,11 @@ import (
 )
 
 type network struct {
-	name      string
-	extIface  *backend.ExternalInterface
-	linkIndex int
-	rl        []netlink.Route
-	lease     *subnet.Lease
-	sm        subnet.Manager
+	name     string
+	extIface *backend.ExternalInterface
+	rl       []netlink.Route
+	lease    *subnet.Lease
+	sm       subnet.Manager
 }
 
 func (n *network) Lease() *subnet.Lease {
@@ -43,6 +42,10 @@ func (n *network) Lease() *subnet.Lease {
 
 func (n *network) MTU() int {
 	return n.extIface.Iface.MTU
+}
+
+func (n *network) LinkIndex() int {
+	return n.extIface.Iface.Index
 }
 
 func (n *network) Run(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


tomdee:
Thanks for the fix @julia-stripe . it looks great to me so merging.

PR #816, opened 2017-09-18
Improve functional tests

tomdee:
- Move to using bash_unit
- Test pings in both directions
- Test kubernetes subnet manager
- Test multiple versions of etcd and k8s during a release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,18 @@ dist/flanneld: $(shell find . -type f  -name '*.go')
 	  -ldflags "-s -w -X github.com/coreos/flannel/version.Version=$(TAG)"
 
 test: license-check gofmt
+	# Run the unit tests
 	go test -cover $(TEST_PACKAGES_EXPANDED)
+
+	# Test the docker-opts script
 	cd dist; ./mk-docker-opts_tests.sh
 
-e2e-test: dist/flanneld-$(TAG)-$(ARCH).docker
-	cd dist; ./functional-test.sh $(REGISTRY):$(TAG)-$(ARCH)
+	# Run the functional tests
+	make e2e-test
+
+e2e-test: bash_unit dist/flanneld-$(TAG)-$(ARCH).docker
+	./bash_unit dist/functional-test.sh
+	./bash_unit dist/functional-test-k8s.sh
 
 cover:
 	# A single package must be given - e.g. 'PACKAGES=pkg/ip make cover'
@@ -73,6 +80,10 @@ update-glide:
 	glide update --strip-vendor
 	# go get -d -u github.com/sgotti/glide-vc
 	glide vc --only-code --no-tests
+
+bash_unit:
+	wget https://raw.githubusercontent.com/pgrange/bash_unit/v1.6.0/bash_unit
+	chmod +x bash_unit
 
 clean:
 	rm -f dist/flanneld*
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -178,6 +188,23 @@ release: tar.gz
 	@echo "Add all *.aci, flanneld-* and *.tar.gz files from dist/ to the Github release"
 	@echo "Use make docker-push-all to push the images to a registry"
 
+release-tests: bash_unit
+	# Run the functional tests with different etcd versions. Currently these are the latest point releases.
+	# This list should be updated during the release process.
+	ETCD_IMG="quay.io/coreos/etcd:v3.2.7"  ./bash_unit dist/functional-test.sh
```
gunjan5: what if we do a `latest` tag here instead?

dist/functional-test-k8s.sh
```diff
@@ -0,0 +1,117 @@
...
+       declare -g ping_dest$host_num=$(docker "exec" --privileged flannel-e2e-test-flannel$host_num /bin/sh -c '\
+		source /run/flannel/subnet.env && \
+		ip link add name dummy0 type dummy && \
+		ip addr add $FLANNEL_SUBNET dev dummy0 && ip link set dummy0 up && \
+		echo $FLANNEL_SUBNET | cut -f 1 -d "/" ')
+    done
+}
+
+test_vxlan() {
+    start_flannel vxlan && sleep 2
```
gunjan5: how about instead of all these `sleep 2`, we put the `pings` in a loop so it retries for 5-10 times to see if pings work, so the tests don't become flakey

dist/functional-test-k8s.sh
```diff
@@ -0,0 +1,117 @@
...
+
+test_host-gw() {
+    start_flannel host-gw && sleep 2
+    create_ping_dest # creates ping_dest1 and ping_dest2 variables
+    pings
+}
+
+pings() {
+    # ping in both directions
+	assert "docker exec -it --privileged flannel-e2e-test-flannel1 /bin/ping -c 5 $ping_dest2" "Host 1 cannot ping host 2"
```
gunjan5: or we could put this in some sort of retry block, maybe

dist/functional-test-k8s.sh
```diff
@@ -0,0 +1,117 @@
...
+	assert "docker exec -it --privileged flannel-e2e-test-flannel2 /bin/ping -c 5 $ping_dest1" "Host 2 cannot ping host 1"
+}
+
+test_manifest() {
+    # This just tests that the API server accepts the manifest, not that it actually acts on it correctly.
+    assert "cat ../Documentation/kube-flannel.yml |  docker run -i --rm --net=host gcr.io/google_containers/hyperkube-amd64:v$K8S_VERSION /hyperkube kubectl create -f -"
```
gunjan5: btw, there's a cool project `garethr/kubeval` that can validate the k8s manifests. Not suggesting we should or shouldn't use it, but just throwing it out there

dist/functional-test.sh
```diff
@@ -1,154 +1,144 @@
...
 
-	docker exec -it --privileged flannel-e2e-test-flannel2 /bin/ping -c 5 $ping_dest
-	exit_code=$?
-
-	# Uncomment to debug (you can nsenter)
-	#if [ $exit_code -eq "1" ]; then
-	#	sleep 10000
-	#fi
+test_vxlan_ping() {
+    write_config_etcd vxlan && sleep 1
```
gunjan5: curious, why is this `sleep 1` vs 2 in the other test?

gunjan5: i'd still advocate for some sort of ping retry instead of sleep

Makefile
tomdee: good idea

dist/functional-test-k8s.sh
tomdee: the sleep is actually there to make sure the subnet.env file is written. I'll update the PR to get rid of the sleeps

PR #840, opened 2017-10-16
Fix #712, allow overwriting the public IP of a Kubernetes node

### LINKED ISSUE #712: Flannel always overwrites public-ip node annotation in Kubernetes on startup
This is an issue when running flannel inside kubernetes as a daemonset on VM's with differing internal and external IP addresses (openstack).  There is no way to use the --public-ip argument to different hosts within a daemonset.   The only other way to vary public-ip on a node is to use the node annotation (flannel.alpha.coreos.com/public-ip).

The problem is flannel overwrites this annotation _every time_ it starts up with the internal IP of the host.  This, with the inability to use --public-ip variations in daemonsets, means there is no way to set a different public-ip to be used with flannel in kubernetes.

I have a fix in mind (see below).

## Expected Behavior
flannel.alpha.coreos.com/public-ip can be set manually and is not overwritten.

## Current Behavior
flannel.alpha.coreos.com/public-ip is overwritten every time flannel starts regardless of it's previous state.

## Possible Solution

Simply commenting this line fixes the problem on a node where flannel has been run before.
https://github.com/coreos/flannel/blob/master/subnet/kube/kube.go#L218

This allows us to set a custom public-ip and flannel will read it on startup.

Public-ip still needs to be set if it does not exist, so I propose we check for a public-ip-override and use the assumed IP only if it's not there.

* Start flannel in a kubernetes daemon set
* `kubectl annotate node <node name> flannel.alpha.coreos.com/public-ip=<some ip> --overwrite`
* Kill flannel container or restart it some other way
* Verify your node annotation flannel.alpha.coreos.com/public-ip is now set to the assumed (incorrect) IP.

## Context
See above description.

## Your Environment

* Flannel version: v0.7.0-96 (applicable on master branch)
* Backend used (e.g. vxlan or udp):  vxlan
* Etcd version: 3.0.17
* Kubernetes version (if used): v1.6.2
* Operating System and version: Ubuntu 16.04

alvaroaleman:
## Description

This PR fixes #712 by allowing to overwrite the public ip of  a Kubernetes node via the `flannel.alpha.coreos.com/public-ip-overwrite` annotation.

Can anyone point me to the appropriate place for docs, is it just `Documentation/Configuration.md`?
Right now the other annotations aren't documented in this repo.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/Documentation/kubernetes.md
+++ b/Documentation/kubernetes.md
@@ -14,6 +14,10 @@ The `flannel` manifest defines four things:
 
 When you run pods, they will be allocated IP addresses from the pod network CIDR. No matter which node those pods end up on, they will be able to communicate with each other.
 
+# Annotations
+
+*  `flannel.alpha.coreos.com/public-ip-overwrite`: Allows to overwrite the public IP of a node. Useful if the public IP can not determined from the node, e.G. because it is behind a NAT
+
 ## Older versions of Kubernetes
 
 `kube-flannel.yaml` has some features that aren't compatible with older versions of Kubernetes, though flanneld itself should work with any version of Kubernetes.
--- a/dist/functional-test-k8s.sh
+++ b/dist/functional-test-k8s.sh
@@ -82,15 +82,15 @@ start_flannel() {
        docker exec flannel-e2e-test-flannel$host_num /bin/sh -c 'mkdir -p /etc/kube-flannel'
        echo $flannel_conf | docker exec -i flannel-e2e-test-flannel$host_num /bin/sh -c 'cat > /etc/kube-flannel/net-conf.json'
        docker exec -d flannel-e2e-test-flannel$host_num /opt/bin/flanneld --kube-subnet-mgr --kube-api-url $k8s_endpt
+       while ! docker exec flannel-e2e-test-flannel$host_num ls /run/flannel/subnet.env >/dev/null 2>&1; do
+         sleep 0.1
+       done
     done
 }
 
 create_ping_dest() {
     # add a dummy interface with $FLANNEL_SUBNET so we have a known working IP to ping
     for host_num in 1 2; do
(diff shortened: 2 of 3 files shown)


tomdee:
@alvaroaleman Thanks for this change. If you can add some documentation and ideally a test, I can get it merged. For the documentation, I would suggest adding an "Annotations" section to https://github.com/coreos/flannel/blob/master/Documentation/kubernetes.md

alvaroaleman:
Added docs. Writing a test seems to be rather complicated, because there are none for the kube subnet manager yet.

tomdee:
I added some tests for kube subnet manager recently - https://github.com/coreos/flannel/blob/master/dist/functional-test-k8s.sh

But I agree, they are still quite complicated!

alvaroaleman:
Thanks for the hint @tomdee, I added a test.

I also moved the check for `ls /run/flannel/subnet.env` into the `start_flannel` function, since it is a generic test if flannel startup was finished and not specific to the other ping tests in the suite.

alvaroaleman:
Recreated the PR because travis was failing after having issues pulling the etcd image.

The e2e tests run twice btw, because they are called in the travis file and from within the `test` maketarget

tomdee:
OK, this looks good. It would be great if you could add a little more information to the documentation, such as _why_ a user might want to set this, and any other annotations that can be set. 

Before I merge can you squash your commits so I'm just merging a single commit?

alvaroaleman:
Updated the explanation and squashed my commits.

As far as I can tell, the other annotations are not supposed to be altered by users thus it doesn't make sense to document them or am I wrong there?

tomdee:
Yes, I think you're right! The tests are now failing, I think you lost a commit when doing the squash? Hopefully you can get it back easily then I can merge.

alvaroaleman:
Added the missing commit back.

PR #842, opened 2017-10-18
Add new ipip backend

chenchun:
Signed-off-by: forrestchen <forrestchen@tencent.com>
Signed-off-by: Chun Chen <ramichen@tencent.com>

@ChenLingPeng


backend/ipip/ipip.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/coreos/flannel/pkg/ip"
+	"github.com/coreos/flannel/subnet"
+	glog "github.com/golang/glog"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/net/context"
+)
+
+const (
+	backendType  = "ipip"
+	tunnelName   = "tunl0"
```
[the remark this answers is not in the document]
chenchun: Instead of using tunl0 device, creating a new `flanneld.ipip` ipip device. As a effect of this we should set `flanneld.ipip`‘s local ip to public ip, or kernel will forward packets to tunl0.

backend/ipip/ipip.go
```diff
@@ -0,0 +1,180 @@
...
+	backend.Register(backendType, New)
+}
+
+type IPIPBackend struct {
+	sm       subnet.Manager
+	extIface *backend.ExternalInterface
+}
+
+func New(sm subnet.Manager, extIface *backend.ExternalInterface) (backend.Backend, error) {
+	if !extIface.ExtAddr.Equal(extIface.IfaceAddr) {
```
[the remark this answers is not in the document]
chenchun: I don't have an environment to test this and currently removed this check.

backend/ipip/ipip.go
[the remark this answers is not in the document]
tomdee: I don't understand your comment about setting the IP address for flanneld.ipip to the public. Could you explain some more?

Documentation/backends.md
```diff
@@ -115,4 +115,5 @@ IPIP kind of tunnels is the simplest one. It has the lowest overhead, but can in
 
 Type:
 * `Type` (string): `ipip`
+* `MTU` (number): The mtu of ipip device. Defaults to 1480.
```
tomdee: Can you write a little more here to help guide users. Why is this this value? Why and when would a user want to change it? And what value would they change it to?

dist/functional-test.sh
```diff
@@ -110,6 +116,12 @@ test_udp_perf() {
     perf
 }
 
+test_ipip_perf() {
+    write_config_etcd ipip
+	create_ping_dest
```
tomdee: bad indent

tomdee:
I added a few more comments (including this one which might be hidden https://github.com/coreos/flannel/pull/842#discussion_r147539848)

backend/ipip/ipip.go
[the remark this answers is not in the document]
chenchun: tunl0 is an namespace default device with IPIP attributes local=any and remote=any.
If we create another IPIP device flannel.ipip with local=any and remote=any. It seems when receiving IPIP protocol packets, kernel will forward them to tunl0 as a fallback device instead of flannel.ipip. See https://github.com/torvalds/linux/blob/master/net/ipv4/ip_tunnel.c#L85-L96 .

So we have two options here, either rename tunl0 to flannel.ipip or setting local attribute of flannel.ipip to distinguish these two devices. Considering tunl0 might be used by user, so I choose the later option.

Documentation/backends.md
chenchun: Fixed

dist/functional-test.sh
chenchun: Fixed

chenchun:
@tomdee PTAL

backend/ipip/ipip.go
```diff
@@ -0,0 +1,176 @@
+// Copyright 2015 flannel authors
```
tomdee: Can you update the copyright lines to 2017 please.

tomdee:
Hi @chenchun I want to make sure I understand the design for this fully before merging it. If I understand your comment (https://github.com/coreos/flannel/pull/842#discussion_r147542937) and the code correctly then you're not setting a "remote" but you are setting the "local" address to the public IP. Can you confirm what effects this has? Which address will be used as the source IP address for packets that originate from the host network namespace? And what if the public IP isn't the same as an IP address on the host, then you can't actually use it?

I also have some questions about the MTU settings. On the vxlan backend, the vxlan tunnel device is associated with a physical ethernet device on the host. This is what is used to determine the MTU value to use. Flanneld will always have an interface that it chooses (specified with the `--iface` option), so maybe you should be getting the MTU value from that device? WDYT?

So maybe the best option would be to "local" address to the address of the interface that flannel selects and also use that as the for the MTU?

tomdee:
When I run your code I'm seeing two IPIP devices, is that expected?

```
6: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
7: flannel.ipip@NONE: <NOARP,UP,LOWER_UP> mtu 1480 qdisc noqueue state UNKNOWN mode DEFAULT group default 
    link/ipip 192.168.5.188 brd 0.0.0.0
```

tomdee:
Also, one final request: Would you mind adding some comments to your code explain _why_ it's written like it is. For some examples of the style, see the vxlan code - e.g. https://github.com/coreos/flannel/blob/master/backend/vxlan/device.go#L108

chenchun:
> If I understand your comment (#842 (comment)) and the code correctly then you're not setting a "remote" but you are setting the "local" address to the public IP. Can you confirm what effects this has? Which address will be used as the source IP address for packets that originate from the host network namespace? And what if the public IP isn't the same as an IP address on the host, then you can't actually use it?

My mistake. I originally had a check if PublicIP is the same as IfaceAddr and I assumed they are equal. Now that I removed this check, I will change the "local" address to the IfaceAddr.

```
if !extIface.ExtAddr.Equal(extIface.IfaceAddr) {
		return nil, fmt.Errorf("your PublicIP differs from interface IP, meaning that probably you're on a NAT, which is not supported by host-gw backend")
	}
```
> maybe you should be getting the MTU value from that device? WDYT?

I agree.

> When I run your code I'm seeing two IPIP devices, is that expected?

Yes, this is expected. tunl0 is created by the ipip kernel module itself and it is not used by this ipip backend.

chenchun:
The last update removed MTU param from backend configuration. We may not need it since we calculate it based on `--iface` device's MTU.

backend/ipip/ipip.go
```diff
@@ -0,0 +1,182 @@
...
+	if link.Type() != "ipip" {
+		return nil, fmt.Errorf("%v isn't an ipip mode device", tunnelName)
+	}
+	ipip, ok := link.(*netlink.Iptun)
+	if !ok {
+		return nil, fmt.Errorf("%s isn't an iptun device %#v", tunnelName, link)
+	}
+
+	// Don't set remote attribute making flannel.ipip an one to many tunnel device.
+	if (ipip.Local != nil && !ipip.Local.Equal(be.extIface.IfaceAddr)) || (ipip.Remote != nil && ipip.Remote.String() != "0.0.0.0") {
```
tomdee: I don't think that just returning an error here is good enough. If there's a `flannel.ipip` that isn't an ipip device then it wasn't created by flannel, so in that case it's best to just return an error. But in this case, the local address can change between flannel restarts, so flannel should be able to recreate the device and not just return an error. You should follow similar logic to vxlan (see the ensureLink function), it should delete and recreate the device if the config changes

backend/ipip/ipip.go
```diff
@@ -0,0 +1,182 @@
...
+	// Due to the extra 20 byte IP header that the tunnel will add to each packet,
+	// MTU size for both the workload and tunnel interfaces should be 20 bytes less than the selected iface (specified with the --iface option).
+	expectMTU := be.extIface.Iface.MTU - 20
+	if expectMTU <= 0 {
+		return nil, fmt.Errorf("MTU %d of iface %s is too small for ipip mode to work", be.extIface.Iface.MTU, be.extIface.Iface.Name)
+	}
+	oldMTU := link.Attrs().MTU
+	if oldMTU > expectMTU || oldMTU == 0 {
+		glog.Infof("current MTU of %s is %d, setting it to %d", tunnelName, oldMTU, expectMTU)
+		err := netlink.LinkSetMTU(link, expectMTU)
```
tomdee: Does this need to explicitly set the link to down first?

backend/ipip/ipip.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		// change MTU as it will be written into /run/flannel/subnet.env
+		link.Attrs().MTU = expectMTU
+	}
+
+	if err := netlink.LinkSetUp(link); err != nil {
+		return nil, fmt.Errorf("failed to set %v UP: %v", tunnelName, err)
+	}
+
+	existingAddrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
```
tomdee: I think this next section of code is identical to the vxlan code, could you extract the code into a common package?

backend/ipip/ipip_network.go
```diff
@@ -0,0 +1,216 @@
+// Copyright 2017 flannel authors
```
tomdee: Much of this file is the same as the host-gw code. It would be great if the common code could be pulled out (and then it could be shared by the vxlan code too but I would do that in a followup PR).

tomdee:
Hi @chenchun I've looked at this code some more and realized how similar so much of is it to existing flannel code. It would be great to refactor and pull out the common code.

A few other things that need to happen before this can be released (but not necessarily before this PR is merged)
* Documentation update (maybe in the troubleshooting guide too) to descrieb that the tunl0 device will be created but that's not a bug (!) - users will get confused about this.
* a ..._windows.go stub needs to be created.
* Some manual functional and scale testing.

backend/ipip/ipip.go
chenchun: I couldn't find any reference about  this. I think changing MTU on the fly is ok. 

From http://linux-ip.net/html/tools-ip-link.html 
> # B.3.5. Using ip link set to change the MTU
> The remaining options to the ip link command cannot be used while the interface is in an UP state.

backend/ipip/ipip.go
chenchun: Fixed

backend/ipip/ipip.go
chenchun: Fixed

backend/ipip/ipip_network.go
chenchun: I'll try to do it

chenchun: Added RouteNetwork in common.go.

tomdee:
@chenchun and @ChenLingPeng I rebased and squashed your commits and added another commit to the PR just to make some small markups. Your original commits are at https://github.com/tomdee/flannel/tree/iptun-orig if you still want them.

chenchun:
Ausome! Thanks for making this happen.

PR #843, opened 2017-10-18
Remove unexpected route from route list

chenchun:


tomdee:
thanks @chenchun this looks like a good fix. Before I merge it, can you share a little bit of information on how you hit this and how you've tested the fix. Ideally there would be an automated test case for this too, would that be possible to add?

chenchun:
Found this bug through reading the code. I've added an unit test. PTAL

tomdee:
Looks good thanks. You'll need to update the Makefile too though, the  TEST_PACKAGES variable needs to include backend/hostgw

chenchun:
Previous tests fails with
```
--- FAIL: TestRouteCache (0.00s)
	hostgw_network_test.go:43: operation not permitted
```

Updated Makefile to makes use of `docker run` to give go test NET_ADMIN capability.

tomdee:
This does mean that Docker is needed for running the unit tests, but I think that's OK now since the end to end tests need Docker too.

PR #845, opened 2017-10-19
Add support for multiarch e2e tests

mkumatag:
## Description
This PR is for enabling e2e tests to run on different architectures like arm64 and ppc64le

Can be run on ppc64le platform like below:

```shell
$ make e2e-test ARCH=ppc64le
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,23 +15,18 @@ PACKAGES_EXPANDED=$(PACKAGES:%=github.com/coreos/flannel/%)
 
 # Set the (cross) compiler to use for different architectures
 ifeq ($(ARCH),amd64)
-	LIB_DIR=/lib/x86_64-linux-gnu
 	CC=gcc
 endif
 ifeq ($(ARCH),arm)
-	LIB_DIR=/usr/arm-linux-gnueabihf/lib
 	CC=arm-linux-gnueabihf-gcc
 endif
 ifeq ($(ARCH),arm64)
-	LIB_DIR=/usr/aarch64-linux-gnu/lib
 	CC=aarch64-linux-gnu-gcc
 endif
 ifeq ($(ARCH),ppc64le)
-	LIB_DIR=/usr/powerpc64le-linux-gnu/lib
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -55,9 +50,10 @@ test: license-check gofmt
 	# Run the functional tests
 	make e2e-test
 
-e2e-test: bash_unit dist/flanneld-$(TAG)-$(ARCH).docker
-	./bash_unit dist/functional-test.sh
-	./bash_unit dist/functional-test-k8s.sh
+e2e-test: bash_unit dist/flanneld-e2e-$(TAG)-$(ARCH).docker
```
tomdee: Shouldn't this depend on `dist/$(REGISTRY):$(TAG)-$(ARCH).docker`

Makefile
```diff
@@ -101,6 +97,19 @@ ifeq ($(ARCH),amd64)
 	docker build -f Dockerfile.$(ARCH) -t $(REGISTRY):$(TAG) .
 endif
 
+# This will build flannel natively using golang image
+dist/flanneld-e2e-$(TAG)-$(ARCH).docker:
+	# valid values for ARCH are [amd64 arm arm64 ppc64le s390x]
+	docker run -e GOARM=$(GOARM) \
```
tomdee: I'm not sure I understand why you've added this? What was wrong with the ` dist/flanneld-$(ARCH)` target?

Makefile
mkumatag: The difference between these 2 targets are:
dist/flanneld-$(TAG)-$(ARCH).docker uses `gcr.io/google_containers/kube-cross` which is not a multiarch image and can be run only on `amd64` platform. And if I run this docker image on other than amd64 platform I'll hit exec format error.

and dist/flanneld-e2e-$(TAG)-$(ARCH).docker target uses `golang:1.8.3` docker image which is multiarch image which will load the respective docker image while running on different architecture.

tomdee: Aha, that all makes sense.

PR #850, opened 2017-10-23
Fix TTY error while running e2e tests in container

mkumatag:
## Description
While running these tests in docker containers I was facing following error:

```
the input device is not a TTY
```
Usually container will not have a TTY and in many places we are trying to run with tty which is not really required, so this PR is submitted for removing such tty option from docker commands.


tomdee:
Looks good, thanks

mkumatag:
Thanks @tomdee

Issue #833, opened 2017-10-04
Support Windows in host-gw and VXLAN modes


rakelkar:
Windows RS3 release introduces support for Kubernetes networking, however setting it up without a Flannel like component is very difficult.The Windows Server RS3 release supports host-gw as L2Bridge and vxlan in Overlay modes - this issue proposes supporting Flannel on windows for host-gw and vxlan modes and links to a proposed PR.

See here for details on new Windows support for K8s networking: http://blog.kubernetes.io/2017/09/windows-networking-at-parity-with-linux.html

## Possible Solution
Proposed PR: https://github.com/coreos/flannel/pull/832
Related CNI changes: https://github.com/containernetworking/plugins/pull/76

## Your Environment
* Operating System and version: Windows RS3

rakelkar:
containernetworking/plugins#77

rakelkar:
containernetworking/plugins#76

cspwizard:
Hi!

Are there any plans when this would be available for use on Windows K8S systems?

Thanks!

rakelkar:
Yes we are working towards making this available... Would you like to try
early bits to see if it works for you?

rjmorse:
Count me in as a tester of early bits as well please.

cspwizard:
@rakelkar yes, sure!

rakelkar:
@cspwizard @rjmorse 
I've posted some slides that describe how flannel works on windows and how to set it up on the k8s sig-windows slack channel... have a look and give it a shot. We're working with the Flannel maintainers to get this merged step by step :) currently stuck at #891

dmitryzykov:
Hello @rakelkar
I've tried to test host-gw and VXLAN modes for Windows 1709 nodes, but unfortunately, it doesn't work at the end. 
What I did after reading slack talks, watching your demo video and presentation from slack:

ran kubernetes v1.9.0 cluster with flannel in VXLAN and then in host-gw mode.
compiled all required windows binaries from: https://github.com/containernetworking/plugins/pull/85 and https://github.com/coreos/flannel/pull/832 and v1.9.0 kubernetes repo
```
├── admin.conf
├── cni
│   ├── config
│   │   └── cni-conf.json
│   ├── flannel.exe
│   ├── host-local.exe
│   ├── win-l2bridge.exe
│   └── win-overlay.exe
├── etc
│   └── kube-flannel
│       └── net-conf.json
├── flanneld.exe
├── kubectl.exe
[...]

PR #854, opened 2017-10-25
network: move ipmasq management into platform-specific files

jroggeman:
## Description
As requested in #832 , we're breaking up the changes into smaller commits.  This PR moves the Setup, monitoring, and teardown of IPMasq settings into platform-specific files.  This is one of the requirements to get Flannel to build on Windows.

From the original PR (#832):

> The goal is to enable users to provision a Kubernetes cluster with windows nodes in VXLAN mode or mixed windows and Linux nodes in host-gw mode.

### DIFF
--- a/network/ipmasq.go
+++ b/network/ipmasq.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/coreos/flannel/pkg/ip"
 	"github.com/coreos/flannel/subnet"
+	"github.com/coreos/go-iptables/iptables"
+	"time"
 )
 
 type IPTablesRules interface {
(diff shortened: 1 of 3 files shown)


jroggeman:
@tomdee This addresses your comment from #832

tomdee:
The tests are failing. I restarted it as it looks like a google connectivity problem.

network/ipmasq.go
```diff
@@ -1,3 +1,5 @@
+// +build linux
```
tomdee: What about just naming the file _linux.go?

rakelkar: In our next PR we are going to touch a bunch of files to get Flannel to compile on windows. If we follow this pattern all those files would get renamed to xxx_linux.go that feels very ugly and intrusive.... Can make the change if you think that it is more correct/idiomatic though.

rakelkar: Also do you care about Nix's other than linux? Just wondering whether the _unspecified file is required in this case? For the larger change we've kept _unspecified for features that are not supported outside linux e.g. awsvpc_unspecfied, see in original PR:  https://github.com/coreos/flannel/pull/832/files#diff-c4be229b15064d7e338159bb6ef8cddf

tomdee:
Since there is no actual Windows specific code in here, why not use this PR to just move iptables stuff out of main.go but not do any file renaming?

jroggeman:
@tomdee updated review to only be the minor refactoring, can you take a look?

tomdee:
Looks good. If you squash the commits I can merge.

jroggeman:
Squashed changes are pushed :)

tomdee:
Looks good, merging.

rakelkar:
#833

PR #860, opened 2017-10-27
Makefile: Cleanup, alpine and amd64 only UDP

tomdee:
Reorganize the Makefile to have different sections for build, test,
release and development.

Move to using Alpine linux as the base image. This means we no longer
need to build iptables and it makes it easier to install other packages
in future (e.g. strongswan).

To allow alpine to be used, static binaries for all platforms are
needed. It's not possible to create these for the UDP backend for
non-amd64 platforms. This is because it relies on CGO. So this PR makes
the UDP backend work on amd64 only.


Dockerfile.amd64
```diff
@@ -1,12 +1,11 @@
-FROM frolvlad/alpine-glibc
+FROM alpine
 
 MAINTAINER Tom Denham <tom@tigera.io>
```
gunjan5: nit: `MAINTAINER` has been deprecated, we should probably switch to `LABEL`

Makefile
```diff
@@ -6,40 +6,61 @@ REGISTRY?=quay.io/coreos/flannel
...
 ARCH?=amd64
+# Only enable CGO (and build the UDP backend) on AMD64
+ifeq ($(ARCH),amd64)
+	CGO_ENABLED=1
+else
+	CGO_ENABLED=0
+endif
+
+# Go version to use for builds
+GO_VERSION=1.8.3
```
gunjan5: any reason we're not using Go 1.9.2?

backend/udp/udp_amd64.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2015 flannel authors
```
gunjan5: 2017

backend/udp/udp_amd64.go
```diff
@@ -0,0 +1,87 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package udp
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"golang.org/x/net/context"
```
gunjan5: context has been move to `"context"` as part of the std lib since Go 1.7

tomdee: Sure, but it hasn't been updated in flannel yet and that change shouldn't be done in this PR

Makefile
tomdee: I don't want to introduce too many changes in this PR. I'll rev the Go version in a different PR

Dockerfile.amd64
tomdee: Already done in another PR :smile:

PR #862, opened 2017-10-31
network/ipforward: Add support for --ip-forward flag

tomdee:
To work around the Docker change from v1.13 which changed the default
FORWARD policy to DROP.

This new flag simply changes the FORWARD polciy to accept.

The change has bitten many many users. A future PR will enable this flag
in the kube-flannel.yaml.

The troubleshooting documentation is also updated talk about the issue.

The glide.lock has been updated to pick up a new version of go-iptables
to get this PR https://github.com/coreos/go-iptables/pull/43


network/ipforward.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2015 flannel authors
```
gunjan5: nit: 2015?

main.go
```diff
@@ -288,6 +290,11 @@ func main() {
 		go setupIPMasq(config, bn)
 	}
 
+	// Set up ipForward if needed
+	if opts.ipForward {
+		go setupIPForward()
```
gunjan5: why is this a goroutine? Seems like the main thread can end before this goroutine is done since it's not part of the waitgroup that waits at the end of the `main()`. So either we can serialize this or make the main thread wait for it. WDYT? (Same applies to `go setupIPMasq()`)

PR #876, opened 2017-11-13
Fix e2e tests for non-amd64 archs

mkumatag:
This PR will fix following things:

- Copy the qemu binaries for e2e-test target
- Remove redundant make targets
- Disable the udp tests for non-amd64 architectures
- Gitignore `dist/here.txt` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 dist/*.tar.gz
 dist/flanneld*
 dist/*.docker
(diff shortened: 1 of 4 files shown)


mkumatag:
@tomdee

tomdee:
@mkumatag As always, thanks for spotting this stuff and getting it cleaned up.

PR #888, opened 2017-11-25
Add portmap plugin to CNI conf

osoriano:
## Description
The portmap plugin is needed for hostPorts to work

The CNI [portmap capability was added in k8s 1.7](https://github.com/kubernetes/kubernetes/issues/23920#issuecomment-306543120). I was able to get past the issue by using the updated kube-flannel.yml on a 1.8.3 cluster.

Closes: https://github.com/coreos/flannel/issues/887

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -55,11 +55,21 @@ data:
   cni-conf.json: |
     {
       "name": "cbr0",
-      "type": "flannel",
-      "delegate": {
-        "hairpinMode": true,
-        "isDefaultGateway": true
-      }
+      "plugins": [
+        {
+          "type": "flannel",
+          "delegate": {
+            "hairpinMode": true,
+            "isDefaultGateway": true
+          }
+        },
+        {
+          "type": "portmap",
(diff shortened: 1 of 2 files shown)


tomdee:
Looks good

jethrogb:
`10-flannel.conf` sorts before `10-flannel.conflist`, so if you update you configuration with this change, this change will not be picked up since `10-flannel.conf` still exists.

zcalusic:
I ended up with both `10-flannel.conf` & `10-flannel.conflist` with the exactly same contents, and **lots** of `kubelet[29256]: W1227 18:53:27.409233   29256 cni.go:125] Error loading CNI config file /etc/cni/net.d/10-flannel.conf: no 'type'; perhaps this is a .conflist?` log spam in journal. Every few seconds... :(

Admittedly, this was applied to 0.9.1 image so maybe some bits & pieces are still missing, possibly will be available in 0.9.2?

For now, I deleted 10-flannel.conf, it fixes the extreme log spam issue. Hopefully everything continues working ok...

PR #891, opened 2017-11-28
flannel: Get flannel running on windows

jroggeman:
## Get flannel running on windows
As per #832, gets flannel to run on windows (Vxlan and host-gateway to come in separate PRs after this)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist/here.txt
 cover.out
 .editorconfig
 .idea/
+default.etcd/
 flannel.exe
 bash_unit
 dist/qemu-*
--- a/pkg/ip/iface_windows.go
+++ b/pkg/ip/iface_windows.go
@@ -15,22 +15,52 @@
 package ip
 
 import (
-	"errors"
+	netsh "github.com/rakelkar/gonetsh/netsh"
 	"net"
 )
 
 func GetIfaceIP4Addr(iface *net.Interface) (net.IP, error) {
-	return nil, errors.New("GetIfaceIP4Addr not implemented for this platform")
-}
+	// get ip address for the interface
+	// prefer global unicast to link local addresses
+	netHelper := netsh.New(nil)
+	ifaceDetails, err := netHelper.GetInterfaceByName(iface.Name)
+	if err != nil {
+		return nil, err
+	}
 
-func GetIfaceIP4AddrMatch(iface *net.Interface, matchAddr net.IP) error {
-	return errors.New("GetIfaceIP4AddrMatch not implemented for this platform")
+	ifAddr := net.ParseIP(ifaceDetails.IpAddress)
+
+	return ifAddr, nil
 }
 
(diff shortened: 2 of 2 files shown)


rakelkar:
#833

jroggeman:
Two important notes:

1. We didn't install the gonetsh using glide; `glide get` also updates packages, and one of the vendored deps it pulled managed to include a file that can't be checked out on Windows due to the filename :) Not sure how to resolve that sort of issue.
2. This PR shouldn't be merged until appveyor is setup for Windows-based CI.

tomdee:
Which file can't be checked out on windows?

tomdee:
I've enabled AppVeyor :+1: But it's failing :-1: https://ci.appveyor.com/project/tomdee/flannel

jroggeman:
Fixed the CI issues; I'm no super familiar with glide and go vendoring but I think that AppVeyor is failing since we haven't vendored rakesh's package via glide.

The file that was pulled in while vendoring with glide:
`fatal: cannot create directory at 'vendor/github.com/docker/distribution/contrib/docker-integration/generated_certs.d/localregistry:5440': Invalid argument`

This does appear to be a folder on linux, and colons are invalid as folder names on windows.

jroggeman:
@tomdee We're still having issues with properly getting our library vendored.  When we run just a `glide install` to fetch deps, it brings in a huge number of files and dependencies (that aren't checked in to git) into the vendor folder.  Are these supposed to be gitignored?  Are they supposed to be vendored (and thus checked in) but aren't?  We're pretty confused at the moment with the dependency management, so any light you could shed on how dependencies are being managed would be quite helpful :)

tomdee:
@jroggeman See the Makefile for how glide should be run - https://github.com/coreos/flannel/blob/master/Makefile#L191

```
update-glide:
	# go get -d -u github.com/Masterminds/glide
	glide update --strip-vendor
	# go get -d -u github.com/sgotti/glide-vc
	glide vc --only-code --no-tests
```

In theory, running `glide install -v && glide vc --only-code --no-tests` without changing glide.lock should result in no changes to the vendor directory.

Since you're adding a new package and updating glide.yaml that will result in a change to glide.lock (since you'll need to run glide update). This means that some other packages will be updated since they are not all pinned in glide.yaml.

It looks like CI is still failing on Appveyor because of this vendoring issue.

On the `docker/distribution/contrib...` issue, maybe you can update the version of k8s.io/client-go that's being used and that will stop relying on those files...

jroggeman:
Even this doesn't seem to be working too well; even after installing, `glide vc` doesn't work at all, just exits.  Appears that `glide list` fails with versioning issues, not able to find pinned versions for any packages, even _without_ explicitly installing the new package we need. Effectively I can't seem to do the specified steps without changing or adding lots of files to the vendor directory

tomdee:
@jroggeman @rakelkar Appveyor is now passing

PR #898, opened 2017-12-07
backend/extension: Add Wireguard configuration

tomdee:
This PR adds
- the wireguard tools to the flannel container images (~0.1MB
size increase)
- an example extensions config
- a one-line change to extension.go to pass in the flannel network

The tests are commented out since they won't pass unless the wireguard
kernel module is loaded.

### DIFF
--- a/dist/extension-wireguard
+++ b/dist/extension-wireguard
@@ -1,0 +1,11 @@
+{
+  "Network": "10.50.0.0/16",
+  "Backend": {
+    "Type": "extension",
+    "PreStartupCommand": "wg genkey | tee privatekey | wg pubkey",
+    "PostStartupCommand": "export SUBNET_IP=`echo $SUBNET | cut -d'/' -f 1`; ip link del flannel-wg 2>/dev/null; ip link add flannel-wg type wireguard && wg set flannel-wg listen-port 51820 private-key privatekey && ip addr add $SUBNET_IP/32 dev flannel-wg && ip link set flannel-wg up && ip route add $NETWORK dev flannel-wg",
+    "ShutdownCommand": "ip link del flannel-wg",
+    "SubnetAddCommand": "read PUBLICKEY; wg set flannel-wg peer $PUBLICKEY endpoint $PUBLIC_IP:51820 allowed-ips $SUBNET",
+    "SubnetRemoveCommand": "read PUBLICKEY; wg set flannel-wg peer $PUBLICKEY remove"
+  }
+}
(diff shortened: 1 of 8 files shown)


klausenbusk:
Nice work :) A bit sad though that installing the Wireguard kernel module on CoreOS is a mess.

dist/extension-wireguard
```diff
@@ -0,0 +1,11 @@
+{
+  "Network": "10.50.0.0/16",
+  "Backend": {
+    "Type": "extension",
+    "PreStartupCommand": "wg genkey | tee privatekey | wg pubkey",
+    "PostStartupCommand": "export SUBNET_IP=`echo $SUBNET | cut -d'/' -f 1`; ip link del flannel-wg 2>/dev/null; ip link add flannel-wg type wireguard && wg set flannel-wg listen-port 8285 private-key privatekey && ip addr add $SUBNET_IP/32 dev flannel-wg && ip link set flannel-wg up && ip route add $NETWORK dev flannel-wg",
```
klausenbusk: Why port `8285` and not `51820` (which is the default WG port (?))?

zx2c4: Indeed `51820` is preferred.

zx2c4:
Great to hear you all are adding WireGuard support to CoreOS. Let me know if I can be of any assistance.

dist/extension-wireguard
tomdee: Good point, I'll update the port.

eranreshef:
I have a question that might be a bit silly but I hope you'll forgive since I'm new in this:
If I deployed the latest flanneld image in my k8s cluster, is there something special I need to do in order to "activate" wireguard? If not, how can I tell if its running?
Thanks.

tomdee:
@eranreshef This wireguard config isn't in a release yet, so unless you really know what you're doing I wouldn't recommend it! If you need more help then maybe swing by the #flannel channel on the calico users slack (see contact info in the readme)

eranreshef:
Well I'm not 100% know what I'm doing but I don't mind experimenting with this.
I deployed the latest flannel image in installed the wireguard module on the host. When the flanneld is starting it prints the following lines and exit:
```
I0116 14:30:01.902096       1 main.go:353] Found network config - Backend type: extension-wireguard

E0116 14:30:01.902115       1 main.go:272] Error fetching backend: unknown backend type: extension-wireguard
```

whats wrong here?

tomdee:
The config needs to look like this https://github.com/coreos/flannel/blob/master/dist/extension-wireguard

Hope that helps!

eranreshef:
Yes it does. Thanks!

PR #929, opened 2018-01-25
Add IPSec backend

tomdee:
* Uses ESP in tunnel mode and a PSK.
* Strongswan is used as the IKE daemon
* Communicate with the strongswan charon using VICI

Strongswan adds ~5MB (uncompressed) to the image size.

I have added minimal documentation and automated tested but it would be great to 
add more before this is released.

I have also only performed minimal manual testing, so this backend is 
"experimental" until more testing has been performed.

Huge thanks to @MohdAhmad @eyakubovich and @mkutsevol for their initial work on 
getting ipsec added to flannel.


rektide:
The authors have been named, but a quick shout out & backlinks to the PR's which have come before:
https://github.com/coreos/flannel/pull/637
https://github.com/coreos/flannel/pull/516
https://github.com/coreos/flannel/pull/290

PR #935, opened 2018-01-29
Added new flag -iptables-resync

SleepyBrett:
## Description
Referenced issue: https://github.com/coreos/flannel/issues/933

This pull creates a flag that can be used to adjust the resync period of the ensure iptables routine. On a larger cluster with many services we found 5 seconds to be too aggressive and created too much iptables lock contention causing kube-proxy to fail to sync it's tables.

## Todos
- [ ] Tests
- [x] Documentation
- [x] Release note

## Release Note
```release-note
Adds new flag --iptables-resync, allows adjustment of IPTables "ensure" check. Expects an integer it will then multiply by seconds, defaults to 5. Guidance is to adjust this upwards if you see too much iptables lock contention.
```


SleepyBrett:
I don't expect i need a test to test that time.Duration(int)*time.Second is functioning properly ;)

tomdee:
Looks great, thanks @SleepyBrett

PR #940, opened 2018-02-01
Log all the available interfaces when none were found

dmathieu:
## Description

I was trying to use the `--iface-regex` and my interface wasn't detected (for a dumb reason. I was adding quotation marks to the argument, and those were added to the string).
At first, I thought this was because my interfaces weren't properly detected by flannel though.

This improves the no detected interfaces error message to show all the available interfaces, name and IP address so it's clearer what the issue could be.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

I haven't written tests here because `main.go` doesn't have any, and setting up unit testing seems outside the purpose of this small PR.

## Release Note

```release-note
None required
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -474,7 +474,13 @@ func LookupExtIface(ifname string, ifregex string) (*backend.ExternalInterface, 
 
 		// Check that nothing was matched
 		if iface == nil {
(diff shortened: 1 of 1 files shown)


tomdee:
Looks good, thanks.

PR #982, opened 2018-04-19
subnet/kube: Handle missed node deletions

richardmarshall:
## Description

If the node controller misses the deletion of a node the DeltaFIFO will insert a tombstone object (DeletedFinalStateUnknown) with the last known view of the object. The current code will panic due to a failed type assertion to v1.Node.

This PR includes a fix for this panic by checking the result of the type assertion and attempting to extract the last know state from the tombstone.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


tomdee:
Sorry, I didn't spot this when I merged #1000

PR #989, opened 2018-04-28
Documentation/k8s: add daemonsets for all archs

### LINKED ISSUE #663: Please include instructions for installing on ARM
so we can link to them from https://kubernetes.io/docs/getting-started-guides/kubeadm/

lentzi90:
This makes it possible to deploy kube-flannel.yml on clusters with multiple architectures without modifications. Fixes #663.

## Description

As suggested in [this comment](https://github.com/coreos/flannel/issues/663#issuecomment-335328617), I have added a daemonset for each architecture in `kube-flannel.yml`. Each daemonset will only add pods to nodes with the correct architecture.

A nicer solution (in the future) for this would be to have something like [docker multi-arch manifests](https://blog.docker.com/2017/11/multi-arch-all-the-things/) so that only a single daemonset would be required.

I have only tried this on an amd64 cluster.

## TODO

- [x] Test on multi-arch cluster.

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -82,7 +82,7 @@ data:
 apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
-  name: kube-flannel-ds
+  name: kube-flannel-ds-amd64
   namespace: kube-system
   labels:
     tier: node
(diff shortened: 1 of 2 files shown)


drake7707:
I've tested this on a v1.11 cluster with the master node being amd64 and a raspberry pi (being arm) and it works without a problem (as soon as I had done the same strategy for kube-proxy, which is required for actually setting up the network interfaces).

Only issue I encountered is that it refused to deploy at first because of DiskPressure (4gb sd card with a default raspbian install doesn't leave much free disk space), which can be solved by setting the critical priority class as per https://github.com/coreos/flannel/issues/925,

tomdee:
Excellent, thanks for the fix.

PR #1040, opened 2018-09-20
Use --random-fully when creating MASQ rules

vainu-arto:
## Description

This makes Flannel apply a workaround for the infamous Linux conntrack race that causes major problems for many Kubernetes deployments, often manifesting as DNS delays or failures (see https://github.com/kubernetes/kubernetes/issues/56903, https://github.com/weaveworks/weave/issues/3287, https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts and many others).

## Todos
- [ ] Tests

## Release Note

```release-note
If the underlying iptables binary supports it, use --random-fully when creating MASQ rules
```

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -46,7 +46,7 @@ imports:
   - pkg/transport
   - pkg/types
 - name: github.com/coreos/go-iptables
-  version: 259c8e6a4275d497442c721fa52204d7a58bde8b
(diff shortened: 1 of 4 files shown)


tomdee:
Change looks good. Confirmed that v1.6.2 of iptables is in the latest alpine release so taking this PR.

PR #1043, opened 2018-09-26
remove "list" typo from daemonset file

aba182:
## Description

There is a type in the deployment doc that writes the configMap settings to `/etc/cni/net.d/10-flannel.conflist` instead of /etc/cni/net.d/10-flannel.conf` where it needs to be. I noticed this after our HairPin settings were not enabled. After a lengthy investigation I was able to track it down to the Documentation file having an error:

https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml

If required I can include a more indepth example but from the diff it should be clear there is a typo.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -110,7 +110,7 @@ spec:
         args:
         - -f
         - /etc/kube-flannel/cni-conf.json
-        - /etc/cni/net.d/10-flannel.conflist
+        - /etc/cni/net.d/10-flannel.conf
         volumeMounts:
         - name: cni
           mountPath: /etc/cni/net.d


aba182:
When we attempted to deploy v0.10.0 we had two different `/etc/cni/net.d/10-flannel.conf*` files one was .conf and the other was .conflist. 

The .conflist had our desired config but it wasn't applied. We discovered this when we attempted to leverage the `hairpin` system. 

To fix this we move from .conflist -> .conf and our issue was resolved.

ttarczynski:
I'm observing a similar issue after upgrading flannel from `v0.9.1` to ` v0.10.0`:
```
# ls -l /etc/cni/net.d/
total 8
-rw-r--r--. 1 root root  92 Oct  3 01:59 10-flannel.conf
-rw-r--r--. 1 root root 267 Oct 12 03:16 10-flannel.conflist
```

I see that the config file name has changed from `10-flannel.conf` to `10-flannel.conflist` in: #888
But I believe there's no mechanism that would delete the old `10-flannel.conf` file after the upgrade.

rajatchopra:
Fair enough. Although this is not a typo, it may cause unintended issues with leftovers.

PR #1061, opened 2018-11-09
Remote subnet feature

ksubrmnn:
## Description

This PR implements the HNS Remote Subnet feature. With this feature, Flannel in overlay will create one network policy per subnet instead of creating several endpoints per subnet.
## Todos
-  Tests
-  Documentation

## Release Note

```release-note
Requires Windows Build 18301 or later
```


rajatchopra:
@ksubrmnn do you plan to add the tests and documentation in this PR?

ksubrmnn:
@rajatchopra Testing and documentation will probably take about 2 weeks. I think it's fine to merge this PR and follow up with testing/docs in a separate PR, if that's okay with you. This has been manually validated.

backend/vxlan/vxlan_network_windows.go
```diff
@@ -87,36 +95,66 @@ func (nw *network) handleSubnetEvents(batch []subnet.Event) {
...
+		hnsnetwork, err := hcn.GetNetworkByName(nw.dev.link.Name)
+		if err != nil {
+			log.Errorf("Unable to find network %v, error: %v", nw.dev.link.Name, err)
+			continue
+		}
+		managementIp := event.Lease.Attrs.PublicIP.String()
+
+		networkPolicySettings := hcn.RemoteSubnetRoutePolicySetting{
+			IsolationId:                 4096,
+			DistributedRouterMacAddress: net.HardwareAddr(vxlanAttrs.VtepMAC).String(),
```
rajatchopra: Just a question. How was it even working with ConjureMAC before? Not ARP right? Or was the MAC manufactured and dropped on the interface?

ksubrmnn: Previously there was a remote endpoint for each IP in the subnet on the host. For example if Host A has subnet 192168.1.0/24 and Host B has 192.168.2.0/24, then Host A had 250+ remote endpoints from 192.168.2.1-254. Vice versa for Host B

This was before HNS had the remote subnet feature. And before Remote Subnet, many things were not working properly like service vip, node port access, etc. With remote subnet feature, this RemoteSubnetPolicy replaces the tens/hundreds of endpoints you previously had to create per other node in the cluster. And this policy cannot use a random mac, it needs the dr/vtep mac of the other hosts by design.

ksubrmnn:
@rajatchopra let me do one final run in the morning for peace of mind, then we can merge. 

Only upcoming changes I can foresee are removing hardcoding of VNI (necessary for now until I can add a small Linux change) and testing. Both of those will be separate PRs from this though.

ksubrmnn:
@rajatchopra this looks good to merge.

backend/vxlan/vxlan_network_windows.go
rajatchopra: One would have hoped that the MAC can be learned through ARP. But I assume this is for added security.

rajatchopra:
@ksubrmnn Thanks.

PR #1068, opened 2018-11-27
Fix getting management ip for host-gw mode running on windows.

song-jiang:
## Description

This PR is to fix issue https://github.com/coreos/flannel/issues/1066
- Create and attach endpoint to cbr0 network before checking management ip.
- Fix code to return non-nil error if checking for management ip timed out. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/backend/hostgw/hostgw_windows.go
+++ b/backend/hostgw/hostgw_windows.go
@@ -160,26 +160,6 @@ func (be *HostgwBackend) RegisterNetwork(ctx context.Context, wg sync.WaitGroup,
 			return nil, errors.Annotatef(err, "failed to create HNSNetwork %s", networkName)
 		}
 
-		// Wait for the network to populate Management IP
-		log.Infof("Waiting to get ManagementIP from HNSNetwork %s", networkName)
-		waitErr = wait.Poll(500*time.Millisecond, 5*time.Second, func() (done bool, err error) {
-			newNetwork, lastErr = hcsshim.HNSNetworkRequest("GET", newNetwork.Id, "")
-			return newNetwork != nil && len(newNetwork.ManagementIP) != 0, nil
-		})
-		if waitErr == wait.ErrWaitTimeout {
-			return nil, errors.Annotatef(lastErr, "timeout, failed to get management IP from HNSNetwork %s", networkName)
-		}
-
-		// Wait for the interface with the management IP
-		log.Infof("Waiting to get net interface for HNSNetwork %s (%s)", networkName, newNetwork.ManagementIP)
-		waitErr = wait.Poll(500*time.Millisecond, 5*time.Second, func() (done bool, err error) {
-			_, lastErr = netshHelper.GetInterfaceByIP(newNetwork.ManagementIP)
-			return lastErr == nil, nil
-		})
-		if waitErr == wait.ErrWaitTimeout {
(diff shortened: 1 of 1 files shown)


song-jiang:
@thxCode @rajatchopra Could you take a look? Thanks.

ksubrmnn:
@madhanrm

backend/hostgw/hostgw_windows.go
```diff
@@ -160,26 +160,6 @@ func (be *HostgwBackend) RegisterNetwork(ctx context.Context, wg sync.WaitGroup,
 			return nil, errors.Annotatef(err, "failed to create HNSNetwork %s", networkName)
 		}
 
-		// Wait for the network to populate Management IP
-		log.Infof("Waiting to get ManagementIP from HNSNetwork %s", networkName)
-		waitErr = wait.Poll(500*time.Millisecond, 5*time.Second, func() (done bool, err error) {
-			newNetwork, lastErr = hcsshim.HNSNetworkRequest("GET", newNetwork.Id, "")
-			return newNetwork != nil && len(newNetwork.ManagementIP) != 0, nil
```
ksubrmnn: Can you add a nil check for newNetwork.ManagementIP here and see if that resolves the error? This may be a better solution

ksubrmnn: Also try increasing the timeout to 2 minutes with this change

ksubrmnn: @song-jiang There is a known issue in RS4 that the ManagementIP may not be populated until the host receives an IP. This does not occur in RS5 onwards. Can you try delaying starting flannel until the host receives an IP? I think flannel may be starting too soon

song-jiang: @ksubrmnn I can confirm the host did receive an IP before starting flannel process. I used powershell script similar to https://github.com/Microsoft/SDN/blob/master/Kubernetes/flannel/l2bridge/start.ps1 with small modifications. 

The flanneld process is started after
- `flanneld.exe` binary downloaded from internet.
- Docker images pulled.
- windows node registered to kube-apiserver. 
- vSwitch created.

ksubrmnn: Is it possible to get access to your set up? This issue should not happen and we have not been able to repro it.

song-jiang: @ksubrmnn I will try to reproduce and come back to you. Could you email your contact information to song@tigera.io for me to send you access credentials? Thanks.

ksubrmnn:
@JocelynBerrendonner

JocelynBerrendonner:
@dineshgovindasamy FYI

manuelbuil:
This PR has not had any activity for years, should we close it?

daschott:
IIRC, this issue does not reproduce on Windows Server 2019, so this PR should no longer be needed?

manuelbuil:
> IIRC, this issue does not reproduce on Windows Server 2019, so this PR should no longer be needed?

Thanks for the info. Let's wait a couple of days and if the author says nothing, we'll close it

manuelbuil:
Let's close it for the moment. We can open it again if needed

PR #1071, opened 2018-12-08
Fix nil  pointer error where get ManagementIP timeout

barnettZQG:
## Description
Fix nil  pointer error where get ManagementIP timeout, and extended get ManagementIP wait time to 30s
Fix issues https://github.com/coreos/flannel/issues/1066


ksubrmnn:
@rajatchopra this looks fine to me. I'm still looking at [PR 1068](https://github.com/coreos/flannel/pull/1068), but this can be merged in the meantime

PR #1078, opened 2019-01-02
Use V2 to get networks

ksubrmnn:
## Description

Gets HNS networks using V2 schema. This removes the redundancy of getting the networks in V1 and then again in V2.

This also fixes an issue users are seeing where a network is listed in the query for V1, but then that network is not able to be found in V2.
## Release Note

```release-note
Updated VXLAN windows to use hcsshim/hcn API entirely instead of hcsshim.
```

### DIFF
--- a/backend/vxlan/vxlan_windows.go
+++ b/backend/vxlan/vxlan_windows.go
@@ -33,7 +33,6 @@ import (
 
 	"golang.org/x/net/context"
 
-	"github.com/Microsoft/hcsshim"
 	"github.com/Microsoft/hcsshim/hcn"
 	"github.com/coreos/flannel/backend"
 	"github.com/coreos/flannel/pkg/ip"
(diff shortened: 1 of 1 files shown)


ksubrmnn:
@rajatchopra

daschott:
ping @rajatchopra PTAL

ksubrmnn:
ping @madhanrm

backend/vxlan/vxlan_windows.go
```diff
@@ -142,12 +141,24 @@ func (be *VXLANBackend) RegisterNetwork(ctx context.Context, wg sync.WaitGroup,
...
+					}
+					providerAddress = policySettings.ProviderAddress
+				}
+			}
+			if providerAddress != be.extIface.ExtAddr.String() {
+				log.Infof("Cannot use DR MAC %v since PA %v does not match %v", remoteDrMac, providerAddress, be.extIface.ExtAddr.String())
+				remoteDrMac = ""
 			}
 		}
 	}
```
madhanrm: At the exit of the for loop, can you add a check for the availability of providerAddress & remoteDrMac. Fail, with a meaningful error if you do not have them, since you cannot proceed.

ksubrmnn: Added below

rajatchopra:
@madhanrm lgty?

PR #1084, opened 2019-01-17
Makefile: fix indentation

rajatchopra:
## Description
Fix Makefile: it requires 'tabs' before actions on targets, also the bash 'if' condition needs a syntax fix.

## Release Note
```release-note
None required
```


rajatchopra:
@tomdee lgty?

Issue #1090, opened 2019-01-28
/run is mouted r/w by default in kubernetes


inercia:
## Current Behavior

Flannel's [official deployment](https://raw.githubusercontent.com/coreos/flannel/master/Documentation/k8s-manifests/kube-flannel-legacy.yml) for kubernetes is currently mounting `/run` with r/w mode in all the pods in the `DaemonSet`. This gives access to too many provileged stuff in the host (including the docker socket).

## Expected Behavior

Taking into account the Flannel container is being run as a `privileged` container in Kubernetes, it would be preferable to mount some other directory or some subdirectory of `/run`.

## Possible Solution

* To use some other directory.
* To mounted `/run/flannel` (or some other subdirectory or `/run`).

PR #1094, opened 2019-01-30
Deploy Flannel with unprivileged PSP

drpaneas:
Flannel is running in privileged mode, and is using the root user,
therefore if the container is compromised, an attacker will
inherit that level of access. An attacker compromising the flannel
container will have full root access to the host system, and access
to all secrets and containers hosted on the system.

This PR makes sure that flannel runs in unprivileged mode.

This is done by changing the flannel manifests and also adding
a new PSP policy that disables both privilege mode and privilege
escallation.

The new PSP activates 'NET_ADMIN' capability, hostNetwork
and allowedHostPaths.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel-old.yaml
+++ b/Documentation/kube-flannel-old.yaml
@@ -1,60 +1,9 @@
 ---
-apiVersion: extensions/v1beta1
-kind: PodSecurityPolicy
-metadata:
-  name: psp.flannel.unprivileged
-  annotations:
-    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
-    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
-    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
-    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
-spec:
-  privileged: false
-  volumes:
-    - configMap
-    - secret
-    - emptyDir
-    - hostPath
-  allowedHostPaths:
-    - pathPrefix: "/etc/cni/net.d"
-    - pathPrefix: "/etc/kube-flannel"
-    - pathPrefix: "/run/flannel"
-  readOnlyRootFilesystem: false
-  # Users and groups
-  runAsUser:
-    rule: RunAsAny
-  supplementalGroups:
-    rule: RunAsAny
-  fsGroup:
-    rule: RunAsAny
-  # Privilege Escalation
-  allowPrivilegeEscalation: false
-  defaultAllowPrivilegeEscalation: false
-  # Capabilities
-  allowedCapabilities: ['NET_ADMIN']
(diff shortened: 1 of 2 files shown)


rajatchopra:
@mrunalp Can you take a look and review? Thanks.

Documentation/kube-flannel.yml
```diff
@@ -1,9 +1,60 @@
...
+    # SELinux is unsed in CaaSP
+    rule: 'RunAsAny'
+---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: flannel
 rules:
+  - apiGroups: ['extensions']
+    resources: ['podsecuritypolicies']
```
rajatchopra: PSP is still beta(?). Nevertheless I presume this will be a paper-over if we run a backward compatible kubernetes without PSP reconciler. Which may mean that we run quite underprivileged. Do we want to document the compatibility?

drpaneas: Removing the beta I end up with:

```
Error from server (BadRequest): error when creating "STDIN": PodSecurityPolicy in version "v1" cannot be handled as a PodSecurityPolicy: no kind "PodSecurityPolicy" is registered for version "extensions/v1"
```

I've also noticed that you are using a really old kubernetes version when it comes to testing:
```
K8S_VERSION="${K8S_VERSION:-1.7.6} ---> ./dist/functional-test-k8s.sh
```
As a result I had to bump the version in the e2e tests to the latest stable one (1.13.2). However this up to you, I don't want to be too intrusive with this PR, just raise the security concerns.

rajatchopra: I suspect we need SYS_ADMIN because we mount /run and write to it (though I would agree with #1090 that we should be more specific). But a real test will prove it.
I am in favour of bumping e2e to updated kube version (PR please?).

This PR is good, but we need some hint about the compatibility - what would be the oldest supported Kube version with this merged in?

drpaneas:
Is there anything else to do? Can we merge this?

rajatchopra:
> Is there anything else to do? Can we merge this?

@drpaneas I think we should maintain a kube-flannel.yaml for older versions. Do you mind doing that within this PR? Call the old one old-kube-flannel.yaml or pre198-kube-flannel.yaml  or something, I will update the Documentation indicating on the file to be used for older deployments. Thanks.

drpaneas:
> > Is there anything else to do? Can we merge this?
> 
> @drpaneas I think we should maintain a kube-flannel.yaml for older versions. Do you mind doing that within this PR? Call the old one old-kube-flannel.yaml or pre198-kube-flannel.yaml or something, I will update the Documentation indicating on the file to be used for older deployments. Thanks.

sure, no problem. done :)

PR #1127, opened 2019-04-17
ipsec: fix host network

martin31821:
## Description

This PR fixes the IPSec backend when used in conjunction with kubernetes. 
It does so by generating a dummy interface and adding the flannel network as route towards the subnet.
After that, if kubernetes generates the cni interface, it will add another route to ensure the node-local network (= allocated network) is routed to the CNI interface.

This PR was tested against kubernetes 1.13.5/kubespray based setup on a three-node cluster. No automated testing was performed. The kubespray changes can be found [here](https://github.com/kubernetes-sigs/kubespray/pull/4527) 

No further user-facing changes are neccessary.

**EDIT** The description is deprecated, see the comments below. 

```release-note
None required
```


martin31821:
I took the last few days to make a better fix for the kubernetes-ipsec host-network problem.
Now there is no interface generated anymore, I just setup the routes correctly like in the strongswan documentation. 
Furthermore I added a flag to indicate whether flannel is running within kubernetes to change source addresses for the ClusterIP services.

alvaroaleman:
> This PR fixes the IPSec backend when used in conjunction with kubernetes.

Could you update your PR to state what issue exactly is being fixed/What doesn't work currently and will work after your PR?

martin31821:
> > This PR fixes the IPSec backend when used in conjunction with kubernetes.
> 
> Could you update your PR to state what issue exactly is being fixed/What doesn't work currently and will work after your PR?

Yeah sure, short description:
When you are using flannel+ipsec backend as underlay network for kubernetes and you use the new IPVS proxy mode, you can't communicate from pods with HostNetwork to a service with a ClusterIP.
This PR fixes exactly this issue by changing the routes for the kube-ipvs0 interface to have a valid prefsrc address. When having this prefsrc address, the kernel won't drop the packets on the return route.
You can also refer to the issues I linked for further information.

PR #1132, opened 2019-04-29
clean redundant code in building kubernetes config

mars1024:
## Description

function [BuildConfigFromFlags](https://github.com/coreos/flannel/blob/ba49cd4c1e49d566da4a08b370384ce8ced0c0e3/vendor/k8s.io/client-go/tools/clientcmd/client_config.go#L515) will build kubernetes config from masterURL or kubeconfig filepath firstly, if both are empty, it will fall back to build in-cluster config, if fails again, it will fall back to default config. So I think the origin codes for building kubernetes config are redundant and need some clean.

```release-note
None required
```

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -68,17 +68,12 @@ func NewSubnetManager(apiUrl, kubeconfig, prefix string) (subnet.Manager, error)
 
 	var cfg *rest.Config
 	var err error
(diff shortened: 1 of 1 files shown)


mars1024:
@ksubrmnn @rajatchopra  Sorry for disturbing, do you have time to look at this pr ? Looking forward to your reviews ~

PR #1135, opened 2019-05-13
add cni version to cni-conf yaml

SchSeba:
This PR adds the cni version to the cni-conf.yaml inside the kube-flannel-cfg configmap

This PR fixes an issue we latest multus version.
Multus is unable to call the cni delete function because the cni version is not configured.

Because of this issue pod addresses are not released

Error from pod:
```
Warning  FailedKillPod  8s     kubelet, node01    error killing pod: failed to "KillPodSandbox" for "cca49d68-758d-11e9-b08c-525500d15501" with KillPodSandboxError: "rpc error: code = Unknown desc = NetworkPlugin cni failed to teardown pod \"pod-example_default\" network: Multus: error in invoke Conflist Del - \"cbr0\": error in getting result from DelNetworkList: failed to convert major version part \"\": strconv.Atoi: parsing \"\": invalid syntax"

```

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -106,6 +106,7 @@ data:
   cni-conf.json: |
     {
       "name": "cbr0",
+      "cniVersion": "0.3.1",
       "plugins": [
         {
           "type": "flannel",
--- a/Documentation/k8s-manifests/kube-flannel-legacy.yml
+++ b/Documentation/k8s-manifests/kube-flannel-legacy.yml
@@ -17,6 +17,7 @@ data:
   cni-conf.json: |
     {
       "name": "cbr0",
+      "cniVersion": "0.3.1",
       "type": "flannel",
       "delegate": {
         "hairpinMode": true,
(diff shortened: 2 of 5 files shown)


Documentation/kube-flannel.yml
```diff
@@ -106,6 +106,7 @@ data:
   cni-conf.json: |
     {
       "name": "cbr0",
+      "cniVersion": "0.3.1",
```
saschagrunert: I guess this has to be added to the other yml files as well, right?

SchSeba: you are right I do it can you please take another look?

saschagrunert:
@rajatchopra do you think we can get this in? It seems to trigger upstream regressions when used together with CRI-O.

lrgar:
Hello,

I was playing with the Kubernetes release 1.16.0-beta.1 released a couple of days ago, and now there, Kubelet's Docker Shim tries to validate the `cni-conf.json` file on Docker nodes.

With the master branch's `Documentation/kube-flannel.yml` config, you'd get the error below from Kubelet, and the nodes won't get Ready:

```
Aug 22 08:38:35 ip-10-0-0-191 kubelet[17160]: W0822 08:38:35.287062   17160 cni.go:202] Error validating CNI config &{cbr0  false ..... ]}: [plugin flannel does not support config version ""]
```

After applying the changes on this PR it works fine.

PR #1137, opened 2019-05-16
Remove dependency on pre-existing Overlay network for Flannel VXLAN for Windows.

### LINKED ISSUE #1111: Windows overlay backend depends on external HNS network
## Expected Behavior
The Windows overlay backend [assumes you have created an external HNS network](https://github.com/coreos/flannel/blob/e4deb05e97ceb68d43df022fb76cd505a38a6973/backend/vxlan/vxlan_windows.go#L137-L167), and fails if it can't find one. This isn't very clearly documented and we had to dig into Microsoft's SDN code to [find reference](https://github.com/Microsoft/SDN/blob/1d5c055bb195fecba07ad094d2d7c18c188f9d2d/Kubernetes/windows/helper.psm1#L96-L118) to what this backend is expecting. This also has the unfortunate consequence of requiring users to import Microsoft's [hns.psm1](https://github.com/Microsoft/SDN/blob/1d5c055bb195fecba07ad094d2d7c18c188f9d2d/Kubernetes/windows/hns.psm1) library to add the `Add-HNSNetwork` function to Powershell. If it were just part of Flannel it could use hcsshim to do the same thing.

It would be better and more intuitive if the overlay backend would just create this itself. I don't know if there's a reason it can't do that.

## Possible Solution
Create the external network if it doesn't already exist.

@ksubrmnn

ksubrmnn:
## Description
Fixes #1111 

Instead of querying an existing Overlay network for a DR MAC address, we proceed with creating the Flannel network. We then get the DR MAC from this network and update the lease accordingly.

The first SubnetAdded event will be ignored by other Linux and Windows nodes since the VtepMAC will be nil. The second SubnetAdded event will contain the MAC so event handling will continue as expected.

## Todos
- [ ] Tests
- [ ] Documentation (to explain that errors from not being able to decode a nil mac are expected and normal for the first subnet event)
- [x] Release note

## Release Note

```release-note
Remove dependency on pre-existing Overlay network for Flannel VXLAN for Windows.
```

### DIFF
--- a/backend/vxlan/vxlan_windows.go
+++ b/backend/vxlan/vxlan_windows.go
@@ -64,9 +64,13 @@ func New(sm subnet.Manager, extIface *backend.ExternalInterface) (backend.Backen
 }
 
 func newSubnetAttrs(publicIP net.IP, vnid uint16, mac net.HardwareAddr) (*subnet.LeaseAttrs, error) {
+	var hardwareAddress hardwareAddr
+	if mac != nil {
+		hardwareAddress = hardwareAddr(mac)
+	}
 	leaseAttrs := &vxlanLeaseAttrs{
 		VNI:     vnid,
-		VtepMAC: hardwareAddr(mac),
+		VtepMAC: hardwareAddress,
 	}
 	data, err := json.Marshal(&leaseAttrs)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ksubrmnn:
@madhanrm @benmoss

ksubrmnn:
@thxCode

ksubrmnn:
@benmoss Can you validate? I cleaned this up a little and didn't do another run yet. Wanted to get this out by EOD, but I'll validate again in the morning.

benmoss:
LGTM! Works without the external network now, as you said

ksubrmnn:
@rajatchopra

ksubrmnn:
@daschott

PR #1153, opened 2019-06-27
iptables: handle errors that prevent rule deletes

### LINKED ISSUE #1146: Ensure that iptables NAT rules are maintained in the correct order
We have encountered cases where the iptables NAT rules that Flannel maintains end up in an incorrect order and affect networking for pods on the host.  Our suspicion is that something other than Flannel is causing this ordering issue, but since Flannel already watches for missing rules this seems like something else that it could check.

## Expected Behavior

**Change/improvement:** Flannel recreates the NAT rules if any of them are missing _or_ if they are in the incorrect order. 

## Current Behavior

The current behavior is that Flannel will recreate the iptables NAT rules that it manages if any of them are missing but not if they are incorrectly ordered.

## Possible Solution

[ipTablesRulesExist](https://github.com/coreos/flannel/blob/v0.11.0/network/iptables.go#L85) could list the rules in the NAT table's POSTROUTING chain and validate that they exist and match the expected order.

## Your Environment

* Flannel version: 0.9.1
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version: 3.3.12
* Kubernetes version (if used): 1.13
* Operating System and version:
* Link to your project (optional):

willgorman:
Fixes #1146

## Description

The cause of the incorrect order of Flannel's rules in #1146 was found to be a case where an error occurs on deleting a rule and Flannel ignores the error and assumes the rule was already deleted.  As a result, the remaining rules are re-appended after the rule that could not be deleted, possibly causing them to be in the wrong order.  This change stops deleting additional rules on error and waits for the next execution of the reconciliation loop to correct the rules (assuming that the error deleting the rule was transient).

Type: Bug fix
Testing: In addition to the unit test, this has been tested as a patch on 0.9.1 (deployed on Kubernetes 1.13) where I verified that the error message for failing to teardown did get logged and that the rules on that host remained in the correct order.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/network/iptables.go
+++ b/network/iptables.go
@@ -34,6 +34,11 @@ type IPTables interface {
 	Exists(table string, chain string, rulespec ...string) (bool, error)
 }
 
+type IPTablesError interface {
+	IsNotExist() bool
+	Error() string
+}
+
 type IPTablesRule struct {
 	table    string
 	chain    string
@@ -143,7 +148,9 @@ func ensureIPTables(ipt IPTables, rules []IPTablesRule) error {
 	// Otherwise, teardown all the rules and set them up again
 	// We do this because the order of the rules is important
 	log.Info("Some iptables rules are missing; deleting and recreating rules")
-	teardownIPTables(ipt, rules)
+	if err = teardownIPTables(ipt, rules); err != nil {
+		return fmt.Errorf("Error tearing down rules: %v", err)
+	}
 	if err = setupIPTables(ipt, rules); err != nil {
 		return fmt.Errorf("Error setting up rules: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


network/iptables.go
```diff
@@ -162,11 +169,23 @@ func setupIPTables(ipt IPTables, rules []IPTablesRule) error {
...
+		if err != nil {
+			e := err.(IPTablesError)
+			// If this error is because the rule is already deleted, the message from iptables will be
+			// "Bad rule (does a matching rule exist in that chain?)". These are safe to ignore.
+			// However other errors (like EAGAIN caused by other things not respecting the xtables.lock)
+			// should halt the ensure process.  Otherwise rules can get out of order when a rule we think
+			// is deleted is actually still in the chain.
+			// This will leave the rules incomplete until the next successful reconciliation loop.
+			if !e.IsNotExist() {
+				return err
```
rajatchopra: Should we not aggregate the errors and return them after running through the loop? We do not want a bad rule to block others from being removed. That can result in `all` rules staying stuck because of one broken one.

willgorman: This presumes that the error deleting the rule was transient and that if we bail out here then it is likely to succeed on the next iteration of the reconciliation loop.  Deleting other rules after a failure is what complicates things when trying to reinsert the rules in the right order. If there’s a rule that consistently fails to delete I don’t know how much it matters that it blocks the others since it would be difficult to guarantee correct behavior in that state anyway.

sethev:
May be worth adding that we were seeing failures due to misordered flannel iptables rules several times a week prior to this fix and haven't had it recur since applying this change.

sethev:
@rajatchopra any more thoughts on this?

benhemp:
This has now been tested in production on dozens of clusters, many with 100+ nodes on kubernetes 1.14.3, for many weeks now, still no issues. Is there any supporting information that would be helpful?

sethev:
@rajatchopra any thoughts on this? Is there more information or digging we could do to help?

clayvan:
@tomdee @rajatchopra Can we get some more eyes on this? This is definitely a fix we would like merged into new versions.

benhemp:
We have still seen no issues since this fix was implemented, we now have 1000+ nodes running on this fix, and have gone through several upgrades. Anything else we can do?

benhemp:
@rajatchopra what more is needed here? 

If it helps, we have been maintaining a fork with this change and have deployed it in production with up to 200 node clusters, at a total of many thousands of nodes running from 1.13 to 1.14 to 1.15 with no re-occurances of the issue linked to this PR. The issue the PR resolves caused us a tremendous amount of pain, so we'd really like it upstream so others don't see it as well.

andersonnicholaskc:
@rajatchopra @tomdee or anyone else.....could we get this merged please? See above comments.

sw-elling:
bump

It would be helpful for us and the community to get these changes merged in.

PR #1162, opened 2019-07-18
Update deprecated API resources

aojea:
## Description

In  the k8s 1.14.0 release notes, there are concrete timelines for deprecated resources in the extensions/v1beta1, apps/v1beta1, and apps/v1beta2 API groups to no longer be served by default.

https://groups.google.com/d/msg/kubernetes-dev/je0rjyfTVyc/gEUw1YcyAQAJ

The flannel installation manifest need to be updated in order to work

## Todos
- [ ] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Update kubernetes deprecated API resources in the installation manifest 
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#deprecations
```

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: psp.flannel.unprivileged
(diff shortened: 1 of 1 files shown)


ksubrmnn:
@rajatchopra

rajatchopra:
@aojea Thanks. This looks good.
We need a new release for this then, which is fine. But we need to prevent users from using the yaml file from master branch for older versions of kubernetes. Any thoughts?

aojea:
The link I added to the description mention this is available since 1.9 so we should be safe with older versions 

> . Migrate to the apps/v1 API, available since v1.9

PR #1165, opened 2019-07-24
Needs to clear NodeNetworkUnavailable flag on Kubernetes

aarnaud:
Like calico : https://github.com/projectcalico/node/pull/134

## Description
http://www.deadunicornz.org/blog/2017/11/08/gce-and-self-hosted-k8s-1-dot-6-noroutecreated/index.html
There is an annoying bug since Kubernetes 1.6 running on GCE with Calico/Flannel networking operating via CNI plugin interface. You may experience same issue if kubelet on your nodes runs with options `--network-plugin=cni --cloud-provider=gce`

When a new node is added to the k8s cluster it’s recognized as ‘Ready’ but no pods except Calico/Flannel pod are being scheduled there.

Please include 
- bug fix
- Fix the `NetworkUnavailable` `False` on GCE with flannel and vxlan backend
- manual test : deploy on our infra. 
- kubeSubnetManager components are affected by this PR

Ref code : https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_network.go#L50

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -280,6 +280,10 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, attrs *subnet.Le
 			return nil, err
 		}
 	}
+	err = ksm.setNodeNetworkUnavailableFalse()
+	if err != nil {
+		glog.Errorf("Unable to set NetworkUnavailable to False for %q: %v", ksm.nodeName, err)
+	}
 	return &subnet.Lease{
 		Subnet:     ip.FromIPNet(cidr),
 		Attrs:      *attrs,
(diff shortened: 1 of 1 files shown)


subnet/kube/kube.go
```diff
@@ -280,6 +280,10 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, attrs *subnet.Le
 			return nil, err
 		}
 	}
+	err = ksm.setNodeNetworkUnavailableFalse()
```
rajatchopra: Shouldn't this be in an exponential retry loop?
Or do you want to rely on kubelet re-spinning the pod?

aarnaud: I don't know realy. I suppose we will never catch the error except if the node doesn't exist anymore. 
Flanneld is restart if kublet restart ?

aarnaud: @rajatchopra it's ok for you ? it's a fix for custom k8s cluster on GCE

PR #1167, opened 2019-08-02
Fix LoadBalancer externalTrafficPolicy:Local issue on upgrade to v0.11

BSWANG:
## Description

Remove iptables create by previous version which without `--ramdom-fully`
If not remove the iptables rules. The `externalTrafficPolicy:Local` will can not reserve the client IP.

```
Chain POSTROUTING (policy ACCEPT 1 packets, 60 bytes)
 pkts bytes target     prot opt in     out     source               destination
  283 20269 KUBE-POSTROUTING  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes postrouting rules */
    0     0 MASQUERADE  all  --  *      !docker0  169.254.123.0/24     0.0.0.0/0
   12   771 MASQUERADE  all  --  *      *       172.30.0.0/16       !224.0.0.0/4
    0     0 MASQUERADE  all  --  *      *      !172.30.0.0/16        172.30.0.0/16
    0     0 RETURN     all  --  *      *       172.30.0.0/16        172.30.0.0/16
    0     0 MASQUERADE  all  --  *      *       172.30.0.0/16       !224.0.0.0/4          random-fully
    0     0 RETURN     all  --  *      *      !172.30.0.0/16        172.30.0.128/25
    0     0 MASQUERADE  all  --  *      *      !172.30.0.0/16        172.30.0.0/16        random-fully
```

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note (need review)

## Release Note

```release-note
Maybe: Fix `externalTrafficPolicy: Local` service can not reserve client ip after upgrade flannel.
```

### DIFF
--- a/network/iptables.go
+++ b/network/iptables.go
@@ -43,6 +43,21 @@ type IPTablesRule struct {
 	rulespec []string
 }
 
+func (r *IPTablesRule) Equal(o *IPTablesRule) bool {
+	if r == nil || o == nil {
+		return false
+	}
+	if r.chain != o.chain || r.table != o.table || len(r.rulespec) != len(o.rulespec) {
+		return false
+	}
+	for i := range r.rulespec {
+		if r.rulespec[i] != o.rulespec[i] {
+			return false
+		}
+	}
+	return true
+}
+
 func MasqRules(ipn ip.IP4Net, lease *subnet.Lease) []IPTablesRule {
 	n := ipn.String()
 	sn := lease.Subnet.String()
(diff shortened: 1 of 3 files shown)


anton-johansson:
Any news on this? We're running `flannel` version `0.11.0` and I'm trying to introduce MetalLB in front of our NGINX pods to make sure we can see the real IP address, but it doesn't seem to work. I stumbled over this issue which seems to be related.

BSWANG:
@rakelkar @SleepyBrett @tomdee

rajatchopra:
@BSWANG Looks useful. Please rebase.

BSWANG:
@rajatchopra 
Rebased.

manuelbuil:
Hey, sorry for not paying attention to this PR. Could you rebase again? After that, I'll review and hopefully merge it!

BSWANG:
@manuelbuil Done!

network/iptables.go
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//go:build !windows
```
manuelbuil, line 14: As we already have line 15, does this line bring any value?

network/iptables_test.go
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//go:build !windows
```
manuelbuil, line 14: same comment

network/iptables.go
```diff
@@ -202,19 +254,40 @@ func DeleteIP6Tables(rules []IPTablesRule) error {
...
-	if exists {
+	ensureDelete := make([]IPTablesRule, 0)
+	for _, ruleToDelete := range rulesToDelete {
+		ruleExist := false
+		for _, rule := range rules {
+			if rule.Equal(&ruleToDelete) {
+				ruleExist = true
+				break
+			}
+		}
```
manuelbuil, line 270: What happens if we try to delete them and they don't exist? I wonder if that might be better than checking if they exist or not

BSWANG: OK, Added rule check before add to rulelist.

network/iptables.go
BSWANG: Remove the useless `go:build` line which added by goland automatic.

manuelbuil:
I was trying to reproduce the issue but I can't see those iptable rules. Are they only being created before v0.11?

BSWANG:
@manuelbuil Yes, those iptables rules created without 'random-fully' before v0.11. And not got cleanup when upgrade to v0.11+.

manuelbuil:
> @manuelbuil Yes, those iptables rules created without 'random-fully' before v0.11. And not got cleanup when upgrade to v0.11+.

In that case, I wonder if we should merge this. This only affects very very old versions and we can't backport. Therefore, it will onlyaffect users moving from v0.10 to v0.15, which is really strange, right?

luthermonson:
this is for such an older release... i hate to say this but maybe this is stale now and we should just close it. unfortunately maybe just a clean install is required now and we'll have to live with that. reopen if you have a problem with that @BSWANG

PR #1177, opened 2019-09-18
apiVersion: Beta -> Stable

RaSla:
* rbac.authorization.k8s.io/v1beta1 -> rbac.authorization.k8s.io/v1
* beta.kubernetes.io/arch -> kubernetes.io/arch

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/kube-flannel-before-1.16.yml
+++ b/Documentation/kube-flannel-before-1.16.yml
@@ -31,7 +31,7 @@ spec:
   allowPrivilegeEscalation: false
   defaultAllowPrivilegeEscalation: false
   # Capabilities
-  allowedCapabilities: ['NET_ADMIN','NET_RAW']
+  allowedCapabilities: ['NET_ADMIN']
   defaultAddCapabilities: []
   requiredDropCapabilities: []
   # Host namespaces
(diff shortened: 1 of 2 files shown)


arasananbu:
LGTM. support for Kubernetes 1.16. Will not work for <1.16. Are we changing the support to k8s >= 1.16, Then should be release noted.

Ref: https://kubernetes.io/docs/setup/release/notes

rajatchopra:
We need to have another file for <1.16 if there is a split here. @RaSla can you copy the older one with an extension like kube-flannel-beta.yaml?

RaSla:
I have saved old file as **kube-flannel-before-1.16.yml**.
And added "**"cniVersion": "0.3.1",**" into **kube-flannel.yml** for correct k8s-1.16 support

rajatchopra:
@RaSla Please update the PR. cniVersion was updated to 0.20 already, so maybe do not include that as part of this PR.

RaSla:
* Synced **Documentation/kube-flannel.yml** with Master (cniVersion, nodeAffinity)
* Patched versions: Beta -> Stable for **nodeAffinity** block

PR #1180, opened 2019-09-19
Updates imports used by bhendo/go-powershell

gab-satchi:
Signed-off-by: Ben Moss <bmoss@pivotal.io>

## Description
- Type: Fix
- Flannel currently doesn't build for Windows due to a broken dependency in bhendo/go-powershell. This PR fixes the import statements to use to the correct urls. 

Note: It seems CI doesn't currently build for Windows. 

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -68,3 +68,4 @@ import:
 - package: github.com/sirupsen/logrus
   version: v1.0.6
 - package: github.com/buger/jsonparser
+- package: github.com/bhendo/go-powershell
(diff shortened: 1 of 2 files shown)


arasananbu:
Is this should be handled by glide. I don't recommend updating vendor directory directly. Please update the package dependencies using glide...

rajatchopra:
@gab-satchi Did you run glide update with the change? I am suspecting if we are missing some updated files here.

gab-satchi:
@rajatchopra I deleted the dependency from `/vendor` and ran `glide install` and it brought in dependency properly again. `glide update` is failing for me. I've tried clearing the glide cache and removing the vendor directory but didn't work.

benmoss:
I think this was fixed by #1195

PR #1181, opened 2019-09-20
Support Kubernetes 1.16 and Linkerd

kellybirr:
## Description
added "cniVersion": "0.2.0" to make it work in Kubernetes 1.16
added NET_RAW support to PodSecurityPolicy to support Linkerd proxy injection

## Release Note
I have only tested this on two clusters built on Kubernetes 1.16.  It's worked for me in all cases.  Linkerd fix was tested with edge 19.9.3 (stable 2.5 does not work on Kubernetes 1.16)

I'm not sure if this is a "proper" fix, this is just what worked for me (twice).


zhangguanzhang:
@rajatchopra

PR #1194, opened 2019-09-30
Bump deps

### LINKED ISSUE #1193: Bug in gonetsh / go-powershell causing goroutines to deadlock
There's a bug in the go-powershell library that the Windows backends rely on to sync routes that causes the goroutines they spin up to never return or finish. Just creating a bug here to track the progress of fixing it upstream.

## Expected Behavior
Each node syncs the routes of all other nodes

## Current Behavior
The bug seems to be pretty non-deterministic, but the typical symptom is that one node will not have routes to another node. In the logs you will not see the "Subnet added" log lines for that node.

## Possible Solution
- Fixing gonetsh to use my fork of go-powershell, PR open [here](https://github.com/rakelkar/gonetsh/pull/11)
- Forking gonetsh if @rakelkar doesn't respond to my issue :smile: 

Unfortunately it's non-deterministic, so reproducing at the Flannel level is difficult. Basically any invocation of a method from the `netroute` [interface](https://github.com/rakelkar/gonetsh/blob/501daadcadf88660f61cc3352bdcb26b5f8ba4ce/netroute/netroute.go#L18-L29) will run in a Powershell instance that is subject to this bug.

## Your Environment
* Flannel version: latest
* Backend used (e.g. vxlan or udp): vxlan & host-gw
* Operating System and version: Windows

benmoss:
So I am really just trying to fix #1193, but it seems like the Glide deps are woefully out of date, so just trying to run a simple `make update-glide` ends up doing all of this. I think several people have added deps without running `update-glide` which prunes a lot of code.

This ends up superceding the work in #1180.

## Description
Fix #1193 by bumping gonetsh and go-powershell

## Release Note
Not clear if we want to have a release note for all the packages that were bumped.


benmoss:
Let's hold for #1195 to be merged first

PR #1161, opened 2019-07-15
Replaces gorillalabs go-powershell with bhendo/go-powershell

professor:
Fixes #1120

Authored-by: Todd Sedano <tsedano@pivotal.io>

## Description
The repository https://github.com/gorillalabs/go-powershell/ no longer exists and has been moved to https://github.com/bhendo/go-powershell

This PR fixes the glide dependency and vendor directory.

```release-note
None required
```

PR #1195, opened 2019-10-01
Fix glide

benmoss:
People have been modifying vendor and glide.lock not using the glide tool, leaving it in a weird state. This adds a new make target `verify-glide`, that just runs `glide list` and checks if there are any errors emitted. Making it a prerequisite of `test` means we'll get CI to run it on new PRs.

The only meaningful change here is to bump github.com/rakelkar/gonetsh to https://github.com/rakelkar/gonetsh/commit/501daadcadf88660f61cc3352bdcb26b5f8ba4ce. This looks like it was improperly done in #1161, instead of bumping gonetsh the vendored code was just modified directly.

Should mean #1194 can be much more straightforward once this is through :smile:


rajatchopra:
/lgtm
@arasananbu  Can you give it a quick glance?

daxmc99:
Is there a reason that flannel cannot move to go modules instead?

benmoss:
I think we should, I just wanted to get Glide fixed first so we could bump a single dependency. When I explored switching to gomod it resulted in even more massive amounts of dependency changes.

PR #1200, opened 2019-10-07
Bump gonetsh

### LINKED ISSUE #1193: Bug in gonetsh / go-powershell causing goroutines to deadlock

benmoss:
## Description
Fixes #1193. Built on top of #1195, I decided to split that out since it's a fairly massive commit (deleting dead vendor code). Once that is merged this'll be cleaner, but it basically just bumps gonetsh. Had to still do some glide surgery to bump a single package, but I think it all checks out.

## Release Note
```release-note
Bumped rakelkar/gonetsh to fix a bug where subnet synchronization goroutines would hang indefinitely
```


benmoss:
Now a more reasonable change set @rajatchopra

glide.lock
```diff
@@ -35,6 +35,11 @@ imports:
   - service/sts
 - name: github.com/buger/jsonparser
   version: f4dd9f5a6b441265aefc1d44872a6f8c10f42b37
+- name: github.com/benmoss/go-powershell
```
daxmc99: Thanks for picking this up @benmoss.

vendor/github.com/benmoss/go-powershell/shell.go
```diff
@@ -104,12 +104,12 @@ func streamReader(stream io.Reader, boundary string, buffer *string, signal *syn
...
 		output = output + string(buf[:read])
 
-		if strings.HasSuffix(output, marker) {
+		if marker.MatchString(output) {
 			break
 		}
 	}
 
-	*buffer = strings.TrimSuffix(output, marker)
+	*buffer = marker.FindStringSubmatch(output)[1]
```
rajatchopra: Quick scan.. but this does not seem to be boundary condition safe. If the regex does not match ever, then? TrimSuffix was safe in this regard, unless I missed the conviction that that regex will always match.

benmoss: This library is hot garbage imho, but it's what gonetsh chose to use to avoid the cost of starting up a new Powershell for each command that it needs to run.

The logic of the library is:
- Start one shell per `powershell.Shell` instance constructed
- Each time `Execute` is called, run the user's command, but append "boundary" tokens on both the stdout and stderr streams to the command
- Buffer the output until those boundary tokens to appear in both streams, then return the output

The problem that I was seeing, and never _fully_ understood, was that sometimes the library would never see the boundary tokens and would be stuck reading from the streams forever. I was able to reproduce this by just [running commands on a single shell as quickly as possible](https://github.com/benmoss/go-powershell/blob/master/shell_test.go#L56-L76). The bug seemed to be that while normally the library assumes you'll see:

```
some output from your command\r\n$gorillabd7478aeea3d4de2d7c0eee9$\r\n
```

sometimes it gets:

```
some output from your command\r\n$gorillabd7478aeea3d4de2d7c0eee9$\r\n\r\n\r\n
```

This results in the `strings.HasSuffix` check the library used to do never returning true, and a permanently stuck read on the output stream.

I never could figure out why/where the extra newlines were coming from. My only guess is it's some deep Powershell bug. The test that reproduces it only successfully reproduces it with complex commands like `ls`, simple stuff like `echo hi` will not trigger it.

So the simplest solution I saw was just to use a regex to test for the boundary instead of `strings.HasSuffix`, ignoring any subsequent characters after we see the boundary marker.

Does that make sense and address your concern? The only way I can see this still never returning would be if the boundary marker is *never* in the output stream, and I'm not sure how that would happen.

benmoss: While exploring this bug again to explain it here I tried a `bufio.Scanner` based solution: https://github.com/benmoss/go-powershell/commit/9d93b2c4ad8c390eb38a5195a8cb4c8d7448b3c1

Seems to work, it's basically `bufio.ScanLines` but where it treats the boundary marker as a `bufio.ErrFinalToken`. It seems cleaner, but was hesitant to change the library so significantly.

rajatchopra: Do you mind just checking on ArrayBoundRead for that line? Let's not assume that FindStringSubmatch will have a valid array with at least two indices.

benmoss:
Closing in favor of #1209

PR #1215, opened 2019-10-30
Use remote VNI for Windows

adelina-t:
Windows currently doesn't support any other VNI that 4096. The VNI for the remote should be retrieved from the subnet lease and set in the Policies of the vxlan HNS-Network. However, on Linu, flannel does not include the VNI in the Lease Attributes.

### DIFF
--- a/backend/vxlan/vxlan_network_windows.go
+++ b/backend/vxlan/vxlan_network_windows.go
@@ -101,6 +101,11 @@ func (nw *network) handleSubnetEvents(batch []subnet.Event) {
 			continue
 		}
 
+		if vxlanAttrs.VNI < 4096 {
+			log.Error("VNI is required to greater than or equal to 4096 on Windows.")
+			continue
+		}
+
 		hnsnetwork, err := hcn.GetNetworkByName(nw.dev.link.Name)
 		if err != nil {
 			log.Errorf("Unable to find network %v, error: %v", nw.dev.link.Name, err)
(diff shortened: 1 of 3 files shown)


zawachte-msft:
@ksubrmnn

backend/vxlan/vxlan_network_windows.go
```diff
@@ -109,7 +109,7 @@ func (nw *network) handleSubnetEvents(batch []subnet.Event) {
 		managementIp := event.Lease.Attrs.PublicIP.String()
 
 		networkPolicySettings := hcn.RemoteSubnetRoutePolicySetting{
-			IsolationId:                 4096,
+			IsolationId:                 vxlanAttrs.VNI,
```
ksubrmnn: Thanks for the PR! This is one of those lingering tasks we hadn't gotten around to.

Can you just add a check to make sure the VNI is 4096 or greater? That's a windows restriction :) Otherwise lgtm

ksubrmnn:
@adelina-t  What's the status of this?

adelina-t:
> @adelina-t What's the status of this?

@ksubrmnn  Pretty much waiting. I've added the check for VNI > 4096 some time ago.

ksubrmnn:
@adelina-t can you update the title to not include WIP?

@rajatchopra to merge. I say LGTM

@JocelynBerrendonner as an FYI. We should probably add an owner from Keith's team when the time is right. No rush though.

adelina-t:
@ksubrmnn removed. Didn't even notice it was still WIP there :)

JocelynBerrendonner:
LGTM

ksubrmnn:
@rajatchopra LGTM. Please merge when you get a chance :)

PR #1264, opened 2020-03-10
Use iptables-restore to preserve order with MASQ & FORWARD rules

ryarnyah:
Use iptables-restore to preserve order with MASQ & FORWARD rules.

## Description
When anything happen when delete/append iptables rules, sometimes you can have your rules out of order.This PR use only iptables-restore to guaranty order in delete/append rules.

This close https://github.com/coreos/flannel/issues/1261 & https://github.com/coreos/flannel/issues/1146

## Todos
- [X] Tests
- [x] Documentation (Needed?)
- [X] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dist/functional-test-k8s.sh
+++ b/dist/functional-test-k8s.sh
@@ -89,7 +89,7 @@ start_flannel() {
         $FLANNEL_DOCKER_IMAGE \
         -c "mkdir -p /etc/kube-flannel && \
             echo '$flannel_conf' > /etc/kube-flannel/net-conf.json && \
-            /opt/bin/flanneld --kube-subnet-mgr --kube-api-url $k8s_endpt" >/dev/null
+            /opt/bin/flanneld --kube-subnet-mgr --ip-masq --kube-api-url $k8s_endpt" >/dev/null
 
        while ! docker exec flannel-e2e-test-flannel$host_num ls /run/flannel/subnet.env >/dev/null 2>&1; do
          status=$(docker inspect --format='{{.State.Status}}' flannel-e2e-test-flannel$host_num)
@@ -121,6 +121,7 @@ test_vxlan() {
     start_flannel vxlan
     create_ping_dest # creates ping_dest1 and ping_dest2 variables
     pings
+    check_iptables
 }
 
 if [[ ${ARCH} == "amd64" ]]; then
@@ -128,6 +129,7 @@ test_udp() {
     start_flannel udp
     create_ping_dest # creates ping_dest1 and ping_dest2 variables
     pings
+    check_iptables
 }
 fi
 
@@ -135,12 +137,14 @@ test_host-gw() {
     start_flannel host-gw
     create_ping_dest # creates ping_dest1 and ping_dest2 variables
     pings
+    check_iptables
 }
 
 test_ipip() {
     start_flannel ipip
     create_ping_dest # creates ping_dest1 and ping_dest2 variables
     pings
+    check_iptables
 }
 
 test_public-ip-overwrite(){
@@ -166,6 +170,37 @@ pings() {
     # ping in both directions
 	assert "docker exec --privileged flannel-e2e-test-flannel1 /bin/ping -c 5 $ping_dest2" "Host 1 cannot ping host 2"
 	assert "docker exec --privileged flannel-e2e-test-flannel2 /bin/ping -c 5 $ping_dest1" "Host 2 cannot ping host 1"
+}
+
+check_iptables() {
+  read -r -d '' POSTROUTING_RULES_FLANNEL1 << EOM
+-P POSTROUTING ACCEPT
+-A POSTROUTING -s 10.10.0.0/16 -d 10.10.0.0/16 -m comment --comment "flanneld masq" -j RETURN
+-A POSTROUTING -s 10.10.0.0/16 ! -d 224.0.0.0/4 -m comment --comment "flanneld masq" -j MASQUERADE --random-fully
+-A POSTROUTING ! -s 10.10.0.0/16 -d 10.10.1.0/24 -m comment --comment "flanneld masq" -j RETURN
+-A POSTROUTING ! -s 10.10.0.0/16 -d 10.10.0.0/16 -m comment --comment "flanneld masq" -j MASQUERADE --random-fully
+EOM
+  read -r -d '' POSTROUTING_RULES_FLANNEL2 << EOM
+-P POSTROUTING ACCEPT
+-A POSTROUTING -s 10.10.0.0/16 -d 10.10.0.0/16 -m comment --comment "flanneld masq" -j RETURN
+-A POSTROUTING -s 10.10.0.0/16 ! -d 224.0.0.0/4 -m comment --comment "flanneld masq" -j MASQUERADE --random-fully
+-A POSTROUTING ! -s 10.10.0.0/16 -d 10.10.2.0/24 -m comment --comment "flanneld masq" -j RETURN
+-A POSTROUTING ! -s 10.10.0.0/16 -d 10.10.0.0/16 -m comment --comment "flanneld masq" -j MASQUERADE --random-fully
+EOM
+  read -r -d '' FORWARD_RULES << EOM
+-P FORWARD ACCEPT
+-A FORWARD -s 10.10.0.0/16 -m comment --comment "flanneld forward" -j ACCEPT
+-A FORWARD -d 10.10.0.0/16 -m comment --comment "flanneld forward" -j ACCEPT
+EOM
+  # check masquerade & forward rules
+  assert_equals "$POSTROUTING_RULES_FLANNEL1" \
+                "$(docker exec --privileged flannel-e2e-test-flannel1 /sbin/iptables -t nat -S POSTROUTING)" "Host 1 has not expected postrouting rules"
+  assert_equals "$POSTROUTING_RULES_FLANNEL2" \
+                "$(docker exec --privileged flannel-e2e-test-flannel2 /sbin/iptables -t nat -S POSTROUTING)" "Host 2 has not expected postrouting rules"
+  assert_equals "$FORWARD_RULES" \
+                "$(docker exec --privileged flannel-e2e-test-flannel1 /sbin/iptables -t filter -S FORWARD)" "Host 1 has not expected forward rules"
+  assert_equals "$FORWARD_RULES" \
+                "$(docker exec --privileged flannel-e2e-test-flannel2 /sbin/iptables -t filter -S FORWARD)" "Host 2 has not expected forward rules"
 }
 
 test_manifest() {
--- a/network/iptables.go
+++ b/network/iptables.go
@@ -18,7 +18,6 @@ package network
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/coreos/go-iptables/iptables"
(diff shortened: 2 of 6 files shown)


rajatchopra:
Looks good. How can we have an end-to-end test for it?
Maybe check on iptables rules after doing a kube e2e in dist/function-test-k8s.sh?

ryarnyah:
e2e tests has been added as requested

ryarnyah:
Do you need something else to have this PR merged?

luthermonson:
@Oats87 take a look would you?

luthermonson:
@Oats87 sign off on this so we can merge it, youre the iptables ninja

Oats87:
@ryarnyah would you mind rebasing this PR (or just fixing the merge conflict)?

ryarnyah:
@Oats87 i will look to rebase it this weekend if i have time ;)

ryarnyah:
@Oats87 It's seem to be good for review now.

network/iptables.go
```diff
@@ -100,21 +99,74 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
...
+	}
+
+	log.V(6).Infof("trying to run iptable-restore < %+v", tablesRules)
+
+	err = iptRestore.ApplyPartial(tablesRules)
+	if err != nil {
+		return fmt.Errorf("failed to apply partial iptables-restore %v", err)
+	}
+
+	log.Infof("bootstrap done")
```
Oats87, line 138: Can we provide some more context in this error message? Does this have to be done at the info level?

network/iptables.go
```diff
@@ -130,11 +182,21 @@ func DeleteIPTables(rules []IPTablesRule) error {
 		log.Errorf("Failed to setup IPTables. iptables binary was not found: %v", err)
 		return err
 	}
-	teardownIPTables(ipt, rules)
+	iptRestore, err := NewIPTablesRestore()
+	if err != nil {
+		// if we can't find iptables, give up and return
+		log.Errorf("Failed to setup IPTables-restore: %v", err)
```
Oats87, line 188: nit: this should probably be `iptables-restore` instead of `IPTables-restore`

dist/functional-test-k8s.sh
```diff
@@ -83,7 +83,7 @@ start_flannel() {
        docker run -e NODE_NAME=flannel$host_num --privileged --name flannel-e2e-test-flannel$host_num -id --entrypoint /bin/sh $FLANNEL_DOCKER_IMAGE >/dev/null
        docker exec flannel-e2e-test-flannel$host_num /bin/sh -c 'mkdir -p /etc/kube-flannel'
        echo $flannel_conf | docker exec -i flannel-e2e-test-flannel$host_num /bin/sh -c 'cat > /etc/kube-flannel/net-conf.json'
-       docker exec -d flannel-e2e-test-flannel$host_num /opt/bin/flanneld --kube-subnet-mgr --kube-api-url $k8s_endpt
+       docker exec -d flannel-e2e-test-flannel$host_num /opt/bin/flanneld --kube-subnet-mgr --ip-masq --kube-api-url $k8s_endpt
```
Oats87, line 86: Do we actually need this `--ip-masq`?

dist/functional-test-k8s.sh
```diff
@@ -148,6 +152,37 @@ pings() {
 	assert "docker exec --privileged flannel-e2e-test-flannel2 /bin/ping -c 5 $ping_dest1" "Host 2 cannot ping host 1"
 }
 
+check_iptables() {
```
Oats87, line 155: Should we maybe have a specific test where we explicitly garble rule order to ensure that `flanneld` corrects the ordering of the rules?

network/iptables.go
```diff
@@ -100,21 +99,74 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
 	return true, nil
 }
 
+func ipTablesCleanAndBuild(ipt IPTables, rules []IPTablesRule) (IPTablesRestoreRules, error) {
+	tablesRules := IPTablesRestoreRules{}
+
+	// Build delete rules
```
Oats87, line 105: Is there any way to ensure the order is wrong before just automatically deleting/creating the rules?

network/iptables.go
```diff
@@ -100,21 +99,74 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
...
+	}
+
+	return tablesRules, nil
+}
+
+func ipTablesBootstrap(ipt IPTables, iptRestore IPTablesRestore, rules []IPTablesRule) error {
+	tablesRules, err := ipTablesCleanAndBuild(ipt, rules)
+	if err != nil {
+		// if we can't find iptables, give up and return
+		return fmt.Errorf("Failed to setup IPTables-restore payload: %v", err)
```
Oats87, line 128: nit: should be `iptables-restore` not `IPTables-restore`

network/iptables.go
```diff
@@ -100,21 +99,74 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
...
+	}
+
+	log.V(6).Infof("trying to run iptable-restore < %+v", tablesRules)
```
Oats87, line 131: nit: `iptables-restore` not `iptable-restore`

network/iptables.go
```diff
@@ -100,21 +99,74 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
...
 	ipt, err := iptables.New()
 	if err != nil {
 		// if we can't find iptables, give up and return
 		log.Errorf("Failed to setup IPTables. iptables binary was not found: %v", err)
 		return
 	}
+	iptRestore, err := NewIPTablesRestore()
+	if err != nil {
+		// if we can't find iptables, give up and return
+		log.Errorf("Failed to setup IPTables. iptables binary was not found: %v", err)
```
Oats87, line 153: Isn't the problem that the `iptables-restore` binary was not found?

network/iptables_restore.go
```diff
@@ -0,0 +1,213 @@
+// Copyright 2015 flannel authors
```
Oats87, line 1: This probably actually belongs in `coreos/go-iptables` but having it here should be fine.

network/iptables_restore.go
```diff
@@ -0,0 +1,213 @@
...
+	ipTablesRestoreCmd  string = "iptables-restore"
+	ip6TablesRestoreCmd string = "ip6tables-restore"
+)
+
+// IPTablesRestore wrapper for iptables-restore
+type IPTablesRestore interface {
+	// ApplyFully apply all rules and flush chains
+	ApplyFully(rules IPTablesRestoreRules) error
+	// ApplyPartial apply without flush chains
+	ApplyPartial(rules IPTablesRestoreRules) error
```
Oats87, line 41: Maybe call this `ApplyNoFlush` or `ApplyWithoutFlush`?

Oats87 (changes requested):

network/iptables.go
briandowns: nit: error messages shouldn't be capitalized

louiznk:
> @ryarnyah could you rebase and respond to the review comments?

Hello @rajatchopra if you and @ryarnyah are agree I can do this for @ryarnyah 
Thanks

luthermonson:
@louiznk please do

network/iptables_restore.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// ipTablesRestore internal type
+type ipTablesRestore struct {
+	path     string
+	protocol iptables.Protocol
```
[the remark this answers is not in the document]
louiznk: :ok_hand:

luthermonson:
@louiznk any chance you can answer/fix for @Oats87 feedback? can probably wrap this up and merge if those are addressed

network/iptables_restore.go
```diff
@@ -0,0 +1,213 @@
...
+// IPTablesRestore wrapper for iptables-restore
+type IPTablesRestore interface {
+	// ApplyFully apply all rules and flush chains
+	ApplyFully(rules IPTablesRestoreRules) error
+	// ApplyPartial apply without flush chains
+	ApplyPartial(rules IPTablesRestoreRules) error
+}
+
+// ipTablesRestore internal type
+type ipTablesRestore struct {
```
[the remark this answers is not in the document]
louiznk: @Oats87 I'm not sur to understood, can you give me more details ?

network/iptables_restore.go
```diff
@@ -0,0 +1,213 @@
...
+		return false, err
+	}
+	v1, v2, v3, err := extractIptablesRestoreVersion(version)
+	if err != nil {
+		return false, err
+	}
+	return ipTablesHasWaitSupport(v1, v2, v3), nil
+}
+
+// Checks if an iptables-restore version is after 1.4.20, when --wait was added
```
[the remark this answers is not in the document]
louiznk: :ok_hand:

network/iptables.go
louiznk: I'm not sure it's possible

network/iptables.go
louiznk: :ok_hand:

network/iptables.go
louiznk: :ok_hand:

network/iptables.go
louiznk: :ok_hand:

network/iptables.go
louiznk: :ok_hand:

network/iptables_restore.go
louiznk: :ok:

dist/functional-test-k8s.sh
louiznk: Without this somes tests fails (test_hors-gw, test_ipip, test_udp, test_vxlan).

network/iptables.go
louiznk: I think it's possible if we have before memorized the order, but I'm not sure we can in every case memorize the order. And we will need to store this state outside the memory (ie: etcd) for the crash case. 
So I'm not sure it's a real benefit compared to the complexity it's will add.

dist/functional-test-k8s.sh
louiznk: I need to check how can I do this. Do you have an idea ?

network/iptables_restore.go
louiznk: So I leave it that way ?

louiznk:
> was requested for review

Hello @luthermonson 
I don't know if you have the opportunity to see the commit and question/reply for @Oats87 feedback?
Thanks a lot for your help and review.

louiznk (commented):
Apply requests

louiznk (approved):

manuelbuil:
Are you still interested in merging this? Could you rebase so that we can review it again? Thanks!

louiznk:
Hello,

Thanks, I can do that next weekend.


Le jeu. 26 mai 2022 à 17:57, Manuel Buil ***@***.***> a
écrit :

louiznk:
> Are you still interested in merging this? Could you rebase so that we can review it again? Thanks!

Hello @manuelbuil ,

Thanks, I can do that this week end

louiznk:
> Are you still interested in merging this? Could you rebase so that we can review it again? Thanks!

Hello @manuelbuil , the rebase is done, I fixe some trouble with this new versions (usage of comments on iptables rules for exemples)
Thanks for your review and feedback

dist/functional-test-k8s.sh
louiznk: As I know with iptables-restore the order is garanty in the transaction and the `--wait` option prevent multiple instances of the program from running concurrently (so no concurrently changes on the iptables rules)

manuelbuil (commented):
IIUC, some users detected a race condition in which these rules: https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L57-L66, are applied in different order. In my opinion, the simplest solution would be adding a lock here https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L30 and then check the lock around here: https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L244 and not allow any other rule to be added until the lock is released. That way the race condition will not exist anymore and the problem would be fixed, or? Using iptables-restore also fixes the problem but it requires adding ~500 lines of code, which makes me wonder if this is an overkill given that the use of iptables in flannel is minimal (if we had some sort of network policy engine using iptables, I would probably think differently)

network/iptables_restore.go
```diff
@@ -0,0 +1,226 @@
...
+
+// IPTablesRestoreRules represents iptables-restore table block
+type IPTablesRestoreRules map[string][]IPTablesRestoreRuleSpec
+
+// IPTablesRestoreRuleSpec represents one rule spec delimited by space
+type IPTablesRestoreRuleSpec []string
+
+// NewIPTablesRestore build new IPTablesRestore for IPv4
+func NewIPTablesRestore() (IPTablesRestore, error) {
+	return NewIPTablesRestoreWithProtocol(iptables.ProtocolIPv4)
```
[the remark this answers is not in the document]
louiznk: > IIUC, some users detected a race condition in which these rules: https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L57-L66, are applied in different order. In my opinion, the simplest solution would be adding a lock here https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L30 and then check the lock around here: https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L244 and not allow any other rule to be added until the lock is released. That way the race condition will not exist anymore and the problem would be fixed, or? Using iptables-restore also fixes the problem but it requires adding ~500 lines of code, which makes me wonder if this is an overkill given that the use of iptables in flannel is minimal (if we had some sort of network policy engine using iptables, I would probably think differently).
> 
> I'd love to hear more opinions!
You're right in the case of flannel alone. But I have trouble with some concurrent access from other programs (in production).
In my case it was kube-router (using for network policy).
I know there are better solutions than using flannel + kube-routeur, but unfortunately not in special my case (some external requirements).
So the main advantage for me with iptables-restore is the "transaction" that allows to preserve the order when you have some concurrents access to iptables.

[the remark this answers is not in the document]
louiznk: > Is there a reason to hardcode IPv4?

Just for respecting the same convention as implementation based on iptables (New => ipv4 by default and NewWithProtocol => protocol in parameter). For exemple see : 
https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L190 and https://github.com/flannel-io/flannel/blob/master/network/iptables.go#L206
I can change it if it's better

network/iptables.go
```diff
@@ -135,24 +134,77 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
...
+			tablesRules[rule.table] = append(tablesRules[rule.table], append(IPTablesRestoreRuleSpec{"-D", rule.chain}, rule.rulespec...))
+		}
+		tablesRules[rule.table] = append(tablesRules[rule.table], append(IPTablesRestoreRuleSpec{"-A", rule.chain}, rule.rulespec...))
+	}
+
+	return tablesRules, nil
+}
+
+func ipTablesBootstrap(ipt IPTables, iptRestore IPTablesRestore, rules []IPTablesRule) error {
+	tablesRules, err := ipTablesCleanAndBuild(ipt, rules)
```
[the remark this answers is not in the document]
louiznk: I will do this, thanks

network/iptables.go
```diff
@@ -135,24 +134,77 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
 	return true, nil
 }
 
+func ipTablesCleanAndBuild(ipt IPTables, rules []IPTablesRule) (IPTablesRestoreRules, error) {
+	tablesRules := IPTablesRestoreRules{}
```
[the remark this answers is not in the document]
louiznk: I will do this, thanks

network/iptables.go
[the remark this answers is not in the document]
louiznk: done @manuelbuil

network/iptables.go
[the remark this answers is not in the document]
louiznk: done @manuelbuil

network/iptables_restore.go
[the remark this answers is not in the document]
manuelbuil: I personally would prefer to use the same functions for both ipv4 and ipv6 and change the arg accordingly. Therefore, I'd prefer `NewIPTablesRestoreWithProtocol(iptables.ProtocolIPv6)` and `NewIPTablesRestoreWithProtocol(iptables.ProtocolIPv4)` in the code. It feels to me more confusing if we use different functions for each protocol

manuelbuil:
Tested and works

manuelbuil (approved):

PR #1265, opened 2020-03-12
Spruce up for release 12.0

rajatchopra:
## Description

- Clean up the test harness to run latest kube versions (both pre 1.16 and post 1.16)
- Changes to etcd tests to be compatible with hosts that have DNS search domains
- Modify the Documentation/kube-flannel.yml so that latest release images are matched

## Release Note

```release-note
None required
```


rajatchopra:
@arasananbu PTAL

/cc @tomdee @dcbw @squeed

Issue #1279, opened 2020-04-07
UDP: bad checksum on VXLAN interface


dmitry-irtegov:
On k8s 1.17 cluster with RHeL 7 nodes, service IPs for pods on other nodes are not accessible. 
Pod IP seem to work fine. Most noticeably, CoreDNS does not work.
Target node dmesg is filled with messages like:
```
[ 1423.035722] UDP: bad checksum. From 172.16.9.99:27503 to 172.16.80.252:8472 ulen 75
[ 1426.036873] UDP: bad checksum. From 172.16.9.99:4894 to 172.16.80.252:8472 ulen 75
[ 1427.537392] UDP: bad checksum. From 172.16.9.99:32693 to 172.16.80.252:8472 ulen 75
[ 1429.037910] UDP: bad checksum. From 172.16.9.99:46133 to 172.16.80.252:8472 ulen 75
```
Turning IP checksum offloading on the flannel.1 interface fixes the issue:
```
[root@ip-172-16-102-241 ~]# nslookup www.google.com 100.64.0.10
;; connection timed out; no servers could be reached
[root@ip-172-16-102-241 ~]# ethtool -K flannel.1 tx-checksum-ip-generic off
Actual changes:
tx-checksumming: off
	tx-checksum-ip-generic: off
tcp-segmentation-offload: off
	tx-tcp-segmentation: off [requested on]
	tx-tcp-ecn-segmentation: off [requested on]
	tx-tcp6-segmentation: off [requested on]
	tx-tcp-mangleid-segmentation: off [requested on]
udp-fragmentation-offload: off [requested on]
[root@ip-172-16-102-241 ~]# nslookup www.google.com 100.64.0.10
Server:		100.64.0.10
Address:	100.64.0.10#53
Non-authoritative answer:
Name:	www.google.com
Address: 172.217.13.228
Name:	www.google.com
Address: 2607:f8b0:4004:80a::2004
[root@ip-172-16-102-241 ~]# 
```
Other people also hit this: https://t.du9l.com/2020/03/kubernetes-flannel-udp-packets-dropped-for-wrong-checksum-workaround/

This happens both with cni-canal and pure cni-flannel, so we decided to report the issue here.

## Expected Behavior
I do not have to adjust interface settings to get flannel to work.

## Your Environment

* Flannel version: 0.11.0
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version:  3.4.3
* Kubernetes version (if used): 1.17.4
[...]

Issue #1243, opened 2020-01-13
ClusterIP services not accessible when using flannel CNI from host machines in Kubernetes


nonsense:
I am trying to access a Kubernetes service through its `ClusterIP`, from a `pod` that is attached to its `host`'s network and has access to DNS, with:

```
  hostNetwork: true
  dnsPolicy: ClusterFirstWithHostNet
```

However the host machine has no ip routes setup for the service CIDR, for example

```
➜  ~ k get services
NAME             TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
kubernetes       ClusterIP   100.64.0.1      <none>        443/TCP    25m
redis-headless   ClusterIP   None            <none>        6379/TCP   19m
redis-master     ClusterIP   100.64.63.204   <none>        6379/TCP   19m
```

```
➜  ~ k get pods -o wide
NAME                       READY   STATUS      RESTARTS   AGE   IP              NODE                                             NOMINATED NODE   READINESS GATES
redis-master-0             1/1     Running     0          18m   100.96.1.3      ip-172-20-39-241.eu-central-1.compute.internal   <none>           <none>
```

```
root@ip-172-20-39-241:/home/admin# ip route
default via 172.20.32.1 dev eth0
10.32.0.0/12 dev weave proto kernel scope link src 10.46.0.0
100.96.0.0/24 via 100.96.0.0 dev flannel.11 onlink
100.96.1.0/24 dev cni0 proto kernel scope link src 100.96.1.1
100.96.2.0/24 via 100.96.2.0 dev flannel.11 onlink
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown
172.20.32.0/19 dev eth0 proto kernel scope link src 172.20.39.241
```

## Expected Behavior
I expect that I should be able to reach services running on Kubernetes from the host machines, but I can only access `headless` services - those that return a pod IP.

The pod CIDR has `ip route`s setup, but the services CIDR doesn't.

## Current Behavior
Can't access services through their `ClusterIP`s from host network.

## Possible Solution
If I manually add an `ip route` to `100.64.0.0/16` via `100.96.1.1`, `ClusterIP` are accessible. But this route is not there by default.

## Your Environment

* Flannel version: v0.11.0
* kops version: Version 1.17.0-alpha.1 (git-501baf7e5)
* Backend used (e.g. vxlan or udp): vxlan
* Kubernetes version (if used): 
* Operating System and version:
[...]

PR #1282, opened 2020-04-13
Disable tx and rx offloading on VXLAN interfaces

aojea:
## Description

Seems that some kernel versions have issues with VXLAN checksum
offloading, causing that flannel stop to work on some scenarios
where the traffic is encapsulated, but the checksum is wrong and is
discarded by the receiver.

A known workaround that works is disabling offloading on the flannel
interface:

ethtool --offload flannel.1 rx off tx off

This PR force to disable it always on VXLAN interfaces

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
disable rx and tx checksum offloading on VXLAN interfaces
```

### DIFF
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -118,6 +118,11 @@ func (dev *vxlanDevice) Configure(ipn ip.IP4Net) error {
 		return fmt.Errorf("failed to set interface %s to UP state: %s", dev.link.Attrs().Name, err)
 	}
 
+	// TODO: Workaround to fix #1282
+	if err := ip.SetChecksumOffloading(dev.link.Attrs().Name, false, false); err != nil {
+		return fmt.Errorf("failed to disable interface %s tx and rx offloading: %s", dev.link.Attrs().Name, err)
+	}
+
 	return nil
 }
 
--- a/pkg/ip/iface.go
+++ b/pkg/ip/iface.go
@@ -22,9 +22,76 @@ import (
 	"fmt"
 	"net"
 	"syscall"
+	"unsafe"
 
 	"github.com/vishvananda/netlink"
 )
+
+const (
+	siocEthtool = 0x8946 // linux/sockios.h
+
+	// #define ETHTOOL_SRXCSUM		0x00000015 /* Set RX hw csum enable (ethtool_value) */
+	ethtoolSRxCsum = 0x00000015 // linux/ethtool.h
+	// #define ETHTOOL_STXCSUM		0x00000017 /* Set TX hw csum enable (ethtool_value) */
+	ethtoolSTxCsum = 0x00000017 // linux/ethtool.h
+
+	maxIfNameSize = 16 // linux/if.h
+)
+
+// linux/if.h 'struct ifreq'
+type ifreq struct {
+	Name [maxIfNameSize]byte
+	Data uintptr
+}
+
+// linux/ethtool.h 'struct ethtool_value'
+type ethtoolValue struct {
+	Cmd  uint32
+	Data uint32
+}
+
+// ethtool executes Linux ethtool syscall.
+func ethtool(iface string, cmd, val uint32) (retval uint32, err error) {
+	if len(iface)+1 > maxIfNameSize {
+		return 0, fmt.Errorf("interface name is too long")
+	}
+	socket, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_DGRAM, 0)
+	if err != nil {
+		return 0, err
+	}
+	defer syscall.Close(socket)
+
+	// prepare ethtool request
+	value := ethtoolValue{cmd, val}
+	request := ifreq{Data: uintptr(unsafe.Pointer(&value))}
+	copy(request.Name[:], iface)
+
+	// ioctl system call
+	_, _, errno := syscall.RawSyscall(syscall.SYS_IOCTL, uintptr(socket), uintptr(siocEthtool),
+		uintptr(unsafe.Pointer(&request)))
+	if errno != 0 {
+		return 0, errno
+	}
+	return value.Data, nil
+}
+
+// SetChecksumOffloading enables/disables Rx/Tx checksum offloading
+// for the given interface.
+func SetChecksumOffloading(ifName string, rxOn, txOn bool) error {
+	var rxVal, txVal uint32
+	if rxOn {
+		rxVal = 1
+	}
+	if txOn {
+		txVal = 1
+	}
+	_, err := ethtool(ifName, ethtoolSRxCsum, rxVal)
+	if err != nil {
+		return err
+	}
+	_, err = ethtool(ifName, ethtoolSTxCsum, txVal)
+	return err
+}
 
 func getIfaceAddrs(iface *net.Interface) ([]netlink.Addr, error) {
 	link := &netlink.Device{


aojea:
Fixes https://github.com/coreos/flannel/pull/1282

aojea:
ethtool code from https://github.com/ligato/vpp-agent/blob/master/plugins/linux/ifplugin/linuxcalls/ethtool_linuxcalls.go

Capitrium:
@aojea I just tested this on my dev cluster, still running into #1243. Seems like this might be more related to #1279, but I'm not seeing that on my cluster so I can't be sure.

aojea:
@Capitrium seems that other people that has tested the patch has it working, 
maybe you need to restart flannel so the interfaces are recreated and the workaround is applied?

aojea:
ping @rajatchopra :-)

Miouge1:
@aojea any idea on which kernel versions are affected?

aojea:
no idea, but this issue created an explosion of issues opened against kubernetes and related projects, it was also discussed in the sig-network mailing list https://groups.google.com/d/msg/kubernetes-sig-network/JxkTLd4M8WM/EW8O1E0PAgAJ
The information is scattered in multiple places sorry 😄

Capitrium:
@aojea Yeah, looks like it was an issue with configuration or stale pods on my end - rebuilt and redeployed, seems like it's working now.

I was still seeing networking issues with pods on one node after deploying the patch and had to kill the node, but I was doing a fair amount of testing with different kube-proxy/kube-router/flannel versions and probably broke something else in the process. Most existing nodes and all new nodes are working properly. :+1:

hakman:
Very cool @aojea. This create a lot of discussions in Kops project also. Really happy to see that there is a solution :).

rajatchopra:
@aojea I am not sure if we should disable rx/tx checksum always. A config parameter maybe? This is a temporary hack, right?

aojea:
> Who is tracking the 'real' fix in the kernel? @dcbw Any idea?

@dcbw posted more info here https://github.com/kubernetes/kubernetes/issues/88986#issuecomment-616906847
there is a patch upstream in the kernel

> @aojea I am not sure if we should disable rx/tx checksum always. A config parameter maybe? This is a temporary hack, right?

@rajatchopra I've considered that, but 

* weave has disabled it in all veth interfaces 4 years ago 
https://github.com/weaveworks/weave/issues/1255#issuecomment-221820171

* this is configured only on the vxlan interfaces, and vxlan RFC recommends to not set the UDP checksum
https://tools.ietf.org/html/rfc7348

>        -  UDP Checksum: It SHOULD be transmitted as zero.  When a packet
>          is received with a UDP checksum of zero, it MUST be accepted
>          for decapsulation.  Optionally, if the encapsulating end point
>          includes a non-zero UDP checksum, it MUST be correctly
>          calculated across the entire packet including the IP header,
>          UDP header, VXLAN header, and encapsulated MAC frame.  When a
>          decapsulating end point receives a packet with a non-zero
>          checksum, it MAY choose to verify the checksum value.  If it
>          chooses to perform such verification, and the verification
>          fails, the packet MUST be dropped.  If the decapsulating
>          destination chooses not to perform the verification, or
>          performs it successfully, the packet MUST be accepted for
>          decapsulation.

and despite we make it configurable we have to disable offloading by default or people will keep hitting the bug ... is it worth the effort?

rikatz:
> @aojea any idea on which kernel versions are affected?

Kernel 3.10.0 that's used in RHEL7/CentOS 7 with minor variations (but I've checked and this happens in CentOS from the 7.0 to 7.7 and with the latest RH provided kernel)

hakman:
@rajatchopra any plans for a bugfix release containing this?
The Kops 1.17 release it almost ready and would like to know if we should wait for this or pick it up in the next release. Thanks! :)
https://github.com/kubernetes/kops/pull/8614

hakman:
The "proper" fix for this [was accepted](http://patchwork.ozlabs.org/project/netfilter-devel/patch/335a95d93767f2b58ad89975e4a0b342ee00db91.1587429321.git.gnault@redhat.com/) by `netfilter`.

The comment says that "the problem seems to be there from day one, although it was probably not visible before UDP tunnels were implemented". This makes this fix the best solution for now for anyone wanting to upgrade to k8s 1.17 and use flannel.

zhangguanzhang:
Oddly enough, if I run a vxlan using a binary flannel file, it will work, whereas if I deploy it using kubeadm, it will run out in 63 seconds

aojea:
> Oddly enough, if I run a vxlan using a binary flannel file, it will work, whereas if I deploy it using kubeadm, the SVC access will timeout in 63 seconds

The root cause is the kernel bug mentioned in one of the issues (honestly I lost track of them :smile: )
I guess that in 1.17 k8s was introduced a trigger for the kernel bug that seems to affect CNIs using overlay, maybe the iptables --random flag, ... I didn't investigate more because the bug was acknowledged by the kernel community and the flannel workaround works, ... personally I don't see any reason to implement a workaround in k8s just for flannel, flannel can implement it, the same way that weave and others did

zhangguanzhang:
I did a control group and found the cause of the problem，

| deploy type|  flannel version  |   flannel is running in pod? |  will 63 sec delay? |
|          :----:            |        :----:     |             :----:           |  :----:  |
|        kubeadm             |       v0.11.0     |              yes             |    yes   |
|        kubeadm             |       v0.12.0     |              yes             |    yes   |
|        kubeadm             |       v0.11.0     |              no              |    yes   |
|        kubeadm             |       v0.12.0     |              no              |    yes   |
|        ansible             |       v0.11.0     |              yes             |    no    |
|        ansible             |       v0.12.0     |              yes             |    no    |
|        ansible             |       v0.11.0     |              no              |    no    |
|        ansible             |       v0.12.0     |              no              |    no    |
The control group above proved that flannel was not the source of the problem, so I made a second control group：

| type(kubeadm or ansible)   |  kube-proxy version  | kube-proxy is running in pod?  |   will 63 sec delay? |
|          :----:            |        :----:        |             :----:             |     :----:    |
|        kubeadm             |       v1.17.5        |               yes              |      yes      |
|        kubeadm             |       v1.17.5        |               no               |      no       |
|        kubeadm             |       v1.16.9        |               yes              |      no       |
|        kubeadm             |       v1.16.9        |               no              |       no      |
|        ansible             |       v1.17.5        |               yes              |      yes      |
|        ansible             |       v1.17.5        |               no               |      no       |


so，If the kube-proxy is running in pod, It will trigger the kernel bug,I found this by comparing submissions on github
https://github.com/kubernetes/kubernetes/commit/fed582333f639dc22e879f4bbb258e403c210c30
so I hack the docker images, there is the `Dockerfile`
```Dockerfile
FROM k8s.gcr.io/kube-proxy:v1.17.5
RUN rm -f /usr/sbin/iptables && 
    clean-install iptables
```
after I set the images use the hack images , it never causing 63 second delays in vxlan mode.

@danwinship Please take a look at why iptables is not installed in the docker image

aojea:
@zhangguanzhang impressive work

iptables is installed in that image, however, due to another bug, it has to use the latest version and use an script to detect if it should use the nft or legacy backend

```
# Install latest iptables package from buster-backports
RUN echo deb http://deb.debian.org/debian buster-backports main >> /etc/apt/sources.list; \
    apt-get update; \
    apt-get -t buster-backports -y --no-install-recommends install iptables
```

so, do you think that the iptables version is the trigger?

jhohertz:
I do. The recent release of kubernetes 1.16.10 is also affecting 1.16 for the first time so I studied the diffs yesterday. The only thing I see is that there is an iptables container build and the version was bumped from 11.x to 12.x, which 1.17 also did.

zhangguanzhang:
The problem, in terms of docker image modification, is with iptables

danwinship:
The iptables packaging bump triggered the `--random-fully` bug because the old (11.x) kube images had a version of iptables that was too old to support `--random-fully`, while the new (12.x) images, in addition to having the whole legacy-vs-nft wrapper thing, also have a much newer version of iptables that does support `--random-fully`. Unfortunately it turns out that just because the iptables binary supports it doesn't mean the kernel supports it. There's a PR to fix this. But anyway, this is a completely separate problem from the vxlan offload bug.

jhohertz:
Yet the workaround often suggested seems to be disabling offload, and it does have an effect on the issue

MarkRose:
Stable kernels 5.6.13, 5.4.41, 4.19.123, 4.14.181 and later have the checksum patch included.

zhangguanzhang:
@MarkRose Which jkernels in Centos?

jhohertz:
Trying to test a flatcar image w/ one of those kernels, but issues @ quay making for a fun day...

jhohertz:
Just going to summarise what I know about this for sure now:

1) kubernetes updating ip-tables to 12.x breaks it (nothing else networking related in the 1.16.10 diff)
2) The systemd workaround for disabling vxlan works around it in superficial testing
3) The kernel patch, now in the stable trees, does nothing to help matters (latest flatcar has a sufficiently recent kernel as of a couple days ago and was used to test.)

Which says, that while it's related to vxlan offload in some manner (because the workaround seems to work...), it is not that exact kernel bug (because it makes no difference when tested).

Looking to understand this random-fully thing now.

danwinship:
@jhohertz and just to confirm, you're running RHEL/CentOS 7 with kernel 3.10.something?

jhohertz:
Not here, I have been using coreos/flatcar container linux

rikatz:
@danwinship just as a follow up from our sig-network meeting, I'll post the tests results from disabling --random-fully scenarios in kube-proxy and flannel, and also the generated rules for each scenario.

This is the same environment I could use to confirm that disabling tx offload solves (CentOS 7 + Flannel 0.11), but in this case I'm enabling tx offload again in everything:

* Before each test, iptables rules are cleaned in the hosts so flannel and kube-proxy recreate them
* In NodePort scenarios, I'm calling my own IP:Port as part of the cluster to reach a Pod running outside of it

**Scenario 1 - Both kube-proxy and flannel with MASQUERADE rules created with --random-fully**

NodePort: Exactly 63s
ClusterIP: Exactly 63s

IPTables rules containing --random-fully:
```
-A POSTROUTING -s 10.244.0.0/16 ! -d 224.0.0.0/4 -j MASQUERADE --random-fully
-A POSTROUTING ! -s 10.244.0.0/16 -d 10.244.0.0/16 -j MASQUERADE --random-fully
-A KUBE-POSTROUTING -m comment --comment "kubernetes service traffic requiring SNAT" -m mark --mark 0x4000/0x4000 -j MASQUERADE --random-fully
```


**Scenario 2 - Original kube-proxy  with random-fully and flannel recompiled without it**
NodePort: Exactly 63s
ClusterIP: Exactly 63s

IPTables rules containing --random-fully:
```
-A KUBE-POSTROUTING -m comment --comment "kubernetes service traffic requiring SNAT" -m mark --mark 0x4000/0x4000 -j MASQUERADE --random-fully
```

**Scenario 3 -  kube-proxy  without random-fully and original flannel with random-fully enabled**
NodePort: 0.03s
ClusterIP: 0.007s

IPTables rules containing --random-fully:
```
-A POSTROUTING -s 10.244.0.0/16 ! -d 224.0.0.0/4 -j MASQUERADE --random-fully
-A POSTROUTING ! -s 10.244.0.0/16 -d 10.244.0.0/16 -j MASQUERADE --random-fully
```

**Scenario 4 -  kube-proxy and flannel without random-fully**
NodePort: 0.005s
ClusterIP: 0.03s

IPTables rules containing --random-fully: NONE

---
This way, we can almost for sure say that the insertion of --random-fully in the MASQUERADE rules triggered an already existing Kernel Bug in vxlan part :)

I'll post this also in the original issue at k/k repo, and please let me know if I can help with any further test

rikatz:
Also I've seen some discussion here about iptables-legacy x nft and this is starting to hit CentOS 8 users: https://github.com/kubernetes/kubernetes/issues/91331

I was going to ask what's the effort and gain of putting the [iptables-wrapper](https://github.com/kubernetes-sigs/iptables-wrappers/) to work here but as noted by the README it seems it wouldn't solve the CentOS 8 case :/

aojea:
/close

This is no longer needed thanks to @danwinship :clap: 
https://github.com/kubernetes/kubernetes/pull/92035

hakman:
@aojea will https://github.com/kubernetes/kubernetes/pull/92035 be cherry-picked to older k8s releases?

aojea:
> @aojea will [kubernetes/kubernetes#92035](https://github.com/kubernetes/kubernetes/pull/92035) be cherry-picked to older k8s releases?

it should be backported to the supported releases, just ping the author in the PR if he can do it, if he can not the process is described here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md

hakman:
Thanks @aojea and @danwinship 😄

Fluxay-666 (commented):
a

Fluxay-666:
Thanks, I used the latest version v0.21.5 plus this PR patch to successfully fix the 63-second connection delay problem.

PR #1283, opened 2020-04-17
Bugfix: VtepMac would be empty when lease re-acquire for windows without kubeSubnetMgr

cwdsuzhou:
## Description

ISSUE: https://github.com/coreos/flannel/issues/1280

When we deploy flannel `vxlan` on windows and do not use `kubeSubnetMgr`.  We found vtepMac would be reset to "". 

See https://github.com/coreos/flannel/blob/master/backend/vxlan/vxlan_windows.go#L134
VtepMac is nil

See https://github.com/coreos/flannel/blob/master/backend/vxlan/vxlan_windows.go#L139
Acquire lease

See https://github.com/coreos/flannel/blob/master/backend/vxlan/vxlan_windows.go#L161
Create Network

Although https://github.com/coreos/flannel/blob/master/backend/vxlan/vxlan_windows.go#L188, re-acquire the lease. but we have not reset it to the `network` and it returned(mac is nil).

Then, `MonitorLease` would renew lease periodically use the network(mac is nil)
```
	// Kube subnet mgr doesn't lease the subnet for this node - it just uses the podCidr that's already assigned.
	if !opts.kubeSubnetMgr {
		err = MonitorLease(ctx, sm, bn, &wg)
		if err == errInterrupted {
			// The lease was "revoked" - shut everything down
			cancel()
		}
	}
```

See

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Bugfix: VtepMac would be empty when lease re-acquire for windows without kubeSubnetMgr
```

### DIFF
--- a/backend/vxlan/vxlan_windows.go
+++ b/backend/vxlan/vxlan_windows.go
@@ -191,7 +191,10 @@ func (be *VXLANBackend) RegisterNetwork(ctx context.Context, wg sync.WaitGroup, 
 	}
 
 	lease, err = be.subnetMgr.AcquireLease(ctx, subnetAttrs)
-
+	if err != nil {
+		return nil, err
+	}
+	network.SubnetLease = lease
 	return network, nil
 }
 


cwdsuzhou:
@arasananbu PTAL

cwdsuzhou:
@rajatchopra PTAL

rajatchopra:
@ksubrmnn Can you review? Thanks.

backend/vxlan/vxlan_windows.go
```diff
@@ -191,7 +191,10 @@ func (be *VXLANBackend) RegisterNetwork(ctx context.Context, wg sync.WaitGroup,
 	}
 
 	lease, err = be.subnetMgr.AcquireLease(ctx, subnetAttrs)
```
rajatchopra, line 193: Why are we re-acquiring the lease anyway?

cwdsuzhou: From the codes, I can guess, the first time we acquire lease is to get `subnet`. But at that time mac is `nil`, so we need re-acquire the lease when the vxlan network is ready(mac is not nil)

cwdsuzhou:
ping @rajatchopra @ksubrmnn

backend/vxlan/vxlan_windows.go
ksubrmnn: Yep that is exactly why. We need to wait until the network is created to pass around the mac

ksubrmnn:
LGTM

ksubrmnn:
@JocelynBerrendonner FYI

JocelynBerrendonner:
Thanks for the heads up, Kalya. LGTM.

cwdsuzhou:
ping @rajatchopra

PR #1287, opened 2020-04-26
doc: Add section about creating new docker network

pehala:
Flannel doesn't need to replace deamon0 network in Docker and does work even if the you create new docker network that uses flannel.

## Description

Flannel when using with docker doesn't need to replace the default docker0 network, instead, it can be attached to a new docker network. I have added documentation for this type of workflow.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


luthermonson:
lgtm

luthermonson (approved):

PR #1299, opened 2020-06-07
Added a hint for firewall rules

seb-kw:
## Description

Completion / supplement of the documentation for the IPsec backend.
In readme: added a hint for firewall rules in backend documentation (IPsec backend).

Could help users to use this backend without problems.


rajatchopra:
@seb-kw May reword this and declare this as requirement? Instead of just a hint.

PR #1316, opened 2020-06-26
docs: set priorityClassName to system-node-critical

frederiko:
## Description
Flannel pods are not considered critical by the scheduler, which may
cause issues. In case of multiple pods are fighting for resources
there is the possibility of flannel pods be in pending mode indefinitely

## Release Note
```release-note
None required
```


frederiko:
Fixed `gofmt` that was 'silently' breaking the build

Oats87:
@frederiko Would you mind rebasing on master? We have already implemented the `go fmt`

frederiko:
@Oats87 Sure. Will do.

Issue #1272, opened 2020-03-24
Flanneld doesn't reconnect to the apiserver


angeloxx:
When flanneld 0.12 on Windows worker node lost the connection with the apiserver, it doesn't retry to connect but continues to log:

`reflector.go:201] github.com/coreos/flannel/subnet/kube/kube.go:307: Failed to list *v1.Node: Get https://10.128.0.12:6443/api/v1/nodes?resourceVersion=0: http2: no cached connection was `available`

## Expected Behavior
Flanneld should reconnect to the apiserver

## Steps to Reproduce (for bugs)
In our environment the easiest way to reproduce the issue is to move the floating IP to a different master node.

## Your Environment
* Flannel version: 0.12.0
* Startup Option: --kube-subnet-mgr --kubeconfig-file=<kubelet-file.conf>
* Backend used: vxlan
* Kubernetes version (if used): 1.16.3 on premise, 3 master nodes and we're using a floating IP (managed by keepalived) that is used by flanneld as backend
* Operating System and version: 
  * Master: Redhat Linux 7
  * Worker: Windows 2019 10.0.0.17763.934

DerrickMartinez:
I have reproduced this as well in 0.12.0. When rolling updating master nodes, it will cause this to go into panic

DerrickMartinez:
Any update on this issue? It's bad enough so 0.12.0 it's not production ready in a windows env. Thanks

luthermonson:
@DerrickMartinez you have any logs you can post? i have to look into something i think is related.

giskou:
I keep seeing this when flannel starts after the host has been restarted.
Not a windows/flannel expert, but from what I understand the HCN network is lost on reboots and this starts happening only if flannel recreates the HCN.
```
I0617 10:20:41.195352    5580 main.go:518] Determining IP address of default interface
I0617 10:20:41.804639    5580 main.go:531] Using interface with name Ethernet0 and address 172.16.13.197
I0617 10:20:41.804639    5580 main.go:548] Defaulting external address to interface address (172.16.13.197)
I0617 10:20:41.840550    5580 kube.go:119] Waiting 10m0s for node controller to sync
I0617 10:20:41.840550    5580 kube.go:306] Starting kube subnet manager
I0617 10:20:42.849946    5580 kube.go:126] Node controller sync successful
I0617 10:20:42.850188    5580 main.go:246] Created subnet manager: Kubernetes Subnet Manager - windows-worker-08
[...]

PR #1322, opened 2020-07-21
add crash when failure in list or watch of kubernetes api server

KnicKnic:
## Description

Address "Flanneld doesn't reconnect to the apiserver" - https://github.com/coreos/flannel/issues/1272

Add crash when failure in list or watch when talking to kubernetes api server. This should result in a retry loop for the connection (due to flannel being restarted).

Ideally the retry loop should be fully encapsulated inside flannel, however I didn't have much time to fix this problem. And it has been a month and no one has offered a better solution. I hope that this PR sparks that.

## Todos
~~- [ ] Tests~~
~~- [ ] Documentation~~
- [x] Release note

## Release Note

```release-note
Workaround for not reconnecting to api-server in windows
```

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -142,10 +142,18 @@ func newKubeSubnetManager(c clientset.Interface, sc *subnet.Config, nodeName, pr
 	indexer, controller := cache.NewIndexerInformer(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				return ksm.client.CoreV1().Nodes().List(options)
+				obj, err := ksm.client.CoreV1().Nodes().List(options)
+				if err != nil {
+					glog.Exit(err, "failed to list nodes in newKubeSubnetManager")
+				}
+				return obj, err
 			},
 			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-				return ksm.client.CoreV1().Nodes().Watch(options)
+				iface, err := ksm.client.CoreV1().Nodes().Watch(options)
+				if err != nil {
+					glog.Exit(err, "failed to watch nodes in newKubeSubnetManager")
+				}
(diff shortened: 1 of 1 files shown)


rajatchopra:
Valid enhancement, but we should probably use the informer factory from client-go tools.

quickstar:
This fix is crucial! I'm struggling with this issue since I first created a windows based kubelet!

rajatchopra:
Not sure if we should Exit. Can we use the 'context' and cancel it instead?
So that an appropriate cleanup can happen.

@Oats87 PTAL

Oats87:
@rajatchopra Looking at this, I think it is safe to merge this PR with the idea to perform a future enhancement to move to utilizing the nodeInformer factory, although interestingly in `client-go` itself there is not behavior similar to this: https://github.com/kubernetes/client-go/blob/master/informers/core/v1/node.go#L64

subnet/kube/kube.go
```diff
@@ -142,10 +142,18 @@ func newKubeSubnetManager(c clientset.Interface, sc *subnet.Config, nodeName, pr
 	indexer, controller := cache.NewIndexerInformer(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				return ksm.client.CoreV1().Nodes().List(options)
+				obj, err := ksm.client.CoreV1().Nodes().List(options)
+				if err != nil {
+					glog.Exit(err, "failed to list nodes in newKubeSubnetManager")
```
rajatchopra, line 147: Why quit? This should recover on its own.

rajatchopra:
For #1272 we should improve the reconciliation loop. We should recover from the error with retries (exponentially backing ones if at all), rather than exiting i.e. establish a new connection to apiserver.

@KnicKnic Your opinion please?

Closing this PR, please re-open if anyone thinks otherwise.

PR #1325, opened 2020-07-24
Add NET_RAW capability to support cri-o

kubealex:
cri-o drops NET_RAW capability by default, so flannel container is not able to self-set its iptables rules.

Adding NET_RAW capability allows flannel to become compatible with k8s using cri-o as runtime.

Otherwise, the pods remain looping in iptables rules creation.
Solves:
coreos#1300


mrunalp:
:+1:

Issue #719, opened 2017-05-14
“kube-subnet-mgr” doesn't work over 100 nodes


drinktee:
## Expected Behavior
I have setup a 170 nodes kubernetes cluster. I used [this daemonset yaml](https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml) to deploy flannel. Network was set 172.17.0.0/16. I found that only aboud 100 nodes works and other nodes didn't hava flannel0 , flannel.1.
## Current Behavior
The log prints 'Waiting %s for node controller to syn'. I found that the code seems to hang at [this place](https://github.com/coreos/flannel/blob/master/subnet/kube/kube.go#L112).
After I delete '--kube-subnet-mgr' and use etcd to config network , the cluster works.Every node has a flannel interface.

## Your Environment

* Flannel version: 0.7.1
* Backend used (e.g. vxlan or udp):vxlan
* Etcd version: 3.1.7
* Kubernetes version (if used): 1.6.2
* Operating System and version:Centos6

tomdee:
@drinktee Do you have the logs from the API server when this was happening?

tomdee:
It's possible (though I can't see how) that it could be related to the "100" here - https://github.com/coreos/flannel/blob/master/subnet/kube/kube.go#L129

Capitrium:
I'm running into the same issue - kube-flannel seems to hang on new nodes once the cluster has reached 100 running nodes. The kube-flannel logs shows "Waiting 10m0s for node controller to sync", but that timeout never seems to expire. I don't see any red flags in the logs myself, but I've included them below.

Logs from a broken kube-flannel pod, which has been running for ~30 minutes now:
```
$ kubectl logs -n kube-system kube-flannel-ds-0763k -f -c kube-flannel
I0518 20:21:27.924866       1 kube.go:111] Waiting 10m0s for node controller to sync
I0518 20:21:27.924939       1 kube.go:315] starting kube subnet manager
```

API server logs:
```
$ kubectl logs -n kube-system kube-apiserver-ip-10-0-15-238.ec2.internal -f
I0518 19:41:16.120565       1 aws.go:762] Building AWS cloudprovider
I0518 19:41:16.120654       1 aws.go:725] Zone not specified in configuration file; querying AWS metadata service
I0518 19:41:16.365252       1 tags.go:76] AWS cloud filtering on ClusterID: dev-cluster
[...]

PR #729, opened 2017-05-19
Allow more events to be buffered before the backend starts

tomdee:
Fixes #719

PR #1352, opened 2020-09-24
Allow more events to be buffered before the backend starts

ChenLingPeng:
Signed-off-by: forrestchen <forrestchen@tencent.com>

## Description
Our k8s cluster has more than 5000 nodes, chan with size 5000 will block flannel to run.

related issue: #719 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


ChenLingPeng:
#729 adjust the size to 5000, and it is not enough now.

rajatchopra:
Should we make this configurable? 
Through cli or env var, to the container in kube-flannel.yml daemonset.
Or, eventually when flannel has a helm chart based installation.

ChenLingPeng:
@rajatchopra PR updated, please take a review

luthermonson (approved):

subnet/kube/kube.go
```diff
@@ -138,7 +139,15 @@ func newKubeSubnetManager(c clientset.Interface, sc *subnet.Config, nodeName, pr
 	ksm.client = c
 	ksm.nodeName = nodeName
 	ksm.subnetConf = sc
-	ksm.events = make(chan subnet.Event, 5000)
+	scale := 5000
+	scaleStr := os.Getenv("K8S_SCALE")
+	if scaleStr != "" {
+		scale, _ := strconv.Atoi(scaleStr)
```
rajatchopra, line 145: We should be logging the error from Atoi?
Perhaps even return the error.

rajatchopra (changes requested):

subnet/kube/kube.go
ChenLingPeng: updated

subnet/kube/kube.go
```diff
@@ -138,7 +139,18 @@ func newKubeSubnetManager(c clientset.Interface, sc *subnet.Config, nodeName, pr
 	ksm.client = c
 	ksm.nodeName = nodeName
 	ksm.subnetConf = sc
-	ksm.events = make(chan subnet.Event, 5000)
+	scale := 5000
+	scaleStr := os.Getenv("K8S_SCALE")
```
rajatchopra, line 143: How about EVENT_QUEUE_DEPTH instead  of K8S_SCALE?
We also want this env var documented. Here: https://github.com/flannel-io/flannel/blob/master/Documentation/configuration.md#environment-variables

manuelbuil:
@ChenLingPeng will you have time to address the review?

ChenLingPeng:
> @ChenLingPeng will you have time to address the review?

updated, PTAL

luthermonson (approved):

rbrtbnfgl (approved):

PR #1368, opened 2020-11-11
subnets: move forward the cursor to skip illegal subnet

Sen666666:
This PR fixs an issue when flannel gets illegal subnet event in
watching leases, it doesn't move forward the etcd cursor and
will stuck in the same invalid event forever.

## Description
As the commit message mentions, in local_manager mode when the "WatchLeases" goroutine gets an invalid subnet event (in detail [function parseSubnetWatchResponse](https://github.com/coreos/flannel/blob/be2cb996dab08c68614a05e3bf30020c0a9fbafc/subnet/etcdv2/registry.go#L228) returns error), it doesn't move the cursor forward to skip this invalid subnet as the error string says.

This PR try to fix the issue.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
subnets: move forward the cursor to skip illegal subnet
```


luthermonson:
lgtm

PR #1372, opened 2020-11-16
Upgrade to k8s client-go 1.19.4

luthermonson:
finalizes removing juju per https://github.com/coreos/flannel/issues/1350
might fix other stuff like https://github.com/coreos/flannel/issues/1359


rajatchopra (approved):
/lgtm

PR #1378, opened 2020-11-23
k3s io go powershell

dweomer:
- Make embeddable
- Remove link local addresses
- Use charon from path if available
- Preserve PATH for extension backend
- Update logging for kube subnet manager
- Log with klog instead of glog
- Support k8s v1.18
- Use pkg/errors instead of juju/errors
- Update aws vpc backend for log formatting
- Remove glide files
- Add go.mod, udpate vendor
- [migration k3s-io] update go-powershell dependency


dweomer:
ugh, not sure how this ended up here. thanks `gh pr create --fill`

Issue #1391, opened 2020-12-27
Failed to restart Flannel HostGW mode in Windows(Server Core, version 1909) with K8S 1.18


thxCode:
## Expected Behavior

Following the official guide or using other platforms(like ACK) to setup HostGW mode in Windows (Server Core, version 1909) with K8S 1.18, and then restart the flannel process or the host. The flannel should continue to run.

## Current Behavior

Failed to restart flannel.
![image](https://user-images.githubusercontent.com/3394724/103163234-7eefa300-4836-11eb-8bb4-756608443bb1.png)

## Possible Solution

As a workaround, we need to remove the `${NETWORK_NAME}_ep` HNSEndpoint manually.
```
Get-HNSEndpoint | ? VirtualNetworkName -eq "cbr0" | ? Name -eq "cbr0_ep" | Remove-HNSEndpoint
```

## Steps to Reproduce (for bugs)

1. Setup K8S 1.18 cluster via any platforms that passed the Certified Kubernetes Conformance Program, like ACK
2. Setup the Windows Node.

## Context

It seems that Windows has added validation to check the re-add action of the same HNSEndpoint.

The same issue has been reported in the K8S community: https://github.com/kubernetes-sigs/sig-windows-tools/issues/94

## Your Environment

* Flannel version: v0.13.0
* Backend used (e.g. vxlan or udp): host-gw
* Etcd version: 
* Kubernetes version (if used): 1.18.8
* Operating System and version: Windows (Server Core, version 1909)
* Link to your project (optional):

luthermonson:
@thxCode thanks frank!

thxCode:
@luthermonson cheers

PR #1392, opened 2020-12-27
fix(host-gw): failed to restart if gateway hnsep existed

thxCode:
Skip the error if the gateway endpoint is existed, #1391

### DIFF
--- a/backend/hostgw/hostgw_windows.go
+++ b/backend/hostgw/hostgw_windows.go
@@ -16,19 +16,21 @@ package hostgw
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/Microsoft/hcsshim"
-	"github.com/coreos/flannel/backend"
-	"github.com/coreos/flannel/pkg/ip"
-	"github.com/coreos/flannel/pkg/routing"
-	"github.com/coreos/flannel/subnet"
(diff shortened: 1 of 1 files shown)


thxCode:
PTAL @rakelkar , @rajatchopra

luthermonson (approved):

backend/hostgw/hostgw_windows.go
```diff
@@ -222,7 +224,15 @@ func (be *HostgwBackend) RegisterNetwork(ctx context.Context, wg *sync.WaitGroup
 	log.Infof("Waiting to attach bridge endpoint %s to host", bridgeEndpointName)
 	waitErr = wait.Poll(500*time.Millisecond, 5*time.Second, func() (done bool, err error) {
 		lastErr = expectedBridgeEndpoint.HostAttach(1)
-		return lastErr == nil, nil
+		if lastErr == nil {
+			return true, nil
+		}
+		// See https://github.com/coreos/flannel/issues/1391 and
+		// hcsshim lacks some validations to detect the error, so we judge it by error message.
+		if strings.Contains(lastErr.Error(), "This endpoint is already attached to the switch.") {
```
rajatchopra, line 232: It would be nice if we could get the correct Error type from the hcsshim API. Matching strings from a library wouldn't fly long term.
Next MR?

PR #1401, opened 2021-01-20
Concern only about flannel ip addresses

### LINKED ISSUE #1060: "link has incompatible addresses" after restarting Flannel k8s pod
Flannel pod starts successfully if `flannel.1` link doesn't exist. But running `kubectl delete pod kube-flannel-...` leaves the `flannel.1` link and the subsequently created pod will fail to start with the following error:

```
I1103 17:54:41.197308       1 main.go:475] Determining IP address of default interface
I1103 17:54:41.198443       1 main.go:488] Using interface with name eth0 and address 192.168.1.244
I1103 17:54:41.198533       1 main.go:505] Defaulting external address to interface address (192.168.1.244)
I1103 17:54:41.698456       1 kube.go:131] Waiting 10m0s for node controller to sync
I1103 17:54:41.698812       1 kube.go:294] Starting kube subnet manager
I1103 17:54:42.699051       1 kube.go:138] Node controller sync successful
I1103 17:54:42.699241       1 main.go:235] Created subnet manager: Kubernetes Subnet Manager - k8s-worker2
I1103 17:54:42.699302       1 main.go:238] Installing signal handlers
I1103 17:54:42.796179       1 main.go:353] Found network config - Backend type: vxlan
I1103 17:54:42.796884       1 vxlan.go:120] VXLAN config: VNI=1 Port=0 GBP=false DirectRouting=false
E1103 17:54:42.799114       1 main.go:280] Error registering network: failed to configure interface flannel.1: failed to ensure address of interface flannel.1: link has incompatible addresses. Remove additional addresses and try again. &netlink.Vxlan{LinkAttrs:netlink.LinkAttrs{Index:715, MTU:1450, TxQLen:0, Name:"flannel.1", HardwareAddr:net.HardwareAddr{0x96, 0x11, 0x68, 0xa1, 0x57, 0x2b}, Flags:0x13, RawFlags:0x11043, ParentIndex:0, MasterIndex:0, Namespace:interface {}(nil), Alias:"", Statistics:(*netlink.LinkStatistics)(0x13c1a0e4), Promisc:0, Xdp:(*netlink.LinkXdp)(0x13d673a0), EncapType:"ether", Protinfo:(*netlink.Protinfo)(nil), OperState:0x0}, VxlanId:1, VtepDevIndex:2, SrcAddr:net.IP{0xc0, 0xa8, 0x1, 0xf4}, Group:net.IP(nil), TTL:0, TOS:0, Learning:false, Proxy:false, RSC:false, L2miss:false, L3miss:false, UDPCSum:true, NoAge:false, GBP:false, Age:300, Limit:0, Port:8472, PortLow:0, PortHigh:0}
I1103 17:54:42.799393       1 main.go:333] Stopping shutdownHandler...
```

This is on a raspberry pi 3 B+, so the arm architecture may be a factor.

## Expected Behavior

Recreating the flannel pod should succeed even if `flannel.1` link exists.

## Current Behavior

Flannel pod goes into CrashLoopBackoff after the pod is recreated. To allow the pod to start successfully, SSH onto the worker and run `sudo ip link delete flannel.1`. Recreating the pod will then start successfully.

## Possible Solution
?

1. Deploy k8s with flannel
2. `kubectl delete pod kube-flannel-...`
3. See that recreated pod does not start and logs above error message

## Context

About once a week my flannel pods enter this state, possibly due to a crash or restart of the pod, and I have to manually SSH in to delete the flannel link on each affected node.

Related issue: https://github.com/coreos/flannel/issues/883

## Your Environment

* Flannel version: v0.10.0
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version: 3.2.24
* Kubernetes version (if used): 1.12.1
* Operating System and version: HypriotOS 1.9.0 on Raspberry Pi 3 B+
* Link to your project (optional): n/a

jcaamano:
## Description
Currently flannel interface ip addresses are checked on startup when
using vxlan and ipip backends. If multiple addresses are found, startup
fails fatally. If only one address is found and is not the currently
leased one, it will be assumed that it comes from a previous lease and
be removed.

This criteria seems arbitrary both in how it is done and in its timing.
It may cause failures in situations where it might not be strictly
necessary like for example if the node is running a dhcp client that is
assigning link local addresses to all interfaces. It also might fail at
flannel unexpected restarts which are completly unrelated to
the external event that caused the unexpected modification in the
flannel interface.

This patch proposes to concern and check only ip address within the
flannel network and takes the simple approach to ignore any other ip
addresses assuming these would pose no problem on flannel operation.

A discarded but more agressive alternative would be to remove all
addresses that are not the currently leased one.

Fixes #1060

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@suse.com>

### DIFF
--- a/pkg/ip/iface.go
+++ b/pkg/ip/iface.go
@@ -23,6 +23,7 @@ import (
 	"syscall"
 
 	"github.com/vishvananda/netlink"
+	log "k8s.io/klog"
 )
 
 func getIfaceAddrs(iface *net.Interface) ([]netlink.Addr, error) {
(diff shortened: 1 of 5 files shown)


pkg/ip/iface.go
```diff
@@ -131,32 +131,32 @@ func DirectRouting(ip net.IP) (bool, error) {
 	return false, nil
 }
 
-// EnsureV4AddressOnLink ensures that there is only one v4 Addr on `link` and it equals `ipn`.
-// If there exist multiple addresses on link, it returns an error message to tell callers to remove additional address.
-func EnsureV4AddressOnLink(ipn IP4Net, link netlink.Link) error {
-	addr := netlink.Addr{IPNet: ipn.ToIPNet()}
+// EnsureV4AddressOnLink ensures that there is only one v4 Addr on `link` within the `net` address space and it equals `ipn`.
```
rajatchopra, line 134: s/and it equals `ipn`/and it equals `ipa`/ ?

pkg/ip/iface.go
```diff
@@ -131,32 +131,32 @@ func DirectRouting(ip net.IP) (bool, error) {
...
+			hasAddr = true
+			continue
+		}
 
-	// If the device has an incompatible address then delete it. This can happen if the lease changes for example.
-	if len(existingAddrs) == 1 && !existingAddrs[0].Equal(addr) {
-		if err := netlink.AddrDel(link, &existingAddrs[0]); err != nil {
-			return fmt.Errorf("failed to remove IP address %s from %s: %s", ipn.String(), link.Attrs().Name, err)
+		if ipn.Contains(FromIP(existingAddr.IP)) {
+			if err := netlink.AddrDel(link, &existingAddr); err != nil {
```
rajatchopra, line 150: Can we log this as Info nevertheless? Useful since we are doing kind of a 'destructive' operation.

rajatchopra (approved):
Suggested some cosmetic changes. Good to go.

jcaamano:
Addressed comments, thanks!

manuelbuil (approved):
LGTM

rajatchopra (approved):
/lgtm

PR #1404, opened 2021-01-26
Preserve environment for extension backend

jcaamano:
## Description
Commands executed for the extension backend preserve the existing
process environment adding flannel specific variables instead of
replacing the whole environment.

### DIFF
--- a/backend/extension/extension.go
+++ b/backend/extension/extension.go
@@ -17,6 +17,7 @@ package extension
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


luthermonson (approved):
@rajatchopra think we can get this in too?

jcaamano:
@luthermonson @rajatchopra Could you take a look also at https://github.com/coreos/flannel/pull/1401? We need it for the same reasons.

PR #1417, opened 2021-03-10
fix(windows): nil pointer panic

thxCode:
Get a nil pointer panic if HCN/HNS failed to query the HostComputeNetwork/HNSNetwork via network id, as the `newNetwork` reference has been overridden with `nil`.


thxCode:
PTAL @luthermonson  @rajatchopra

luthermonson (approved):

PR #1422, opened 2021-03-18
moving go modules to flannel-io/flannel and updating to go 1.16

luthermonson:
## Description
Since moving to the new org we should update all the git checkouts and modules to point to the new location.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


rajatchopra (approved):
/lgtm
Thanks for taking care of this.

Issue #1431, opened 2021-04-26
Add a flag for whether to clear NodeNetworkUnavailable on start


xh4n3:
## Expected Behavior
NodeNetworkUnavailable should not get cleared on flannel's start.

## Current Behavior
NodeNetworkUnavailable is cleared on flannel's start, users' pod are possibly scheduled to newly-created nodes before cloud controller sets their routes, then these pods cannot communicate with other pods.

## Possible Solution
can we add a new flag indicating whether to clear NodeNetworkUnavailable on start? I can send a pr if agreed.

## Context
the PR #1165 forced flannel to clear NodeNetworkUnavailable on start, but when deploying in cloud environments, cloud controller are supposed to add routes for nodes and and mark NodeNetworkUnavailable to false, like 
 this cloud-provider-alibaba-cloud https://github.com/kubernetes/cloud-provider-alibaba-cloud/blob/master/cloud-controller-manager/controller/route/controller.go#L436

## Your Environment
Alibaba Cloud + cloud-provider-alibaba-cloud + flannel 0.13.0
related issue https://github.com/kubernetes/kubernetes/issues/56794

xh4n3:
@rajatchopra could you please take a look at this issue?
I can send a pr about this if agreed.

PR #1435, opened 2021-05-13
add flag setNodeNetworkUnavailable

xh4n3:
## Description
This is to add flag for #1431


luthermonson:
with defaulting the new bool to true this is a very safe PR. not sure the consequenced of merging it but feel comfortable doing it. if anyone has other opinions that this shouldnt be gated by a bool flag then please say something now.

luthermonson (approved):

PR #1436, opened 2021-05-14
Fix ipsec

hvenev-vmware:
## Description
f4d044b6680b73a19f58bae520b17336d2f70cc5 configures strongswan to try to reestablish a tunnel after connectivity loss. The default is to drop it.

23a6e337919a47e2412c9a8c9c39d26afd7c3e93 enables rekeying which is necessary if we ever want to send more than 2^32 packets between two nodes.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil:
LGTM!

manuelbuil (approved):

PR #1443, opened 2021-05-21
Do not require subnet calculation if kubeSubnetMgr

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>

## Description
When using kubeSubnetMgr, the subnet for each node is fetched from the podCIDR field of the node. Therefore, it is not needed to calculate the subnet based on parameters like SubnetLen, SubnetMin or SubnetMax.

This will help with the dual-stack PR 

## Todos
- [X] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/subnet/config.go
+++ b/subnet/config.go
@@ -45,67 +45,71 @@ func parseBackendType(be json.RawMessage) (string, error) {
 	return bt.Type, nil
 }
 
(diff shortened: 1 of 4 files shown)


manuelbuil:
It does not make sense because the bridge CNI requires the Network parameter

Issue #248, opened 2015-07-10
*: add IPv6 support


GabMgt:
Hello, I am using Flannel in CoreOS 717.1.0 with IPv6 but it failed to launch. When I look at the logs, I see the message:

```
Failed to find IPv4 address for interface eth1.
```

Is Flannel compatible with IPv6 ?

Thank you.

eyakubovich:
@Nurza flannel currently does not support IPv6. However let's leave this issue open as we should add IPv6 support in the future.

pierrebeaucamp:
+1

eyakubovich:
@patrickhoefler Can I ask about your use case for IPv6? Are you running out of RFC1918 IPv4 addresses? Or want to migrate to IPv6 throughout your org? Something else?

I'm not against IPv6 support but it is work and considering that it's usually ran in 1918 address space, I don't see it as a high priority item. But I would like to know more.

patrickhoefler:
@eyakubovich You can definitely ask me, but it would probably make more sense to ask @Nurza instead ;)

Nurza:
Because I own a few servers with free IPv6 blocks (IPv4 is SO expensive) and I would like to use flannel with them.

fskale:
I think there are enough use cases, which require ipv6 addresses. I'm working on such and the lack of ipv6 support in kubernetes, flannel forces me to use a single docker host with pipework. Docker supports ipv6 using the command line option: --ipv6 --fixed-cidr-v6=\"xxxx:xxxx:xxxx:xxxx:/64\"". I'm Working with Unique Local Unicast Addresses. The usecase is clear. Ipv6only with DNATS ans SNATS to the pods. Personally i think, you should give ipv6 more priority. So i'm wirting on something like kiwi to track changes to a pod, and then issue pipework cmd to go back to kubernetes.

colinrgodsey:
:+1:

goacid:
+1

hwinkel:
+1

atta:
+1

glennswest:
Currently if you start doing sizing, you can get 300-400 hosts in a rack. (Supermicro microblades), two racks will go over the flannel/ipv4 limites. If you start looking at maximums, the flannel/ipv4 limit sounds ok, at around 261K containers max, but the reality its only 1-2 racks today. If you actually start applying devops/microservices into realworld apps, the container counts explode. I did a sizing for a day trading app im designing and its about 1.8million containers. 

[...]

PR #1446, opened 2021-05-28
Add dual stack support

manuelbuil:
Add new option for flannel daemon to support dual stack:
- "publicIPv6": "IPv6 accessible by other nodes for
  inter-host communication"

Add new option into `net-conf.json` configuration, like following:
{
  "EnableIPv4": true,
  "EnableIPv6": true,
  "Network": "172.16.0.0/16",
  "IPv6Network": "fc00::/48",
  "Backend": {
    "Type": "vxlan"
  }
}
EnableIPv4 default value is true for using kube subnet manager.
EnableIpv6 default value is false.

Flannel dual stack feature has limitation, only work with vxlan backend
and kube subnet manager now. To enable flannel dual stack feature, need
to do the following step:
1. setting flanneld daemon with "--kube-subnet-mgr"
2. settting "EnableIPv6" and "IPv6Network" in "net-conf.json" like the
above configuration.
3. setting network interface that flannel used ipv6 address and
default ipv6 gateway in the host node.
4. vxlan support ipv6 tunnel require kernel version >= 3.12
5. We require the Alpine image to move to version 3.13. Otherwise we get `Failed to ensure iptables rules: Error tearing down rules: running [/sbin/ip6tables -t nat -D POSTROUTING -s fc00::/48 ! -d ff00::/8 -j MASQUERADE --random-fully --wait]: exit status 2: ip6tables v1.8.4 (legacy): unknown option "--random-fully"` ==> https://github.com/flannel-io/flannel/pull/1447

Signed-off-by: yaoice <yao3690093@gmail.com>

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Dual-stack support for flannel with vxlan backend
```
It also need flannel cni plugin to support dual stack ip allocation, so it
depends on: containernetworking/plugins#570

#248

### DIFF
--- a/backend/common.go
+++ b/backend/common.go
@@ -24,9 +24,11 @@ import (
 )
 
 type ExternalInterface struct {
-	Iface     *net.Interface
-	IfaceAddr net.IP
-	ExtAddr   net.IP
+	Iface       *net.Interface
+	IfaceAddr   net.IP
+	IfaceV6Addr net.IP
+	ExtAddr     net.IP
+	ExtV6Addr   net.IP
 }
 
 // Besides the entry points in the Backend interface, the backend's New()
--- a/backend/vxlan/device.go
+++ b/backend/vxlan/device.go
@@ -124,6 +124,18 @@ func (dev *vxlanDevice) Configure(ipa ip.IP4Net, flannelnet ip.IP4Net) error {
 	return nil
 }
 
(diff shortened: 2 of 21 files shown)


manuelbuil:
Wrong branch!

PR #1465, opened 2021-08-02
Adding copy of flannel-cni binary to the host cni/bin folder

AleksandrNull:
## Description
Because flannel cni binary wouldn't be a part of upstream we need to copy it to the host on the first start

### DIFF
--- a/Documentation/kube-flannel.yml
+++ b/Documentation/kube-flannel.yml
@@ -168,16 +168,16 @@ spec:
       - name: install-cni
         image: quay.io/coreos/flannel:v0.14.0
         command:
-        - cp
+        - /bin/sh
         args:
-        - -f
-        - /etc/kube-flannel/cni-conf.json
-        - /etc/cni/net.d/10-flannel.conflist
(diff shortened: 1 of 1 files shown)


AleksandrNull:
@rajatchopra Please review.

manuelbuil:
According to the flannel config, we also need the binary portmap:
```    
{
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    }
```

yuanying:
Since flannel also depends on the bridge and ipam/host-local plugins, I guess the original manifest expects that the CNI plugins it depends on are already installed except flannel CNI plugin itself.

commarmi76:
It doesn't work for me, I have to change:

`        - [ '-c', 'cp -f /opt/bin/flannel /host/opt/cni/bin/flannel;cp -f /etc/kube-flannel/cni-conf.json /etc/cni/net.d/10-flannel.conflist' ]`

to 
```
- -c
- 'cp -f /opt/bin/flannel /host/opt/cni/bin/flannel;cp -f /etc/kube-flannel/cni-conf.json /etc/cni/net.d/10-flannel.conflist'
```

But the binary /opt/bin/flannel doesn't exist in "quay.io/coreos/flannel:v0.14.0" image.

PR #1501, opened 2021-10-30
Add dual-stack support to the extension backend

sjoerdsimons:
## Description

This adds dual-stack support to the extension backend; Mostly for use with wireguard. 

Tested to succeed with the existing wireguard functional tests (after uncommenting them) as well as with integration into k3s: https://github.com/sjoerdsimons/k3s/tree/flannel-wireguard-dualstack

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -83,7 +83,7 @@ Set `healthz-port` to a non-zero value will enable a healthz server for flannel.
 
 ## Dual-stack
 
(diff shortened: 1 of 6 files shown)


manuelbuil:
I guess we can close this one after merging the wireguard PR, right?

sjoerdsimons:
It depends on whether you see dual-stack support in extensions to be valuable or not; I personally cared about wireguard, which now ofcourse has the official native backend, but the same code works for others.

Let me know what you prefer, i'm happy to update this branch as it conflicts now or close it

manuelbuil:
I'd say close it because we should use the new way :)

PR #1513, opened 2021-12-09
ip: Add helpers to determine if an IP address is private

sjoerdsimons:
Add IsPrivate methods to IP4 and IP6 to determine if a ip address is in
the private range. similar to IsPrivate of net.IP in newer golang
versions.

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

## Description

Add some helpers for determining of address are in a private range or not; useful for determining if masquarading should be enabled.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
```release-note
None required
```

### DIFF
--- a/pkg/ip/ip6net.go
+++ b/pkg/ip/ip6net.go
@@ -110,6 +110,12 @@ func (ip6 *IP6) UnmarshalJSON(j []byte) error {
 	}
 }
 
+// IsPrivate reports whether ip is a private address, according to
+// RFC 4193 (IPv6 addresses).
+func (ip6 *IP6) IsPrivate() bool {
+	return (*big.Int)(ip6).Bytes()[0]&0xfe == 0xfc
+}
+
 // similar to net.IPNet but has uint based representation
 type IP6Net struct {
 	IP        *IP6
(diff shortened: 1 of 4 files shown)


manuelbuil:
Are you planning on implementing the code that will apply masq if it is private?

sjoerdsimons:
fwiw here is an example usage: https://github.com/sjoerdsimons/k3s/commit/f53921d1c3cb636f57673a3a575f1a4d9c1a32f4 

that's my old k3s branch for the previous rounds of wireguard dualstacking; I was submitting as i was prepping an MR to k3s for what is now merged in flannel which will also use this commit ;).

luthermonson:
@sjoerdsimons can we just upgrade go however needed and use the net.IP functionaliry?

sjoerdsimons:
> @sjoerdsimons can we just upgrade go however needed and use the net.IP functionaliry?

I don't know what flannels policy is for updating the Go version. But iirc (it's been a while) if that's updated to a new enough version most if not all of this patch would unneeded indeed

sjoerdsimons:
looks like i accidentally closed this one; woops

sjoerdsimons:
Sorry for the noise; now i remember why i deleted the relevant branch; the branch name ending in .go confuses the flannel testsuite and on cannot change the source branch of an MR in github :facepalm:

PR #1523, opened 2022-01-16
Split ipv4 and ipv6 masquarading options

sjoerdsimons:
## Description

Masquarading policies for ipv4 and ipv6 are ortogonal e.g. a cluster
could have a private range of ipv4 addresses but a public ipv6 range. In
which case you'd probably want to masquarade ipv4 but not ipv6.

## Release Note

```release-note
* the --ipv6-masq option has been added to enable ipv6 masquarading; --ip-masq now only enables ipv4 masquarading
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -90,6 +90,7 @@ type CmdLineOpts struct {
 	iface                     flagSlice
 	ifaceRegex                flagSlice
 	ipMasq                    bool
+	ipv6Masq                  bool
 	subnetFile                string
 	subnetDir                 string
 	publicIP                  string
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -767,10 +774,10 @@ func WriteSubnetFile(path string, config *subnet.Config, ipMasq bool, bn backend
 		ip6Sn.IncrementIP()
 		fmt.Fprintf(f, "FLANNEL_IPV6_NETWORK=%s\n", ip6Nw)
 		fmt.Fprintf(f, "FLANNEL_IPV6_SUBNET=%s\n", ip6Sn)
+		fmt.Fprintf(f, "FLANNEL_IPV6_MASQ=%v\n", ip6Masq)
```
manuelbuil, line 777: This variable should be consumed by the flannel binary but there is no `FLANNEL_IPV6_MASQ` there

manuelbuil (commented):
To achieve this, we will need to provide the difference masquerading options in the Bridge CNI, which is the CNI that the flannel binary uses to implement the intra-node communication. Currently, the Bridge CNI implements masquerade on both ipv4 and ipv6 if FLANNEL_MASQ=true: https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L612-L620

main.go
sjoerdsimons: you mean the flannel binary from https://github.com/flannel-io/cni-plugin right?  What behaviour would you like to see there; This stuff is all new to me so i might get this wrong. But from reading the code of the flannel cni plugin it tells the delegate CNI plugin (e.g. bridge) to apply ip masquarding *if* flannel doesn't do ip masquarding. 

So basically when using flannel via CNI you cannot even configure your cluster without ip masquarading... All the CNI plugin does is avoiding double ipmasq config.

So either the documentation should simply state that when using as CNI both masq options should be kept in sync (e.g. both false (the default) or both set... Otherwise we need to tumble down the rabbit hole and update the CNI plugin interface from containers and teach it about seperate ipv4 vs. ipv6 config :/

fwiw the only reason i worked on this patch was the suggestion it was needed to allow k3s to configure v4/v6 ip masq which in the mean seems to have been fixed regardless so...

manuelbuil: Yes, that's the project. The function `WriteSubnetFile` in flannel generates a file called subnet.env which is the config file for the flannel binary. Then, this function in the flannel binary: https://github.com/flannel-io/cni-plugin/blob/main/flannel.go#L114, reads the subnet.env file where `FLANNEL_IPV6_MASQ` is. As you can observe below the `switch` (line 126), there is no `FLANNEL_IPV6_MASQ` there.

Yes, you are right. When passing `FLANNEL_MASQ=true`, the flannel binary, or to be specific the bridge binary, will not do masquerading. It is important to note that if `FLANNEL_MASQ=false`, the bridge binary does the masquerade for both ipv4 and ipv6. So, if we want to differentiate between them, we need `FLANNEL_MASQ=true` and do the masquerade in flanneld.

In summary, the main part of this PR is good and helpful, thanks! Just remove the `FLANNEL_IPV6_MASQ` thing and make sure that when `ipv6-masq=true` then `FLANNEL_MASQ=true`

PR #1526, opened 2022-01-23
rebase add mips64le support

jiangxiaobin96:
The ARCH is mips64le because mipsel is a joke created by Debian. (Because little endian reverse the byte order, they also reversed the letter LE to form el.)

The change version can run successfully by cross-compile. But still need to test because I don't have mips64el environment.

### DIFF
--- a/Dockerfile.mips64le
+++ b/Dockerfile.mips64le
@@ -1,0 +1,13 @@
+FROM loongnix/alpine:3.11
+
+LABEL maintainer="Tom Denham <tom@tigera.io>"
+
+ENV FLANNEL_ARCH=mips64le
+
+ADD dist/qemu-$FLANNEL_ARCH-static /usr/bin/qemu-$FLANNEL_ARCH-static
+RUN apk add --no-cache iproute2 net-tools ca-certificates iptables strongswan && update-ca-certificates
+RUN apk add wireguard-tools --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing
+COPY dist/flanneld-$FLANNEL_ARCH /opt/bin/flanneld
+COPY dist/mk-docker-opts.sh /opt/bin/
+
+ENTRYPOINT ["/opt/bin/flanneld"]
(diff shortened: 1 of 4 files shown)


manuelbuil:
Please rebase, there is a conflict in Makefile

Dockerfile.mips64le
```diff
@@ -0,0 +1,13 @@
+FROM loongnix/alpine:3.11
+
+LABEL maintainer="Tom Denham <tom@tigera.io>"
```
manuelbuil, line 3: I'm not sure if Tom wants to be the maintainer of this image

manuelbuil (changes requested):
Could you please update https://github.com/flannel-io/flannel/blob/master/.github/workflows/build.yaml and https://github.com/flannel-io/flannel/blob/master/.github/workflows/release.yml too?

Dockerfile.mips64le
jiangxiaobin96: Oh so sorry i have refer to the old version and forgot changing this line. Now i have follow the latest version.

jiangxiaobin96: the image is from [#76](https://github.com/alpinelinux/docker-alpine/issues/76) for mips64le

manuelbuil (approved):

Dockerfile.mips64le
```diff
@@ -0,0 +1,13 @@
+FROM loongnix/alpine:3.11
```
rosskirkpat, line 1: Please use `mips64le/debian:stable-20211220-slim` to match what we use in flannel's cni-plugin for [mips64le](https://github.com/flannel-io/cni-plugin/blob/main/Dockerfile.mips64le).

rosskirkpat (changes requested):
The base image for the mips64le dockerfile needs to be swapped. Other than that, LGTM

Dockerfile.mips64le
jiangxiaobin96: OK, i have changed the image and modify the relevant package. If there is no problems, i will delete the annotations.

rosskirkpat (approved):
LGTM. Thanks for addressing my feedback!

PR #1545, opened 2022-03-08
Add --iface-can-reach option as new way to decide IF

takmatsu:
### Background
Actually this is similar to Calico's `can-reach=DESTINATION` feature. https://projectcalico.docs.tigera.io/reference/node/configuration#can-reachdestination

maybe background is same, this option is really convenient for the cluster having many kind of nodes.

### Tests
Actually we're using in production cluster.

### Effect
Completely new feature, current feature shouldn't be affected.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
--iface-can-reach is added as new option to decide which interface will be used 
```


manuelbuil (commented):
Thanks for the contribution. The value of `--iface-can-reach=` should be an IP, right? Could you add an example in the documentation about how to use it? Thanks!

manuelbuil (commented):
Thanks. It looks very good. Could you add a simple test case in `pkg/ipmatch/match_test.go`? There is one unit test for each option to find the interface

luthermonson:
@takmatsu seconding some tests, can you wrap GetInterfaceBySpecificIPRouting with a unit test and we can get this merged

takmatsu:
sorry for late reply, I added some tests and also found unexpected behavior. and it should be fixed.

manuelbuil (approved):

rbrtbnfgl (approved):

unixfox:
Does this work with IPv6 too or only with IPV4?

PR #1555, opened 2022-03-28
fix document and comment of configuration

ari1021:
## Description

- fix description of `SubnetLen`, `SubnetMin`, `IPv6SubnetLen`, `IPv6SubnetMin`. 
  - [SubnetLen](https://github.com/flannel-io/flannel/blob/master/subnet/config.go#L77)
  - [SubnetMin](https://github.com/flannel-io/flannel/blob/master/subnet/config.go#L96)
  - [IPv6SubnetLen](https://github.com/flannel-io/flannel/blob/master/subnet/config.go#L134)
  - [IPv6SubnetMin](https://github.com/flannel-io/flannel/blob/master/subnet/config.go#L153)
- fix typo about why we skip over the first subnet

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -16,19 +16,19 @@ The value of the config is a JSON dictionary with the following keys:
 * `EnableIPv6` (bool): Enables ipv6 support
 
 * `SubnetLen` (integer): The size of the subnet allocated to each host.
-   Defaults to 24 (i.e. /24) unless `Network` was configured to be smaller than a /24 in which case it is one less than the network.
+   Defaults to 24 (i.e. /24) unless `Network` was configured to be smaller than a /22 in which case it is two less than the network.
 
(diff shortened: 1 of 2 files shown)


ari1021:
It appears that `SubnetLen`, `SubnetMin`, `IPv6SubnetLen`, `IPv6SubnetMin` are not used when using kube subnet manager. If this is true, it should be noted in the document, or is it already written somewhere?

manuelbuil:
> It appears that `SubnetLen`, `SubnetMin`, `IPv6SubnetLen`, `IPv6SubnetMin` are not used when using kube subnet manager. If this is true, it should be noted in the document, or is it already written somewhere?

You are right. When using kube subnet those parameters are controlled by kubernetes flags. In that case, kubernetes is the one distributing ips across nodes

manuelbuil (approved):

luthermonson (approved):

PR #1556, opened 2022-04-08
Add validation that PodCIDR(s) matches the network specified in the `net-conf.json`

AlexPykavy:
Faced issue that `iptables` forwarding rules have been created for the default subnet from the `net-conf.json` config while actual PodCIDR was different.

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -350,9 +350,17 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, attrs *subnet.Le
 		Expiration: time.Now().Add(24 * time.Hour),
 	}
 	if cidr != nil && ksm.enableIPv4 {
+		if !containsCIDR(ksm.subnetConf.Network.ToIPNet(), cidr) {
+			return nil, fmt.Errorf("subnet %q specified in the flannel net config doesn't contain %q PodCIDR of the %q node.", ksm.subnetConf.Network, cidr, ksm.nodeName)
+		}
+
 		lease.Subnet = ip.FromIPNet(cidr)
 	}
 	if ipv6Cidr != nil {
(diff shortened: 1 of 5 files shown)


AlexPykavy:
Hi @manuelbuil , unfortunately `e2e` tests failed due to timeout. I couldn't find any errors in the log, so assume that the failure was caused by the external load on the server when the build was running. Could you please restart that pipeline?

manuelbuil:
> Hi @manuelbuil , unfortunately `e2e` tests failed due to timeout. I couldn't find any errors in the log, so assume that the failure was caused by the external load on the server when the build was running. Could you please restart that pipeline?

Yes, let's try again :)

AlexPykavy:
As it turned out, the problem was on my side and the tests helped to detect it. However troubleshooting them was quite difficult, so I decided to improve them a bit.

subnet/kube/kube.go
```diff
@@ -453,3 +461,9 @@ func (ksm *kubeSubnetManager) setNodeNetworkUnavailableFalse(ctx context.Context
 	_, err = ksm.client.CoreV1().Nodes().PatchStatus(ctx, ksm.nodeName, patch)
 	return err
 }
+
+func ContainsCIDR(ipnet1, ipnet2 *net.IPNet) bool {
```
luthermonson, line 465: can you write a unit test around this function please

luthermonson:
@AlexPykavy write a unit test around your ContainsCIDR func and this is able to be merged.

AlexPykavy:
@luthermonson , completely agree with you) Added as suggested, so could you please trigger a build to make sure that it works as we expect?

AlexPykavy:
@manuelbuil, @luthermonson, tests pass. Can we merge this?

manuelbuil:
> @manuelbuil, @luthermonson, tests pass. Can we merge this?

I'm fine. Waiting on @luthermonson

luthermonson (approved):

PR #1562, opened 2022-05-05
Upgrade etcd client to v3

thomasferrandiz:
## Description

This PR upgrades the etcd client to v3 to solve a CVE issue in the v2 client.

The registry tests now use `go.etcd.io/etcd/tests/v3/integration` instead of  `mock_etcd`.
This is why there are many new dependencies.

## Todos
- [x] Tests: connection to a cluster using TLS
- [ ] Documentation: should I update `Documentation/upgrade.md`?
- [X] Release note

## Release Note

```release-note
The v2 and v3 API of etcd are not compatible so if a live upgrade is performed, 
the flannel data stored in etcd must be migrated to be accessible by the v2 client after the upgrade of flannel.  

See the etcd documentation (https://etcd.io/docs/v3.5/op-guide/v2-migration/).
```


thomasferrandiz:
> @thomasferrandiz can you merge all the code commits into one commit e.g. squash the lint changes into the rest of the code. I also would like you to split out the go.mod, go.sum and ./vendor changes into a separate commit with commit message of `go mod` or something so the code changes are completely split off from the go modules changes.

Done in https://github.com/flannel-io/flannel/pull/1562/commits/6b6ba94d15f02ad023f9420e930395ebe0095da1 and https://github.com/flannel-io/flannel/pull/1562/commits/3a4cdf457c3702757a073057ec65b7f571f7319e.

luthermonson:
@thomasferrandiz merged your netlink upgrade, can you rebase this and we'll get it merged

luthermonson:
@manuelbuil give this a quick review but im feeling like we're done here now that ci works

manuelbuil:
If we replace etcdv2 with etcdv3, we should also change the name of the subnet/etcdv2 directory, right?

manuelbuil:
> @manuelbuil thought about this a bit too after we talked... maybe just rename to etcd? do we really need the version in the dir name?

good point. I don't think we do. So we could change it to etcd

thomasferrandiz:
I thought about changing the name but it would have made the diff hard to read so I decided not to rename the folder.
I'll do it now

luthermonson:
@thomasferrandiz squash that last commit into the first and let's get this merged

PR #1563, opened 2022-05-05
Add note about putting a flanneld binary in /opt/bin

radiosilence:
## Description

It took me ages to figure this out

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,9 @@ Though not required, it's recommended that flannel uses the Kubernetes API as it
 Flannel can be added to any existing Kubernetes cluster though it's simplest to add `flannel` before any pods using the pod network have been started.
 
 For Kubernetes v1.17+
-`kubectl apply -f https://raw.githubusercontent.com/flannel-io/flannel/master/Documentation/kube-flannel.yml`
+
+1. Make sure a `flanneld` binary exists at `/opt/bin/flanneld` on each node
+2. `kubectl apply -f https://raw.githubusercontent.com/flannel-io/flannel/master/Documentation/kube-flannel.yml`
 
 See [Kubernetes](Documentation/kubernetes.md) for more details.
 


luthermonson (approved):

radiosilence:
Not sure how a doc change could cause tests to fail :)

luthermonson:
ya it's not you, the e2e tests are broke

Thyroxine:
Hello!

Why do we need the manual installation of the `flanneld` binary onto the nodes?
We already have `flanneld` inside the Docker container.

It is very confusing addition.

We recently deployed a number of K8s clusters by `kubeadm` and everything is working perfectly by just applying the manifest, as stated in the previous version of the docs.

Flannel Pods are starting automatically after node reboot, as expected.

cprivitere:
Just gonna +1 this question, what's up with this requirement, @luthermonson ?

primeos-work:
I'm also confused by this change. Shouldn't it get reverted?

I guess, if necessary, it could be documented somewhere else in the README for alternative setups but it really doesn't seem to belong in front of `kubectl apply -f https://raw.githubusercontent.com/flannel-io/flannel/master/Documentation/kube-flannel.yml`.

`kube-flannel.yml` declares a `kube-flannel` container that currently uses the `docker.io/rancher/mirrored-flannelcni-flannel:v0.19.0` image which already contains `/opt/bin/flanneld`:
```console
[michael@groot ~]$ /opt/bin/flanneld --version
-bash: /opt/bin/flanneld: No such file or directory
[michael@groot ~]$ podman run -it --entrypoint /bin/sh docker.io/rancher/mirrored-flannelcni-flannel:v0.19.0
/ # /opt/bin/flanneld --version
v0.19.0
```
So the advice of installing `/opt/bin/flanneld` on the K8s node must simply be wrong in this context.

Issue #1581, opened 2022-06-08
Flannel IPv6: Incorrect sn6.IP.Cmp(config.IPv6SubnetMax) >= 0 comparison in allocateSubnet 


pranavadn:
Incorrect sn6.IP.Cmp(config.IPv6SubnetMax) >= 0 comparison in allocateSubnet  causes failure in leasing out last subnet of a network.

## Expected Behavior
The comparison should be sn6.IP.Cmp(config.IPv6SubnetMax) > 0.

## Current Behavior
Fails to lease out the last subnet of a network. According to the definition it should be able to lease out the last subnet -

_IPv6SubnetMax (string): The end of the IPv6 range at which the subnet allocation should end with. **Defaults to the last subnet of Ipv6Network**._

## Possible Solution
Change comparison to sn6.IP.Cmp(config.IPv6SubnetMax) > 0 or to <=0  like you have in ipv4 -
sn.IP <= config.SubnetMax

## Steps to Reproduce (for bugs)

1. Have a 3 node setup.
2. have /65 ipv6 network and subnet length of /67
3. Third node fails to get a lease even though 3 subnets are available. 
4. It errors out with "Error registering network: failed to acquire lease: out of subnets"

## Context
We need flannel to work with v6 for dual stack.

## Your Environment
Flannel version:
v0.17.0

Backend used (e.g. vxlan or udp):
host-gw

Etcd version:
3.3.15

Kubernetes version (if used):
1.20

Operating System and version:
3.10.0-1160.62.1.el7.x86_64
* Link to your project (optional):

manuelbuil:
Thanks for reporting this. Note that Kubernetes 1.20 does not support ipv6. You should move to v1.23 where the feature became stable. Can you try there?

pranavadn:
@manuelbuil These issues are coding errors in Flannel. They are present in latest Flannel binary too. We are not using Kubernetes subnet manager either. Hence, it is the subnet leasing logic of Flannel that is being used on our setup. Could we please have these Flannel coding errors fixed?

pranavadn:
Yes. Please close the issue.

Issue #1580, opened 2022-06-08
Flannel IPv6: allocateSubnet does not move sn6 with Next() in case of Overlap


pranavadn:
When using Flannel with IPv6, allocateSubnet does not move sn6 with Next() in case of overlap. This results in "out of subnets" error leading to nodes not being able to obtain a subnet.

## Expected Behavior
When a v6 subnet is found to overlap with an existing lease, we should move to next leave with sn6.Next().

## Current Behavior
When using Flannel with IPv6, allocateSubnet does not move sn6 with Next() in case of overlap. This results in "out of subnets" error leading to nodes not being able to obtain a subnet.

```
OuterLoop:
	for ; sn.IP <= config.SubnetMax && len(availableIPs) < 100; sn = sn.Next() {
		if !sn6.Empty() {
			if sn6.IP.Cmp(config.IPv6SubnetMax) >= 0 {
				break OuterLoop
			}
		}
		for _, l := range leases {
			if sn.Overlaps(l.Subnet) {
				continue OuterLoop
			}
			if !sn6.Empty() && sn6.Overlaps(l.IPv6Subnet) {
				continue OuterLoop
			}
		}
```

## Possible Solution
Remove the sn.Next() in the loop header and call sn.Next() and sn6.next() inside the loop appropriately for overlap and available IP accumulation cases.

## Steps to Reproduce (for bugs)

3 nodes cluster with the following flannel network config led to one node erroring out and not getting a lease -

Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: getNetworkConfig  {
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "EnableIPv6": true,
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "EnableIPv4": true,
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "Network": "192.168.0.0/16",
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "SubnetLen": 24,
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "IPv6Network": "2000:192:1600::0/64",
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "IPv6SubnetLen": 108,
Jun  7 00:11:49 a1-bqkp73611004-node-1 flanneld: "Backend": {
[...]

PR #1582, opened 2022-06-10
Fixed subnet allocation in case of etcd manager with IPv6

rbrtbnfgl:
## Description

Fix the subnet allocation in case of etcd when using IPv6 #1580 #1581 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/subnet/etcd/local_manager.go
+++ b/subnet/etcd/local_manager.go
@@ -202,11 +202,8 @@ func (m *LocalManager) allocateSubnet(config *Config, leases []Lease) (ip.IP4Net
 		log.Infof("Picking ipv6 subnet in range %s ... %s", config.IPv6SubnetMin, config.IPv6SubnetMax)
 	}
 
-	type IPs struct {
-		ip  ip.IP4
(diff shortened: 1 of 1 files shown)


pranavadn (commented):
Does sn6.IP.Cmp(config.IPv6SubnetMax) < 0 comparison not exclude the last subnet? This was also an issue in the original code that raised with 1581. It should be like the v4 comparison sn.IP <= config.SubnetMax.

rbrtbnfgl:
yes you are right, I thought that it was the first not acceptable IP but it is the latest. I'll fix it.

pranavadn:
Looks good to me. Please go ahead and merge.

pranavadn:
@rbrtbnfgl @manuelbuil Could we please get 0.18.2 image with these fixes?

pranavadn (approved):
Looks fine to me.

pranavadn (approved):

PR #1600, opened 2022-06-27
Public ip dualstack

### LINKED ISSUE #1596: Flannel Dual Stack Does not work when both public-ip and public-ipv6 are specified
## Current Behavior
Flannel Dual Stack Does not work when both public-ip and public-ipv6 are specified.

When both ipv4 and ipv6 addresses are specified via command line arguments public-ip and public-ipv6, it takes the following code path and errors out -

```
main 
----- calls ----
if len(opts.iface) == 0 && len(opts.ifaceRegex) == 0 && len(opts.ifaceCanReach) == 0 {
		extIface, err = ipmatch.LookupExtIface(opts.publicIP, "", "", ipStack, optsPublicIP)
---- calls -----
		case dualStack:
				iface, err = ip.GetInterfaceByIP(ifaceAddr)
				if err != nil {
					return nil, fmt.Errorf("error looking up interface %s: %s", ifname, err)
				}
				v6Iface, err := ip.GetInterfaceByIP6(ifaceAddr)
				if err != nil {
					return nil, fmt.Errorf("error looking up v6 interface %s: %s", ifname, err)
				}

```
Since only opts.publicIP is passed into LookupExtIface, ip.GetInterfaceByIP6(ifaceAddr) gets called with v4 address and errors out with 

`return nil, errors.New("No interface with given IPv6 found")
`
## Possible Solution
Need to pass provided ipv6 address to LookupExtIface call and use it to find the interface.

invoke flannel with both public-ip and public-ipv6 options

## Your Environment
Flannel version:
v0.17.0

Backend used (e.g. vxlan or udp):
host-gw

Etcd version:
3.3.15

Kubernetes version (if used):
1.20

Operating System and version:
3.10.0-1160.62.1.el7.x86_64

rbrtbnfgl:
## Description

Fix #1596 when Public IPv6 is configured the right interface will be selected and no errors occur in case of dualstack.

## Todos
- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -265,7 +265,11 @@ func main() {
 	}
 	// Check the default interface only if no interfaces are specified
 	if len(opts.iface) == 0 && len(opts.ifaceRegex) == 0 && len(opts.ifaceCanReach) == 0 {
-		extIface, err = ipmatch.LookupExtIface(opts.publicIP, "", "", ipStack, optsPublicIP)
+		if len(opts.publicIP) > 0 {
+			extIface, err = ipmatch.LookupExtIface(opts.publicIP, "", "", ipStack, optsPublicIP)
+		} else {
+			extIface, err = ipmatch.LookupExtIface(opts.publicIPv6, "", "", ipStack, optsPublicIP)
+		}
 		if err != nil {
 			log.Error("Failed to find any valid interface to use: ", err)
 			os.Exit(1)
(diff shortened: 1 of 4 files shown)


pranavadn (approved):
We patched and tested. Flannel came up fine and picked up the IPs we specified in command line args. Looks good to me.

thomasferrandiz (approved):

PR #1618, opened 2022-07-20
Updated documentation

rbrtbnfgl:
Updated documentation with right flannel version and docker registry.
Removed CoreOS integration paragraph because CoreOS is not shipped with flannel any more.

```release-note
None required
```


pgonin (approved):
fine by me

PR #1622, opened 2022-07-27
`public-ip-overwrite` using node-labels

skirsten:
Currently (as far as I know) the only way to overwrite the public-ip in kubernetes mode is via node annotations.
For k3s this means there is no way (I know of) to set the annotation before it will be consumed by the integrated flannel.
In this case k3s would have to be restarted to apply the change which is a deal-breaker for me.

This PR simply allows to configure the exact same overwrite via node-labels.
If both are present, the annotation is preferred.

I have not tested this locally but I added a functional test.


manuelbuil:
Have you tried the `node-external-ip` flag in k3s? Doesn't it fulfill your use case?

skirsten:
@manuelbuil [As far as I can tell](https://github.com/k3s-io/k3s/discussions/5911#discussioncomment-3268984), currently there is no logic to pass any k3s IP setting `node-ip` or `node-external-ip` through to flannel.
I think this was only implemented on the old non-native wireguard implementation on k3s side.

manuelbuil:
Good point. From flannel perspective, what would change if it knew there is a public-ip? In other words, could you describe a bit the use case you have in mind? I fear that what you are trying to achieve might not be available in flannel

manuelbuil:
Could you please rebase (to get the CI fix)? And could you please explain why having the same parameter in labels instead of annotations fixes your problem? Thanks

skirsten:
Hi, sorry for the delay.
Basically this PR was to set these options on k3s using the `--node-label` flag. So no manual interaction (`kubectl annotate`) was required.
Unfortunately I ran into the same problem for other projects (e.g. [kilo](https://github.com/squat/kilo)) so I had to build a custom annotation solution anyway.

So for me this PR is not needed anymore. But it might still help other people that want to configure these options without any manual interaction or a daemon.

PR #1625, opened 2022-08-03
Delete unmaintained cloud backends

thomasferrandiz:
## Description

This PR deletes 3 cloud backends that don't seem to be used and are not maintained: alivpc, awsvpc and gce.
 
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil (approved):

manuelbuil:
All these unmaintained backends are adding a lot of CVEs because of using old code. We suspect nobody is using them, therefore we will delete them. We are open to bring them back if somebody has the time to maintain its code

Issue #1646, opened 2022-09-15
when extend node, flannel cni report error: open subnet.env error


JornShen:
## Current Behavior

![image](https://user-images.githubusercontent.com/18069202/190374102-02ce0196-4349-4536-a047-42fa5d5ea9f4.png)

## Possible Solution
back off flanneld set network unavaliable false after wrote the subnet.env file

for this MR is https://github.com/flannel-io/flannel/pull/1647

## Steps to Reproduce (for bugs)
1. deploy a daemonset pod
2. extend a node
3. watch the pod error

## Context

may confuse our user of cluster 

## Reason

in current flanneld  code
https://github.com/flannel-io/flannel/blob/d5a5b16891204e2ae206fca02eaf0f48af1e6729/subnet/kube/kube.go#L349-L357

where main call here 
https://github.com/flannel-io/flannel/blob/d5a5b16891204e2ae206fca02eaf0f48af1e6729/main.go#L328-L334

which is before the subnet.env wrote 

https://github.com/flannel-io/flannel/blob/d5a5b16891204e2ae206fca02eaf0f48af1e6729/main.go#L375-L380

And if the process run after RegisterNetwork the set the NetworkUnavailable condition for node. kubelet start to create 
Pod, and run to flannel cni bin.

https://github.com/flannel-io/cni-plugin/blob/6e8bb11373c7743a00571a52d4f27ce7c07256a1/flannel.go#L235-L238

and read file subnet.env and cause error

https://github.com/flannel-io/cni-plugin/blob/6e8bb11373c7743a00571a52d4f27ce7c07256a1/flannel.go#L115-L118

## Your Environment

* Flannel version:  v0.14.0
* Backend used (e.g. vxlan or udp): alloc
* Etcd version:
* Kubernetes version (if used): v1.20.15
* Operating System and version: debian 10
* Link to your project (optional):

PR #1647, opened 2022-09-15
back off set NodeNetworkUnavailableFalse

JornShen:
## Description
fix https://github.com/flannel-io/flannel/issues/1646

## Release Note

```release-note
back off set NodeNetworkUnavailableFalse after subnet.env file in kube subnet mode.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -189,7 +189,7 @@ func newSubnetManager(ctx context.Context) (subnet.Manager, error) {
 	prevSubnet := ReadCIDRFromSubnetFile(opts.subnetFile, "FLANNEL_SUBNET")
 	prevIPv6Subnet := ReadIP6CIDRFromSubnetFile(opts.subnetFile, "FLANNEL_IPV6_SUBNET")
 
-	return etcd.NewLocalManager(ctx, cfg, prevSubnet, prevIPv6Subnet)
+	return etcd.NewLocalManager(ctx, cfg, prevSubnet, prevIPv6Subnet, opts.subnetLeaseRenewMargin)
 }
 
 func main() {
@@ -392,10 +392,10 @@ func main() {
 		log.Errorf("Failed to notify systemd the message READY=1 %v", err)
 	}
 
-	// Kube subnet mgr doesn't lease the subnet for this node - it just uses the podCidr that's already assigned.
-	if !opts.kubeSubnetMgr {
-		err = MonitorLease(ctx, sm, bn, &wg)
-		if err == errInterrupted {
+	err = sm.CompleteLease(ctx, bn.Lease(), &wg)
+	if err != nil {
+		log.Errorf("CompleteLease execute error err: %v", err)
+		if strings.EqualFold(err.Error(), errInterrupted.Error()) {
 			// The lease was "revoked" - shut everything down
 			cancel()
 		}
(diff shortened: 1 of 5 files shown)


subnet/etcd/local_manager.go
```diff
@@ -347,6 +347,10 @@ func (m *LocalManager) WatchLeases(ctx context.Context, cursor interface{}) (Lea
 	}
 }
 
+func (m *LocalManager) SetNodeNetworkUnavailableFalse(ctx context.Context) error {
```
manuelbuil, line 350: Can you add a comment on top of the function specifying that this function only makes sense for the kube subnet manager?

subnet/etcd/local_manager.go
```diff
@@ -347,6 +347,10 @@ func (m *LocalManager) WatchLeases(ctx context.Context, cursor interface{}) (Lea
...
+	return nil
```
manuelbuil, line 351: Could you return an error saying: unimplemented? For example: https://github.com/flannel-io/flannel/blob/master/subnet/kube/kube.go#L458

main.go
```diff
@@ -379,6 +379,16 @@ func main() {
...
+	// back off set node network unavailable for flannel cni bin read subnet.env file read error
+	if opts.setNodeNetworkUnavailable {
+		err = sm.SetNodeNetworkUnavailableFalse(ctx)
+		if err != nil {
+			log.Errorf("Unable to set NodeNetworkUnavailable to False: %v", err)
+		}
+	} else {
+		log.Infoln("Skip setting NodeNetworkUnavailable")
+	}
+
```
manuelbuil, line 391: It probably makes more sense to add it after the backend has already started to make sure everything is ready. Could you add it as part of the condition in line 405? Something like:
```
// Set NetworkUnavailable false in kube as flannel is ready or start monitoring lease for etcd
if opts.kubeSubnetMgr {
// your code
} else {
    err = MonitorLease(ctx, sm, bn, &wg)
    if err == errInterrupted {
    // The lease was "revoked" - shut everything down
        cancel()
     }
  }
```

subnet/subnet.go
```diff
@@ -148,5 +148,6 @@ type Manager interface {
 	WatchLease(ctx context.Context, sn ip.IP4Net, sn6 ip.IP6Net, cursor interface{}) (LeaseWatchResult, error)
 	WatchLeases(ctx context.Context, cursor interface{}) (LeaseWatchResult, error)
 
+	SetNodeNetworkUnavailableFalse(ctx context.Context) error
```
manuelbuil, line 151: I wonder if it is possible to create a function that "completes" the work for both subnet managers. For kube, it would set the networkunavailable condition to False and for etcd, it would start monitoring the leases. With the current PR, we would have both functions in main.go which I think is not the best approach :(. Besides, we have functions here that are implemented by only one subnetManager (SetNodeNetworkUnavailableFalse for kube and RenewLease or WatchLease for etcd). If we could move all these to a general function under the Manager called something like "CompleteNetworkConfig", that would look much better. Could you spend some time on this? We could help

manuelbuil (changes requested):
Thanks for the PR! I totally agree with it

subnet/etcd/local_manager.go
JornShen: look at new update

subnet/etcd/local_manager.go
JornShen: look at new update

main.go
JornShen: look at new update

subnet/subnet.go
JornShen: I create other function CompleteLease which  for kube, it would set the networkunavailable condition to False and for etcd, it would start monitoring the leases. look at new MR

JornShen:
/recc @manuelbuil 

I New a function `CompleteLease`. And move the MonitorLease to the etcd's LocalManager.  for kube, it would set Nodecondition.

PTAL :)

subnet/kube/kube.go
```diff
@@ -457,9 +449,14 @@ func (ksm *kubeSubnetManager) Name() string {
 	return fmt.Sprintf("Kubernetes Subnet Manager - %s", ksm.nodeName)
 }
 
-// Set Kubernetes NodeNetworkUnavailable to false when starting
+// CompleteLease Set Kubernetes NodeNetworkUnavailable to false when starting
 // https://kubernetes.io/docs/concepts/architecture/nodes/#condition
-func (ksm *kubeSubnetManager) setNodeNetworkUnavailableFalse(ctx context.Context) error {
+func (ksm *kubeSubnetManager) CompleteLease(ctx context.Context, lease *subnet.Lease, wg *sync.WaitGroup) error {
+	if !ksm.setNodeNetworkUnavailable {
```
manuelbuil, line 455: You forgot to call `err = ksm.setNodeNetworkUnavailableFalse(ctx)` when ksm.setNodeNetworkUnavailable = true

manuelbuil: You are right, I'm stupid

manuelbuil (approved):

Issue #1619, opened 2022-07-20
flannel FORWARD chain rules are placed after terminating user rules


PAI5REECHO:
The issue I have is that flannel inserts its iptables rules after some terminating user rules I've made for logging purposes.

## Expected Behavior
For flannel to provide some way to change the insertion order of its iptables rules in the FORWARD chain.

## Current Behavior
```console
# iptables -nvL FORWARD --line-numbers
Chain FORWARD (policy DROP 0 packets, 0 bytes)
num   pkts bytes target                  prot opt in     out     source               destination
1    40659 2294K KUBE-FORWARD            all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding rules */
2    40659 2294K KUBE-SERVICES           all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes service portals */
3    40295 2260K KUBE-EXTERNAL-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes externally-visible service portals */
4    40876 2293K LOG                     all  --  *      *       0.0.0.0/0            0.0.0.0/0            LOG flags 8 level 4 prefix "FORWARD (rejecting): "
5    40876 2293K REJECT                  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* Log all rejected packets */ reject-with icmp-host-prohibited
6        0     0 ACCEPT                  all  --  *      *       10.1.0.0/16          0.0.0.0/0            /* flanneld forward */
7        0     0 ACCEPT                  all  --  *      *       0.0.0.0/0            10.1.0.0/16          /* flanneld forward */
```

## Steps to Reproduce (for bugs)
1. Set a default DROP policy for all filter chains, and also add additional logging and rejection rules for the FORWARD chain.
2. flannel inserts it's rules AFTER the `REJECT` rules so they are never reached in the `FORWARD` chain

## Your Environment
* Flannel version: `v0.18.0`
* Backend used (e.g. vxlan or udp): I don't know how to check this
* Etcd version: `3.3.27`
* Kubernetes version (if used): `v1.23.8`
* Operating System and version: `Linux 5.18.10, NixOS, 22.11 (Raccoon), 22.11.20220619.0d68d7c`

rbrtbnfgl:
Flannel append its rule right after the all the iptables rule to avoid conflicts with user firewall configured rules.
[...]

Issue #1542, opened 2022-02-22
ordering of iptable rules


gfrankliu:
I see some discussions about iptable rules ordering, such as https://github.com/flannel-io/flannel/issues/1033

My host is on multiple networks and I have a specific iptable rule at the top of the FORWARD chain to block incoming from one NIC that is connected to the insecure network:
```
-A FORWARD -i ens192 -j REJECT --reject-with icmp-port-unreachable
```
When flannel is installed, the rule gets moved down a few positions, and my cluster becomes accessible from that NIC now, eg: clients on the insecure network can reach the NodePort on my host now.
Is there a way to talk flannel to leave my rule on the top?

manuelbuil:
Hello. Right now it is not possible to do this. However, if you add your rule after deploying flannel, it should stay in first position

Issue #1033, opened 2018-08-27
iptable forwarding rules added after default REJECT rule on CentOS 7


geauxvirtual:
When deploying flannel on a CentOS 7 host, the iptables forwarding rules for the pod network are added to the FORWARD chain after the default REJECT rule. This causes issues with internal traffic of a kubernetes cluster (pod to pod and pod to service) being rejected before the forwarding rules added by Flannel are hit.

![forward](https://user-images.githubusercontent.com/1395030/44684718-a3d17280-a9fe-11e8-9479-9fe11da8ecfa.JPG)

## Expected Behavior
Forwarding rules are added before any default DROP or REJECT rules in the CentOS FORWARD chain.

## Current Behavior
Forwarding rules are appended to the end of the FORWARD chain after the default DROP and REJECT rules.

## Steps to Reproduce (for bugs)
1. Install kubernetes 1.11.2 on CentOS 7 via kubeadm
2. Install flannel network -- `kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml`

## Context
No internal traffic in Kubernetes cluster (pod to pod and pod to service) works as traffic is rejected by a default REJECT rule in the FORWARD chain. 

## Your Environment

* Flannel version: 0.10.0
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version: 
* Kubernetes version (if used): 1.11.2
* Operating System and version: CentOS 7

vnlinuser:
I get same problem too...How fix it？

perrymana:
Looks like we are experiencing exactly the same issue with Flannel and CentOS. Manually re-ordering the rules so those 2 ACCEPT rules are above the REJECT rule (or turning the firewall off!) resolves the problem.

Did you find a workaround? I was about to see if I could manually declare some firewall rules to resolve it.

PR #1650, opened 2022-09-16
Added a different chain for iptables rules

rbrtbnfgl:
## Description

Added a different chain for Flannel iptables rules to order the rules before the one already present on the node.
It should solve #1619 #1542 #1033

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/network/iptables.go
+++ b/network/iptables.go
@@ -28,6 +28,7 @@ import (
 
 type IPTables interface {
 	AppendUnique(table string, chain string, rulespec ...string) error
+	ChainExists(table, chain string) (bool, error)
 	Delete(table string, chain string, rulespec ...string) error
 	Exists(table string, chain string, rulespec ...string) (bool, error)
 }
@@ -364,6 +365,25 @@ func teardownIPTables(ipt IPTables, iptr IPTablesRestore, rules []IPTablesRule) 
 
 	// Build delete rules to a transaction for iptables restore
 	for _, rule := range rules {
+		if rule.chain == "FLANNEL-FWD" || rule.rulespec[len(rule.rulespec)-1] == "FLANNEL-FWD" {
+			chainExists, err := ipt.ChainExists(rule.table, "FLANNEL-FWD")
+			if err != nil {
+				// this shouldn't ever happen
+				return fmt.Errorf("failed to check rule existence: %v", err)
+			}
+			if !chainExists {
+				continue
+			}
+		} else if rule.chain == "FLANNEL-POSTRTG" || rule.rulespec[len(rule.rulespec)-1] == "FLANNEL-POSTRTG" {
+			chainExists, err := ipt.ChainExists(rule.table, "FLANNEL-POSTRTG")
+			if err != nil {
+				// this shouldn't ever happen
+				return fmt.Errorf("failed to check rule existence: %v", err)
+			}
+			if !chainExists {
+				continue
+			}
+		}
 		exists, err := ipt.Exists(rule.table, rule.chain, rule.rulespec...)
 		if err != nil {
 			// this shouldn't ever happen
--- a/network/iptables.go
+++ b/network/iptables.go
@@ -39,16 +39,9 @@ type IPTablesError interface {
 
 type IPTablesRule struct {
 	table    string
+	action   string
 	chain    string
 	rulespec []string
-}
-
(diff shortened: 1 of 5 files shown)


network/iptables.go
```diff
@@ -135,7 +173,7 @@ func ipTablesRulesExist(ipt IPTables, rules []IPTablesRule) (bool, error) {
 }
 
 // ipTablesCleanAndBuild create from a list of iptables rules a transaction (as string) for iptables-restore for ordering the rules that effectively running
-func ipTablesCleanAndBuild(ipt IPTables, rules []IPTablesRule) (IPTablesRestoreRules, error) {
+func ipTablesCleanAndBuild(ipt IPTables, rules []IPTablesRule, toChain bool) (IPTablesRestoreRules, error) {
```
manuelbuil, line 176: Can you explain in the function comment what "toChain" mean?

network/iptables.go
```diff
@@ -43,6 +43,14 @@ type IPTablesRule struct {
...
 }
 
+var MasqChain = []IPTablesRule{
+	{"nat", "POSTROUTING", []string{"-m", "comment", "--comment", "flanneld masq", "-j", "FLANNEL-POSTRTG"}},
+}
+
+var FwdChain = []IPTablesRule{
+	{"filter", "FORWARD", []string{"-m", "comment", "--comment", "flanneld forward", "-j", "FLANNEL-FWD"}},
+}
+
```
manuelbuil, line 53: You can have both definitions in one var ( ...) section

network/iptables_windows.go
```diff
@@ -31,11 +31,21 @@ type IPTablesRule struct {
...
-func MasqIP6Rules(ipn ip.IP6Net, lease *subnet.Lease) []IPTablesRule { return nil }
-func DeleteIP4Tables(rules []IPTablesRule) error                     { return nil }
-func DeleteIP6Tables(rules []IPTablesRule) error                     { return nil }
+var MasqChain = []IPTablesRule{
+	{"nat", "POSTROUTING", []string{"-m", "comment", "--comment", "flanneld masq", "-j", "FLANNEL-POSTRTG"}},
+}
+
+var FwdChain = []IPTablesRule{
+	{"filter", "FORWARD", []string{"-m", "comment", "--comment", "flanneld forward", "-j", "FLANNEL-FWD"}},
+}
```
manuelbuil, line 40: You can define both variables in one var section

network/iptables_test.go
```diff
@@ -114,7 +114,7 @@ func TestDeleteRules(t *testing.T) {
 	baseRules := MasqRules(ip.IP4Net{}, lease())
 	expectedRules := expectedTearDownIPTablesRestoreRules(baseRules)
 
-	err := ipTablesBootstrap(ipt, iptr, baseRules)
+	err := ipTablesBootstrap(ipt, iptr, baseRules, false)
```
manuelbuil, line 117: All use `toChain=false`, is it possible to create one that uses `toChain=true`?

network/iptables.go
rbrtbnfgl: If the rule is used to send the traffic to the chain and it has to be added with -I and not -A.
Maybe I can modified to a more meaningful name.

network/iptables.go
```diff
@@ -88,34 +93,70 @@ func MasqIP6Rules(ipn ip.IP6Net, lease *subnet.Lease) []IPTablesRule {
...
+		log.Errorf("Failed to setup IPTables. Error on creating the chain: %v", err)
+		return
+	}
+}
+
+func CreateIP6Chain(table, chain string) {
+	ipt, err := iptables.NewWithProtocol(iptables.ProtocolIPv6)
+	if err != nil {
+		// if we can't find iptables, give up and return
+		log.Errorf("Failed to setup IPTables. iptables binary was not found: %v", err)
```
manuelbuil, line 152: Could you add v6 in the logs? Just to differentiate with the ipv4 case. Imagine we get an error in a dual-stack env, we would not know where it is coming from

manuelbuil (commented):
Looks good. Only one comment

manuelbuil (approved):

thomasferrandiz (approved):

PR #1652, opened 2022-09-22
Split subnet config

thomasferrandiz:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+type SubnetManagerType uint
+
+const (
+	EtcdType SubnetManagerType = iota
+	KubeType
+)
+
+type Config struct {
+	etcd.EtcdSubnetMngrConfig
+	kube.KubeSubnetMngrConfig
```
manuelbuil, line 34: Instead of this, I would add all the "official" config parameters that we document in the [docs](https://github.com/flannel-io/flannel/blob/master/Documentation/configuration.md#configuration) here:

```
type Config struct {
	EnableIPv4  bool
	EnableIPv6  bool
	BackendType string          `json:"-"`
	Backend     json.RawMessage `json:",omitempty"`
	Network       ip.IP4Net
	IPv6Network   ip.IP6Net
	SubnetMin     ip.IP4
	SubnetMax     ip.IP4
	IPv6SubnetMin *ip.IP6
	IPv6SubnetMax *ip.IP6
	SubnetLen     uint
	IPv6SubnetLen uint
}
```

In theory, all could be called when using kube or subnet, even though most of them don't make any sense in kube. Probably we should add a warning or something when using kube that they don't make sense

pkg/config/etcd/config.go
```diff
@@ -0,0 +1,160 @@
...
+	IPv6Network   ip.IP6Net
+	SubnetMin     ip.IP4
+	SubnetMax     ip.IP4
+	IPv6SubnetMin *ip.IP6
+	IPv6SubnetMax *ip.IP6
+	SubnetLen     uint
+	IPv6SubnetLen uint
+}
+
+func CheckConfig(esmConfig *EtcdSubnetMngrConfig, enableIPv4, enableIPv6 bool) error {
```
manuelbuil, line 36: I would replace this function with one called CheckConfig(....) which verifies common things for both etcd and kube subnet Manager. I think right now we can only check that there is a backend, that there is a backendType and that there is a Network.
The rest of the stuff, I would move it inside "GetNetworkConfig(ctx context.Context) (*config.Config, error)" of "type Manager interface", (probably in a different function which gets called by "GetNetworkConfig")

pkg/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+
+	if t == EtcdType {
+		err = etcd.CheckConfig(&cfg.EtcdSubnetMngrConfig, cfg.EnableIPv4, cfg.EnableIPv6)
+	} else if t == KubeType {
+		err = kube.CheckConfig(&cfg.KubeSubnetMngrConfig, cfg.EnableIPv4, cfg.EnableIPv6)
+	} else {
+		return nil, fmt.Errorf("unknown SubnetManagerType: %d", t)
+	}
```
manuelbuil, line 70: With my suggestions, I think we would not need this anymore

subnet/kube/kube.go
```diff
@@ -106,7 +107,7 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
 		return nil, fmt.Errorf("failed to read net conf: %v", err)
 	}
 
-	sc, err := subnet.ParseConfig(string(netConf))
+	sc, err := config.ParseConfig(string(netConf), config.KubeType)
```
manuelbuil, line 110: What about moving the "config.ParseConfig" to main.go? That way we have in main.go three steps:

1 - Configuration is parsed
2 - SubnetManager is started using the parsed config
3 - BackendManager is started using the parsed config + subnetManager to acquire the lease

pkg/config/kube/config.go
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import "github.com/flannel-io/flannel/pkg/ip"
+
+type KubeSubnetMngrConfig struct {
+	Networks     []ip.IP4Net
+	IPv6Networks []ip.IP6Net
```
manuelbuil, line 21: I guess you are preparing for the new kube API, right? I wonder if we should have both "Network" and "Networks" or just have Networks and check in etcd that it only has one element

Issue #1615, opened 2022-07-19
namespace changed to kube-flannel


nardusg:
Hi

Not a bug but more a question. I saw the namespace is now kube-flannel instead of kube-system, will this break upgrades ? Any extra steps need to be taken during upgrades ?

Regards,
Nardus

instantlinux:
In the file [kube-flannel-rbac.yml](https://github.com/flannel-io/flannel/blob/master/Documentation/k8s-manifests/kube-flannel-rbac.yml), the old namespace still appears at line 42. Shall I file a new issue, or can you re-open this one?

instantlinux:
PR [#1657](https://github.com/flannel-io/flannel/pull/1657) fixes that issue, if you'd like to review it.

PR #1657, opened 2022-10-07
Fix for #1615 namespace is now kube-flannel

instantlinux:
## Description

Documentation fix for k8s-manifest RBAC yaml.
This fixes the following error message:
```
Failed to create SubnetManager: error retrieving pod spec for 'kube-flannel/kube-flannel-ds-lkrmx':
 pods "kube-flannel-ds-lkrmx" is forbidden: User "system:serviceaccount:kube-flannel:flannel"
 cannot get resource "pods" in API group "" in the namespace "kube-flannel"
```
I have manually verified the fix allows flannel pods to bootstrap properly.
## Todos
- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Documentation/k8s-manifests/kube-flannel-rbac.yml
+++ b/Documentation/k8s-manifests/kube-flannel-rbac.yml
@@ -39,4 +39,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: flannel
-    namespace: kube-system
+    namespace: kube-flannel


rbrtbnfgl:
That's manifest is an old one. You have to look at kube-flannel.yaml that already implements the RBAC. https://github.com/flannel-io/flannel/blob/master/Documentation/kube-flannel.yml

instantlinux:
OK...so delete the outdated kube-flannel-rbac.yaml, please? It's still referenced in online searches, and breaks the installation if you follow some of the tutorials. Better for outdated files to be gone, making it obvious which of those tutorials are outdated.

rbrtbnfgl:
On the Documentation the kube-flannel-rbac.yaml is referred if you are using kubernetes v1.6~v1.15.
We had to move to kube-flannel namespace because PSP was removed from kubernetes v1.25. https://github.com/flannel-io/flannel/blob/master/Documentation/kubernetes.md#for-kubernetes--v124

PR #1663, opened 2022-10-20
fix: do not return "can't connect to etcd: no cert file found" error,…

ttys3:
fix: do not return "can't connect to etcd: no cert file found" error, allow connect with etcd without cert

force user to use a cert will breaking current exists cluster

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1667, opened 2022-10-28
don't throw an error when using http to connect to etcd 

thomasferrandiz:
## Description
- don't throw an error when using http to connect to etcd, log a warning instead
- upgrade go/text to fix CVE warning

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


rbrtbnfgl (approved):

ttys3:
this seems almost the same purpose as https://github.com/flannel-io/flannel/pull/1663

thomasferrandiz:
> this seems almost the same purpose as #1663

yes, it's the same idea but with a warning about using http instead of https.
If you want, you can add the warning in your PR and I remove it from this one.

Issue #1629, opened 2022-08-05
k8s IPv6 dual stack: Observed a panic: Address is not an IPv4 address


bingoct:
i setup k8s 1.23.9 by kubeadm.

i put ipv6 CIDR before ipv4 CIDR in kubeadm-config，and config kube-apiserver advertise-ip in host ipv6.
```yaml
controllerManager:
  extraArgs:
    allocate-node-cidrs: "true"
    bind-address: "::"
    service-cluster-ip-range: "fd00::1234:5678:1:0/112,10.96.0.0/12"
    cluster-cidr: "fd00::1234:5678:100:0/104,10.244.0.0/16"
    node-cidr-mask-size-ipv6: "120"
    node-cidr-mask-size-ipv4: "24"
```

kubectl logs -n kube-flannel kube-flannel-ds-z9ggx  -f
E0805 05:20:57.156568       1 runtime.go:76] Observed a panic: Address is not an IPv4 address

## Expected Behavior

kube-flannel  pods run

## Current Behavior

for k8s:
kube-flannel   kube-flannel-ds-z9ggx                           ●   0/1            5 CrashLoopBackOff  

for host: flannel interface create fail
 flannel.1: <BROADCAST,MULTICAST> mtu 1414 qdisc noop state DOWN group default
    link/ether f2:57:d9:14:3d:d2 brd ff:ff:ff:ff:ff:ff

## Steps to Reproduce (for bugs)

1. flannel config
```yaml
  net-conf.json: |
    {
      "Network": "10.244.0.0/16",
      "SubnetLen": 24,
      "IPv6Network": "fd00::1234:5678:100:0/104",
      "IPv6SubnetLen": 120,
      "Backend": {
        "Type": "vxlan"
      }
    }
```
## Context

if i only config flannel  ipv6 single  stack, it works.
```yaml
  net-conf.json: |
    {
      "EnableIPv4": false,
      "EnableIPv6": true,
      "IPv6Network": "fd00::1234:5678:100:0/104",
      "IPv6SubnetLen": 120,
      "Backend": {
        "Type": "vxlan"
      }
    }
```
However 

## Your Environment

* Flannel version: mirrored-flannelcni-flannel:v0.19.0
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version: 
* Kubernetes version (if used): 1.23.9
* Operating System and version: Centos 7
* Link to your project (optional):

rbrtbnfgl:
[...]

PR #1672, opened 2022-11-09
Fixed IPv4 podCIDR check in case spec.PodCIDR is IPv6

rbrtbnfgl:
## Description

When using kubeadm if the podSubnet is dualstack and specified as IPv6,IPv4 spec.PodCIDR of the node will be the IPv6 and Flannel will fail to check the IPv4 CIDR. (#1629).
This PR will remove the check of spec.PodCIDR to always check spec.PodCIDRs for both IPv4 and IPv6.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/subnet/kube/kube.go
+++ b/subnet/kube/kube.go
@@ -294,19 +294,16 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, attrs *subnet.Le
 	}
 
 	var cidr, ipv6Cidr *net.IPNet
-	_, cidr, err = net.ParseCIDR(n.Spec.PodCIDR)
-	if err != nil {
-		return nil, err
-	}
 
 	for _, podCidr := range n.Spec.PodCIDRs {
 		_, parseCidr, err := net.ParseCIDR(podCidr)
 		if err != nil {
 			return nil, err
 		}
-		if len(parseCidr.IP) == net.IPv6len {
+		if len(parseCidr.IP) == net.IPv4len {
+			cidr = parseCidr
+		} else if len(parseCidr.IP) == net.IPv6len {
 			ipv6Cidr = parseCidr
-			break
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


subnet/kube/kube.go
```diff
@@ -294,19 +294,27 @@ func (ksm *kubeSubnetManager) AcquireLease(ctx context.Context, attrs *subnet.Le
...
 
 	var cidr, ipv6Cidr *net.IPNet
-	_, cidr, err = net.ParseCIDR(n.Spec.PodCIDR)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, podCidr := range n.Spec.PodCIDRs {
-		_, parseCidr, err := net.ParseCIDR(podCidr)
+	if len(n.Spec.PodCIDRs) == 0 {
```
manuelbuil, line 297: What about adding a switch:
```
switch len(n.Spec.PodCIDRs) {
case 0:
     // do stuff with n.Spec.PodCIDR
case len(n.Spec.PodCIDRs) < 3:
     // do stuff with n.Spec.PodCIDRs
default:
    // fail!
```
That way we are protected of weird cases

subnet/kube/kube.go
```diff
@@ -430,9 +438,24 @@ func (ksm *kubeSubnetManager) nodeToLease(n v1.Node) (l subnet.Lease, err error)
 		}
 		l.Attrs.BackendData = json.RawMessage(n.Annotations[ksm.annotations.BackendData])
 
-		_, cidr, err := net.ParseCIDR(n.Spec.PodCIDR)
-		if err != nil {
-			return l, err
+		cidr := new(net.IPNet)
```
manuelbuil, line 441: I think it is cleaner to define it as it was in line 296

subnet/kube/kube.go
```diff
@@ -430,9 +438,24 @@ func (ksm *kubeSubnetManager) nodeToLease(n v1.Node) (l subnet.Lease, err error)
...
+		if len(n.Spec.PodCIDRs) == 0 {
```
manuelbuil, line 442: Same here with the switch

subnet/kube/kube.go
```diff
@@ -430,9 +438,24 @@ func (ksm *kubeSubnetManager) nodeToLease(n v1.Node) (l subnet.Lease, err error)
...
+			_, cidr, err = net.ParseCIDR(n.Spec.PodCIDR)
```
manuelbuil, line 443: Don't you need to check if this is ipv6 or ipv4?

subnet/kube/kube.go
```diff
@@ -430,9 +438,24 @@ func (ksm *kubeSubnetManager) nodeToLease(n v1.Node) (l subnet.Lease, err error)
...
+			if err != nil {
+				return l, err
+			}
+		} else {
+			log.Infof("Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: %v", n.Spec.PodCIDRs)
```
manuelbuil, line 448: This log will only be correct if len =1, right?

manuelbuil (changes requested):
I have just checked on 1.24 and podCIDRs == 0 is something that should never happen. It might be because of the old version of Hyperkube we are using, I'd recommend to first update that version and then see if this issue still exists

subnet/kube/kube.go
rbrtbnfgl: No because if IPv4 is enabled and no IPv4 CIDR is specified on the pod subnet Flannel should throw an error before the subnet manager starts.

manuelbuil (approved):

Issue #1679, opened 2022-11-21
60+  seconds stuck when call a http service pod


rkonfj:
when `flanneld` version upgrading to v0.20.1 and `curl` http  service pod in different node via `ClusterIP` will stuck 60+ seconds.
 
## Expected Behavior

no stuck
## Current Behavior

stuck 60+ seconds
## Possible Solution

eh...  may be caused by `double-NAT`, i have no idea

## Steps to Reproduce (for bugs)

it will stuck curl when `nat POSTROUTING` order like this:
```
Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         
FLANNEL-POSTRTG  all  --  anywhere             anywhere             /* flanneld masq */
KUBE-POSTROUTING  all  --  anywhere             anywhere             /* kubernetes postrouting rules */
```
it works fine like this:
```
Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         
KUBE-POSTROUTING  all  --  anywhere             anywhere             /* kubernetes postrouting rules */
FLANNEL-POSTRTG  all  --  anywhere             anywhere             /* flanneld masq */
```

## Context

this pr(https://github.com/kubernetes/kubernetes/pull/92035) looks like to solve this issue, but  I still have this problem when I use flanneld v0.20.1

## Your Environment

* Flannel version: v0.20.1
* Backend used (e.g. vxlan or udp): vxlan
* Etcd version: 3.5.3
* Kubernetes version (if used): v1.25.4
* Operating System and version: Archlinux (kernel version 6.0.8)
* Link to your project (optional):

rbrtbnfgl:
Hi @rkonfj thanks for reporting this.
I tried to setup a cluster with flannel with two pods deployed on two different nodes (one client and the other with nginx).
If I used wget on the exposed service address from the client pod the IPs are correctly translated by Kube-proxy and the connection works fine.
Could you give more info about your setup?
Can you check the output of `kubectl get service -A` and `kubectl get service -A`?

rkonfj:
@rbrtbnfgl pod to pod via service ip may works fine, but node to pod via service ip will be stuck

rbrtbnfgl:
I understand your issue but I still didn't get the same result.
I deployed the nginx pod on a node and I tried `curl` from a different node to the one where the pod was deployed and it worked.

rbrtbnfgl:
Could you share the content of `/etc/cni/net.d/10-flannel.conflist`?

[...]

PR #1681, opened 2022-11-23
Fixed masquerade rule to avoid double NAT bug

rbrtbnfgl:
## Description

Added iptables rule that check marked packed by Kube-proxy to avoid double-NAT that could slow down the connection or in some kernel versions loss packets.

Issue: #1679

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/network/iptables_test.go
+++ b/network/iptables_test.go
@@ -130,8 +130,8 @@ func TestDeleteRules(t *testing.T) {
 	if err != nil {
 		t.Error("Error setting up iptables")
 	}
-	if len(ipt.rules) != 5 {
-		t.Errorf("Should be 5 masqRules, there are actually %d: %#v", len(ipt.rules), ipt.rules)
+	if len(ipt.rules) != 6 {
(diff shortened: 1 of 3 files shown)


network/iptables_test.go
```diff
@@ -130,7 +130,7 @@ func TestDeleteRules(t *testing.T) {
 	if err != nil {
 		t.Error("Error setting up iptables")
 	}
-	if len(ipt.rules) != 5 {
+	if len(ipt.rules) != 6 {
 		t.Errorf("Should be 5 masqRules, there are actually %d: %#v", len(ipt.rules), ipt.rules)
```
thomasferrandiz, line 134: small note: the error message should be come:
`Should be 6 masqRules`

thomasferrandiz (changes requested):

manuelbuil (approved):

thomasferrandiz (approved):

Issue #1426, opened 2021-04-02
[Feature] Add Helm chart


nfisher:
Add helm chart configuration of the stock YAML.

## Expected Behavior

Install flannel with optional configuration parameters such as image overrides, etc.

```
helm install flannel \
  --repo https://flannel-io.github.io/helm \
  --namespace ${NAMESPACE} \
  flannel
```

## Current Behavior

Can only install flannel via Kubectl.

## Context

Helm provides a number of benefits over static YAML such as ease of templating for consumers and simplified upgrades.

## Your Environment

* Flannel version: latest
* Backend used (e.g. vxlan or udp): both
* Etcd version: NA
* Kubernetes version (if used): 1.17-1.20
* Operating System and version: Various releases of Ubuntu

kron4eg:
The `wontfix` is very unfortunate.

rbrtbnfgl:
The stale bot automatically set it for every old issues. At least now we have clearly all the needed issues opened and we can start work on them.

ethanbergstrom:
Thank you!

PR #1733, opened 2023-02-24
Add helm chart  to deploy flannel

rbrtbnfgl:
## Description

This PR is a WIP based on #1426 
The idea is to generate the helm chart on a tgz archive with the right version during release publishing it with the release files.
The command to install it through helm should be like this
```
helm install flannel --set podCidr="10.244.0.0/16" https://github.com/flannel-io/flannel/releases/latest/download/flannel.tgz
```

## Todos
- [x] Add helm chart skeleton
- [x] Create github action to generate tgz archive with release
- [x] Documentation

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ endif
 
 # Make a release after creating a tag
 # To build cross platform Docker images, the qemu-static binaries are needed. On ubuntu "apt-get install  qemu-user-static"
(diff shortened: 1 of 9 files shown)


thomasferrandiz (approved):
LGTM

PR #1799, opened 2023-08-31
fix: goroutine leak during watch leases (kube)

iiiceoo:
## Description

In the implementation of Kubernetes API, `leaseWatchChan` did not close correctly after ctx cancellation:

https://github.com/flannel-io/flannel/blob/148b6d6a58deb864519bbc474f4c531a0799e910/pkg/subnet/kube/kube.go#L482-L484

Will cause the following loop to not end:

https://github.com/flannel-io/flannel/blob/148b6d6a58deb864519bbc474f4c531a0799e910/pkg/subnet/subnet.go#L148-L165

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

Signed-off-by: iiiceoo <iiiceoo@foxmail.com>

### DIFF
--- a/pkg/subnet/etcd/subnet_test.go
+++ b/pkg/subnet/etcd/subnet_test.go
@@ -440,7 +440,7 @@ func TestRenewLease(t *testing.T) {
 		t.Fatal("RenewLease failed: ", err)
 	}
 	//we expect the new lease to have an expiration date in exactly 24h
-	acceptableMargin := 5 * time.Second
+	acceptableMargin := 10 * time.Second
 	expectedExpiration := time.Now().Add(subnetTTL).Round(time.Duration(acceptableMargin))
 
 	etcdResp, err := kvApi.Get(ctx, "/coreos.com/network/subnets", etcd.WithPrefix())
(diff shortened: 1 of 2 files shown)


iiiceoo:
When the CI was executing, an error unrelated to this PR was reported ([workflow](https://github.com/flannel-io/flannel/actions/runs/6033971119/job/16420683876?pr=1799)), the key logs are as follows:

```tex
I0901 14:38:03.171316    4220 subnet_test.go:451] found key: /coreos.com/network/subnets/10.3.16.0-24
    subnet_test.go:462: Failed to renew lease: bad expiration; expected 2023-09-02 14:38:05 +0000 UTC, got 2023-09-02 14:38:00 +0000 UTC
...
--- FAIL: TestRenewLease (11.04s)
```

This should be due to the fact that the `acceptableMargin` is too short and `kvApi.Get(...)` takes more time than `acceptableMargin/2`. I push another commit and fixed it by the way.

manuelbuil (approved):

rbrtbnfgl (approved):

PR #1808, opened 2023-09-18
Revert "fix: goroutine leak during watch leases (kube)"

rbrtbnfgl:
Reverts flannel-io/flannel#1799
With the previous PR the socket was closed when it's still used. Further investigations needed to check if it has to be closed.


manuelbuil (approved):

rbrtbnfgl:
The error seems not related to this.

PR #1829, opened 2023-11-23
feat: added persistent-mac option

### LINKED ISSUE #1828: Persistent MAC address for flannel.1 beyond reboots
We are using openyurt.io to manage edge nodes in our cluster. These edge nodes are small IoT devices like Raspberry PI that could reboot or be offline any time. In cases when the device reboots and is offline in order to maintain the containers on the edge flannel needs to be started with the same MAC address. We are using VXLAN.

## Expected Behavior

I propose a command line argument called `persistent-mac` that reads the nodes annotations where the MAC address is stored. Instead of creating a random MAC address, it re-uses the one read from the node annotations.
See https://openyurt.io/docs/user-manuals/network/edge-pod-network#background for further reference.

## Current Behavior

The current behavior re-generates a new MAC address on startup. This can be problematic for other nodes that want to reach the rebooted node. 

## Possible Solution

Like said, a flag that enables the reading of the nodes annotations to reinitialize with the known MAC address.

siredmar:
## Description
fixes #1828 

New feature: flag `persistent-mac` that reads the nodes annotation during startup and spins up the flannel.1 interface with the same MAC address. This can be useful when a reboot of the node happened but the old setup needs to be restored.

## Release Note

```release-note
The MAC address of the flannel interface is now persistent beyond restarts
```

### DIFF
--- a/pkg/subnet/kube/kube.go
+++ b/pkg/subnet/kube/kube.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -65,6 +66,7 @@ type kubeSubnetManager struct {
 	enableIPv4                bool
 	enableIPv6                bool
 	annotations               annotations
+	annotationPrefix          string
 	client                    clientset.Interface
 	nodeName                  string
 	nodeStore                 listers.NodeLister
@@ -160,6 +162,7 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
 func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet.Config, nodeName, prefix string, useMultiClusterCidr bool) (*kubeSubnetManager, error) {
 	var err error
 	var ksm kubeSubnetManager
+	ksm.annotationPrefix = prefix
 	ksm.annotations, err = newAnnotations(prefix)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


rbrtbnfgl:
Hi sorry I was on vacation the last few weeks. Why adding a flag on the cmdline to enable this? Could it be enough to only check the annotation and if it's configured it'll use the defined MAC?

siredmar:
@rbrtbnfgl no problem :)
Sure, this would also be possible but would break current behavior. 
If the new randomized MAC address after a flannel restart isn't by intention or not important at all this would be a nice solution.
What do you think?

thomasferrandiz:
One important thing to note is that the k8s client should not be imported in `main.go`.
All the k8s-related code should be in the `kube` package (https://github.com/flannel-io/flannel/tree/master/pkg/subnet/kube) since flannel can be used k8s.

siredmar:
@thomasferrandiz moved k8s related code 
@rbrtbnfgl removed persistent-flag argument

siredmar:
@thomasferrandiz  @rbrtbnfgl  not sure why the k3s e2e test fails... do you have any idea why?

thomasferrandiz:
I restarted the test and it's fine now.
It tends to fail sometimes, it's not related to your PR.

siredmar:
would you like to take a look again so we can merge this?

thomasferrandiz (approved):

manuelbuil:
By reading the release notes, it seems there is going to be a new flag, but I don't see it in the code.

Could you add in the docs how user could operate this new feature please? Thanks!

rbrtbnfgl:
I suggested to remove the flag. The feature should be enabled by default and it shouldn't break anything theoretically.

pkg/subnet/kube/kube.go
```diff
@@ -634,3 +637,53 @@ func (m *kubeSubnetManager) HandleSubnetFile(path string, config *subnet.Config,
 	}
 	return subnet.WriteSubnetFile(path, config, ipMasq, sn, ipv6sn, mtu)
 }
+
+// GetStoredMacAddress reads MAC address from node annotations when flannel restarts
+func (ksm *kubeSubnetManager) GetStoredMacAddress() string {
+	// The kube subnet mgr needs to know the k8s node name that it's running on so it can annotate it.
+	// If we're running as a pod then the POD_NAME and POD_NAMESPACE will be populated and can be used to find the node
+	// name. Otherwise, the environment variable NODE_NAME can be passed in.
+	nodeName := os.Getenv("NODE_NAME")
```
manuelbuil, line 646: `kubeSubnetManager struct` already includes the node (nodeName field), can't you use that one?

pkg/subnet/kube/kube.go
```diff
@@ -160,6 +162,7 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
 func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet.Config, nodeName, prefix string, useMultiClusterCidr bool) (*kubeSubnetManager, error) {
 	var err error
 	var ksm kubeSubnetManager
+	ksm.annotationPrefix = prefix
```
manuelbuil, line 165: Why do you need this? `ksm.annotations` is already included the prefix

manuelbuil (changes requested):

pkg/subnet/kube/kube.go
siredmar: you are absolutely right!

pkg/subnet/kube/kube.go
siredmar: I need to store the prefix that is passed newKubeSubnetManager in order to parse the annotations for the stored MAC address. Otherwise i would need to look at ksm.annotations and split any of them to get the prefix.

siredmar:
> By reading the release notes, it seems there is going to be a new flag, but I don't see it in the code.
> 
> Could you add in the docs how user could operate this new feature please? Thanks!

I changed the release notes

pkg/subnet/kube/kube.go
manuelbuil: ok, you have a point. Actually, it is strange that we allow to configure this

manuelbuil (approved):

siredmar:
@manuelbuil the e2e test failed again. Would you be so kind to restart it?

thomasferrandiz:
@siredmar could you squash the commits please?
Then we can merge the PR

siredmar:
@thomasferrandiz done

siredmar:
Thx

PR #1864, opened 2024-02-02
add nftables adr

thomasferrandiz:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil:
Looks good! Could we have a config option to enable this or not? So that we can give the option to use it instead of iptables or not

thomasferrandiz:
yes for sure we need a CLI flag and I don't think nftables will be the default for a long time

manuelbuil (approved):

PR #1865, opened 2024-02-07
Refactor iptables module to prepare for nftables implementation

thomasferrandiz:
## Description
This PR creates a trafficmngr package which declares the TrafficManager interface.
This interface holds the currently used chains and rules management functions for the iptables module.

This will allow us to also provide an nftables-based implementation of this interface.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/trafficmngr/trafficmngr.go
```diff
@@ -0,0 +1,39 @@
...
+	CreateIP4Chain(table, chain string)
+	CreateIP6Chain(table, chain string)
+	MasqRules(cluster_cidrs []ip.IP4Net, lease *lease.Lease) []IPTablesRule
+	MasqIP6Rules(cluster_cidrs []ip.IP6Net, lease *lease.Lease) []IPTablesRule
+	ForwardRules(flannelNetwork string) []IPTablesRule
+	SetupAndEnsureIP4Tables(getRules func() []IPTablesRule, resyncPeriod int)
+	SetupAndEnsureIP6Tables(getRules func() []IPTablesRule, resyncPeriod int)
+	DeleteIP4Tables(rules []IPTablesRule) error
+	DeleteIP6Tables(rules []IPTablesRule) error
+}
```
manuelbuil, line 39: I like the change but taking advantage of it, I think we can refactor code a little bit. I think it is possible to simplify this interface and have only:
* ForwardRulesToAccept (which would be called here: https://github.com/flannel-io/flannel/blob/master/main.go#L356-L362 and would hide most of that code)
* AddMasqRules (which would be called here https://github.com/flannel-io/flannel/blob/master/main.go#L340-L390 and would hide most of that code)

Then, the rest of the calls (recycle, which ends up calling Delete; CreateIPChain...) can be private functions inside the implementation of iptables or nftables. That would also make `main.go` a bit cleaner. Right now we are even defining functions https://github.com/flannel-io/flannel/blob/master/main.go#L356-L362 which I find very ugly.

What do you think?

manuelbuil (changes requested):
So there will be a

pkg/trafficmngr/trafficmngr.go
thomasferrandiz: I agree
I was planning to do that in another PR to keep things readable :-)

I can do it in one PR if you prefer

manuelbuil: that's also fine for me

manuelbuil (approved):

PR #1867, opened 2024-02-12
SubnetManager should use the main context

manuelbuil:
## Description

For some unknown reason, subnetManager was using a different context than the main one, thus the `cancel()` method to clean up flannel resources will never work in this case.

This PR fixes that and subnetManager will use the main context

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/backend/vxlan/vxlan.go
+++ b/pkg/backend/vxlan/vxlan.go
@@ -146,7 +146,7 @@ func (be *VXLANBackend) RegisterNetwork(ctx context.Context, wg *sync.WaitGroup,
 	// When flannel is restarted, it will get the MAC address from the node annotations to set flannel.1 MAC address
 	var hwAddr net.HardwareAddr
 
-	macStr := be.subnetMgr.GetStoredMacAddress()
+	macStr := be.subnetMgr.GetStoredMacAddress(ctx)
 	if macStr != "" {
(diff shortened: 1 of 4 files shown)


thomasferrandiz (approved):

thomasferrandiz:
I just noticed that this context is wrong as well: https://github.com/flannel-io/flannel/blob/f5b88d962d963cafac1ef91293608645597643cd/pkg/subnet/kube/kube.go#L612

`Context.TODO` is only for dev

manuelbuil:
> I just noticed that this context is wrong as well:
> 
> https://github.com/flannel-io/flannel/blob/f5b88d962d963cafac1ef91293608645597643cd/pkg/subnet/kube/kube.go#L612
> 
> `Context.TODO` is only for dev

Should it be Context.Background? I wonder if we could do something to pipe the main context

thomasferrandiz:
I think we can get the main context because `GetStoredMacAddress` is called in `RegisterNetwork` where it is available.

manuelbuil:
> I think we can get the main context because `GetStoredMacAddress` is called in `RegisterNetwork` where it is available.

done!

rbrtbnfgl (approved):

PR #1880, opened 2024-02-21
Use manual test to ensure iptables-* binaries are present

thomasferrandiz:
The sanity check in `iptables-wrapper-installer.sh` is important in case a version of the base image for some arch doesn't include all the needed iptables-related files.
But it fails for multi-arch build so we have to use a manual test.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -33,6 +33,9 @@ COPY --from=build /build/dist/flanneld /opt/bin/flanneld
 COPY dist/mk-docker-opts.sh /opt/bin/
 COPY --from=build /iptables-wrapper/iptables-wrapper-installer.sh /
 COPY --from=build /iptables-wrapper/bin/iptables-wrapper /
+# check manually that iptables-legacy and iptables-nft are present since
+# iptables-wrapper-installer.sh sanity check doesn't work for multi-arch build
+RUN which iptables-legacy && which iptables-nft
 RUN /iptables-wrapper-installer.sh --no-sanity-check
 
 


images/Dockerfile
```diff
@@ -33,6 +33,9 @@ COPY --from=build /build/dist/flanneld /opt/bin/flanneld
 COPY dist/mk-docker-opts.sh /opt/bin/
 COPY --from=build /iptables-wrapper/iptables-wrapper-installer.sh /
 COPY --from=build /iptables-wrapper/bin/iptables-wrapper /
+# check manually that iptables-legacy and iptables-nft are present since
+# iptables-wrapper-installer.sh sanity check doesn't work for multi-arch build
```
manuelbuil, line 37: but if you look at the last line, it says `--no-sanity-check`, are we really doing the sanity check?

manuelbuil (commented):
why does it fail in multi-arch?

images/Dockerfile
thomasferrandiz: no because it doesn't work.
Instead I use `which` to check that `iptables-nft` and `iptables-legacy` exist in the image

thomasferrandiz:
in a multi-arch build, `--no-sanity-check` fails for all arch except `amd64` even when `iptables-nft` and `iptables-legacy` are present

> why does it fail in multi-arch?

the script actually tries to run `iptables-legacy` and `iptables-nft`.
In non amd64 arch, `iptables-nft --version` fails with `protocol unsupported`, I guess because we are using qemu.

manuelbuil (approved):

PR #1881, opened 2024-02-22
Nftables implementation

thomasferrandiz:
## Todos
- [X] Tests
- [X] Documentation
- [X] Release note

## Release Note

```release-note
This PR adds experimental support for natively using nftables when masquerading traffic.
Set `"EnableNFTables": true` to activate this feature.
```


pkg/trafficmngr/nftables/nftables.go
```diff
@@ -0,0 +1,314 @@
...
+		cleanupCtx, cleanUpCancelFunc := context.WithTimeout(context.Background(), cleanUpDeadline*time.Second)
+		defer cleanUpCancelFunc()
+		err := nftm.cleanUp(cleanupCtx)
+		log.Errorf("nftables: error while cleaning-up: %v", err)
+		wg.Done()
+	}()
+	return nil
+}
+
+func initTable(ctx context.Context, ipFamily knftables.Family, name string) (knftables.Interface, error) {
```
manuelbuil, line 71: could you add a one-liner description for the function following [go best practices](https://tip.golang.org/doc/comment#:~:text=A%20function's%20doc%20comment%20should,any%20special%20syntax%20like%20backquotes.)

pkg/trafficmngr/nftables/nftables_windows.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/flannel-io/flannel/pkg/ip"
+	"github.com/flannel-io/flannel/pkg/lease"
+	"github.com/flannel-io/flannel/pkg/trafficmngr"
+)
+
+type NFTablesManager struct {
+}
+
+func (nftm *NFTablesManager) Init(ctx context.Context, wg *sync.WaitGroup) error {
+	return trafficmngr.ErrUnimplemented
```
manuelbuil, line 30: If we return an error here, will it not stop flanneld.exe? In that case, we'd better log the error and return nil

pkg/trafficmngr/nftables/utils.go
```diff
@@ -0,0 +1,57 @@
...
+
+	log "k8s.io/klog/v2"
+	"sigs.k8s.io/knftables"
+)
+
+const (
+	masqueradeTestTable = "masqueradeTest"
+)
+
+func (nftm *NFTablesManager) checkRandomfully(ctx context.Context) bool {
```
manuelbuil, line 30: If you could add one-liner describing this function, that'd be great

pkg/trafficmngr/iptables/iptables_windows.go
```diff
@@ -29,14 +33,18 @@ type IPTables interface {
 	Exists(table string, chain string, rulespec ...string) (bool, error)
 }
 
-func (iptm IPTablesManager) SetupAndEnsureForwardRules(flannelIPv4Network ip.IP4Net, flannelIPv6Network ip.IP6Net, resyncPeriod int) {
+func (iptm IPTablesManager) Init(ctx context.Context, wg *sync.WaitGroup) error {
+	return trafficmngr.ErrUnimplemented
```
manuelbuil, line 37: same as in nftables, I am not 100% sure but I fear that if we return an error, flanneld.exe will stop immediatly

manuelbuil (commented):
looks very good!

Makefile
```diff
@@ -5,7 +5,7 @@ REGISTRY?=quay.io/coreos/flannel
 QEMU_VERSION=v3.0.0
 
 # Default tag and architecture. Can be overridden
-TAG?=$(shell git describe --tags --always)
+TAG?=$(shell git describe --tags --dirty --always)
```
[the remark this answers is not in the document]
thomasferrandiz: It's needed to be coherent with the version used in the tests: https://github.com/flannel-io/flannel/blob/faff867936f4dc934367cbe98918ca444e342326/e2e/version.sh#L5

This is only visible when running the e2e tests locally but I should have aligned them when I added them the k3s-based e2e tests.

e2e/run-e2e-tests.sh
```diff
@@ -150,32 +157,50 @@ test_vxlan() {
...
+    delete-flannel
+    check_iptables_removed
+}
+
+test_vxlan_nft() {
+    prepare_test vxlan true
+    pings
+    check_nftables
+    delete-flannel
+    # check_nftables_removed
```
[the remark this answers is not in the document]
thomasferrandiz: that's a mistake...

pkg/trafficmngr/nftables/nftables.go
thomasferrandiz: done

pkg/trafficmngr/nftables/utils.go
thomasferrandiz: done

main.go
```diff
@@ -336,7 +337,15 @@ func main() {
 	}
 
 	//Create TrafficManager and instanciate it based on whether we use iptables or nftables
-	trafficMngr := newTrafficManager()
+	trafficMngr := newTrafficManager(config.EnableNFTables)
+	err = trafficMngr.Init(ctx, &wg)
```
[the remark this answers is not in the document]
thomasferrandiz: I think it's done by the library that we use to manipulate nftables.
When we try to create the Tables during the init procedure, if there is no nftables support, the library will return an error.

manuelbuil (approved):

dereknola (approved):
LGTM

pkg/trafficmngr/nftables/nftables.go
```diff
@@ -0,0 +1,317 @@
...
+	if !flannelIPv6Network.Empty() {
+		log.Infof("Changing default FORWARD chain policy to ACCEPT (ipv6)")
+		tx := nftm.nftv6.NewTransaction()
+
+		//TODO how to express that the default is drop?
+		tx.Add(&knftables.Chain{
+			Name:     forwardChain,
+			Comment:  knftables.PtrTo("chain to accept flannel traffic"),
+			Type:     knftables.PtrTo(knftables.FilterType),
+			Hook:     knftables.PtrTo(knftables.InputHook),
```
rbrtbnfgl, line 137: Should it be on ForwardHook?

pkg/trafficmngr/nftables/nftables.go
```diff
@@ -0,0 +1,317 @@
...
+	if !flannelIPv4Network.Empty() {
+		log.Infof("Changing default FORWARD chain policy to ACCEPT")
+		tx := nftm.nftv4.NewTransaction()
+
+		//TODO how to express that the default is drop?
+		tx.Add(&knftables.Chain{
+			Name:     forwardChain,
+			Comment:  knftables.PtrTo("chain to accept flannel traffic"),
+			Type:     knftables.PtrTo(knftables.FilterType),
+			Hook:     knftables.PtrTo(knftables.InputHook),
```
rbrtbnfgl, line 102: Should it be on ForwardHook?

pkg/trafficmngr/nftables/nftables.go
```diff
@@ -0,0 +1,317 @@
...
+
+// It is needed when using nftables? accept seems to be the default
+// warning: never add a default 'drop' policy on the forwarChain as it breaks connectivity to the node
+func (nftm *NFTablesManager) SetupAndEnsureForwardRules(ctx context.Context,
+	flannelIPv4Network ip.IP4Net, flannelIPv6Network ip.IP6Net, resyncPeriod int) {
+	if !flannelIPv4Network.Empty() {
+		log.Infof("Changing default FORWARD chain policy to ACCEPT")
+		tx := nftm.nftv4.NewTransaction()
+
+		//TODO how to express that the default is drop?
```
rbrtbnfgl, line 97: Do we need to DROP by default? I think we can add a simple "drop" rule at the end of the chain. I saw that on the class the `policy` value is missing it should be something like this `type filter hook forward priority -100; policy drop;`

rbrtbnfgl: Yes I was confused by the question. It shouldn't drop but it should continue the chain in case it's not matching the flannel subnet using the default configuration of the node.

thomasferrandiz: I think that's the default behavior so it should be fine as it is.
From the comment in flannel code, the forward rules seem needed for old versions of docker anyway.

PR #1890, opened 2024-02-27
Added additional check for n.spec.podCIDRs

rbrtbnfgl:
## Description

Added additional chek to verify that `n.spec.podCIDRs` is rightly configured on the node
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil:
Did you find a config where this happens?

rbrtbnfgl:
Here https://github.com/k3s-io/k3s/issues/9569
It shouldn't happen when flannel runs as a pod because the check is done before. When using K3s that directly calls the flannel function it misses the other checks.

thomasferrandiz (approved):

manuelbuil (approved):

PR #1907, opened 2024-03-15
Add tolerations to the helm chart

manuelbuil:
## Description

Flannel might deploy in clusters whose nodes contain taints and thus we need the tolerations field to be configurable

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/chart/kube-flannel/templates/daemonset.yaml
+++ b/chart/kube-flannel/templates/daemonset.yaml
@@ -27,9 +27,10 @@ spec:
                 - linux
       hostNetwork: true
       priorityClassName: system-node-critical
+      {{- with .Values.flannel.tolerations }}
       tolerations:
-      - operator: Exists
-        effect: NoSchedule
+        {{- toYaml . | trim | nindent 8 }}
+      {{- end }}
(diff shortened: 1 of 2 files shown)


thomasferrandiz (approved):

mgfritch (approved):

mgfritch (approved):

fungaren:
This commit introduced a bug that not all taints are tolerated, at least the default taint for master node `node-role.kubernetes.io/master:NoSchedule` is not tolerated.

Please be careful if you are changing the behavior.

brandond:
@fungaren the `node-role.kubernetes.io/master` label has been deprecated for quite some time; in what environment do you still see this taint being set?

fungaren:
@brandond I have an old Kubernetes cluster which is using 1.24. While upgrading flannel version the problem appeared. Now I have decide to add `--set flannel.tolerations[0].effect=NoSchedule` to address it. 

Maybe it should not be considered as a bug, but it is indeed a breaking change.

PR #1925, opened 2024-03-26
Restore default chart behaviour

manuelbuil:
## Description

This PR restores the default chart behavior which was tolerating all kind of taints. It even expands it to "NoSchedule" too

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The flannel chart now tolerates all kind of taints
```

### DIFF
--- a/chart/kube-flannel/values.yaml
+++ b/chart/kube-flannel/values.yaml
@@ -54,13 +54,7 @@ flannel:
   # General daemonset configs
   #
   tolerations:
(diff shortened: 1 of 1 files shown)


brandond (approved):
💯 💯 💯

rbrtbnfgl (approved):

rbrtbnfgl:
I don't know if we can create a release only for the chart. Right now a new chart is generated when a new flannel release is created

PR #1929, opened 2024-04-04
troubleshooting.md: add `ethtool -K flannel.1 tx-checksum-ip-generic off` for NAT

AkihiroSuda:
## Description

When the public IP is behind NAT, the UDP checksum fields of the VXLAN packets can be corrupted. In that case, try running the following commands to avoid corrupted checksums:

```bash
/usr/sbin/ethtool -K flannel.1 tx-checksum-ip-generic off
```

To automate the command above via udev, create `/etc/udev/rules.d/90-flannel.rules` as follows:

```
SUBSYSTEM=="net", ACTION=="add|change|move", ENV{INTERFACE}=="flannel.1", RUN+="/usr/sbin/ethtool -K flannel.1 tx-checksum-ip-generic off"
```

ref:
- flannel-io/flannel#1279
- kubernetes/kops#9074
- karmab/kcli@b1a8eff658d17cf4e28162f0fa2c8b2b10e5ad00

## Todos
(None)

## Release Note

```release-note
None required
```


manuelbuil:
Hey, thanks for the PR. This is a workaround for a bug in some kernels or? If we do this, we would be creating a penalty in the kernels which fixed this, or?

AkihiroSuda:
> Hey, thanks for the PR. This is a workaround for a bug in some kernels or? If we do this, we would be creating a performance penalty in the kernels which fixed this, or?

I'm not sure if this is a bug or a designed behavior on the kernel's side, but I guess the behavior may potentially change in a future version of kernel, perhaps with some sysctl.

So I added the command only in `troubleshooting.md`.

manuelbuil:
> > Hey, thanks for the PR. This is a workaround for a bug in some kernels or? If we do this, we would be creating a performance penalty in the kernels which fixed this, or?
> 
> I'm not sure if this is a bug or a designed behavior on the kernel's side, but I guess the behavior may potentially change in a future version of kernel, perhaps with some sysctl.
> 
> So I added the command only in `troubleshooting.md`.

Yes, it is supposed to be a kernel problem: https://github.com/kubernetes/kubernetes/issues/88986#issuecomment-635367089 but TBH, I haven't seen a real fix in those versions stated in the comment

manuelbuil (approved):

thomasferrandiz:
@AkihiroSuda can you please fix the merge conflict ?

AkihiroSuda:
> @AkihiroSuda can you please fix the merge conflict ?

done

manuelbuil (approved):

PR #1933, opened 2024-04-09
Added leaseAttr string method with logs on VxLan

rbrtbnfgl:
## Description

Updated lease.go with String function for leaseAttr and added a log on vxlan subnet handler with the attr info.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil:
When using vxlan, we will get two logs with the lease,right? The one we get right now with a lot of numbers to decode and this one, right?

rbrtbnfgl:
Yes

thomasferrandiz (approved):

manuelbuil (approved):

Issue #1939, opened 2024-04-12
Improve windows logs when it comes to iptables


manuelbuil:
## Current Behavior

We see the following logs:
```
I0412 09:55:06.903005    4612 main.go:567] Starting flannel in iptables mode
W0412 09:55:06.903005    4612 iptables_windows.go:39] unimplemented
W0412 09:55:06.903005    4612 main.go:530] no subnet found for key: FLANNEL_IPV6_NETWORK in file: /run/flannel/subnet.env
W0412 09:55:06.903005    4612 main.go:530] no subnet found for key: FLANNEL_IPV6_SUBNET in file: /run/flannel/subnet.env
W0412 09:55:06.903005    4612 iptables_windows.go:50] unimplemented
I0412 09:55:06.903005    4612 kube.go:476] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [10.42.2.0/24]
```
which could be considered alarming by users when reading `unimplemented` or reading `iptables mode` even if they are in Windows

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Flannel version:
* Backend used (e.g. vxlan or udp):
* Etcd version:
* Kubernetes version (if used):
* Operating System and version:
* Link to your project (optional):

manuelbuil:
It's done

PR #1947, opened 2024-04-17
Bug fixes

thomasferrandiz:
- show content of stdout and stderr when running iptables-restore returns an error
- fix comparison with previous networks in SetupAndEnsureMasqRules
- add empty WindowsTrafficManager to make logs clearer

Related issue: https://github.com/flannel-io/flannel/issues/1939

### DIFF
--- a/main.go
+++ b/main.go
@@ -561,11 +561,8 @@ func ReadIP6CIDRsFromSubnetFile(path string, CIDRKey string) []ip.IP6Net {
 
 func newTrafficManager(useNftables bool) trafficmngr.TrafficManager {
 	if useNftables {
-		log.Info("Starting flannel in nftables mode")
 		return &nftables.NFTablesManager{}
 	} else {
-		log.Info("Starting flannel in iptables mode")
 		return &iptables.IPTablesManager{}
-
 	}
 }
--- a/pkg/trafficmngr/iptables/iptables.go
+++ b/pkg/trafficmngr/iptables/iptables.go
@@ -48,6 +48,8 @@ type IPTablesManager struct {
 }
 
 func (iptm *IPTablesManager) Init(ctx context.Context, wg *sync.WaitGroup) error {
+	log.Info("Starting flannel in iptables mode...")
+
 	iptm.ipv4Rules = make([]trafficmngr.IPTablesRule, 0, 10)
 	iptm.ipv6Rules = make([]trafficmngr.IPTablesRule, 0, 10)
 	wg.Add(1)
--- a/pkg/trafficmngr/iptables/iptables.go
+++ b/pkg/trafficmngr/iptables/iptables.go
@@ -109,7 +109,7 @@ func (iptm *IPTablesManager) SetupAndEnsureMasqRules(ctx context.Context, flanne
 
 	if !flannelIPv4Net.Empty() {
 		// recycle iptables rules only when network configured or subnet leased is not equal to current one.
-		if prevNetwork != flannelIPv4Net && prevSubnet != currentlease.Subnet {
+		if !(flannelIPv4Net.Equal(prevNetwork) && prevSubnet.Equal(currentlease.Subnet)) {
 			log.Infof("Current network or subnet (%v, %v) is not equal to previous one (%v, %v), trying to recycle old iptables rules",
 				flannelIPv4Net, currentlease.Subnet, prevNetwork, prevSubnet)
 			newLease := &lease.Lease{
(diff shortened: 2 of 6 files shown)


rbrtbnfgl (approved):

main.go
```diff
@@ -560,12 +562,15 @@ func ReadIP6CIDRsFromSubnetFile(path string, CIDRKey string) []ip.IP6Net {
 }
 
 func newTrafficManager(useNftables bool) trafficmngr.TrafficManager {
+	if runtime.GOOS == "windows" {
+		log.Info("Starting flannel in windows mode...")
+		return &windows.WindowsManager{}
+	}
 	if useNftables {
-		log.Info("Starting flannel in nftables mode")
+		log.Info("Starting flannel in nftables mode...")
```
manuelbuil, line 570: Why the three dots?

pkg/trafficmngr/iptables/iptables.go
```diff
@@ -109,7 +109,7 @@ func (iptm *IPTablesManager) SetupAndEnsureMasqRules(ctx context.Context, flanne
 
 	if !flannelIPv4Net.Empty() {
 		// recycle iptables rules only when network configured or subnet leased is not equal to current one.
-		if prevNetwork != flannelIPv4Net && prevSubnet != currentlease.Subnet {
+		if !(flannelIPv4Net.Equal(prevNetwork) && prevSubnet.Equal(currentlease.Subnet)) {
```
manuelbuil, line 112: Instead of "&&", shouldn't it be "||"? I understand you want to recreate the iptables rules if any changes

pkg/trafficmngr/iptables/iptables.go
```diff
@@ -126,7 +126,7 @@ func (iptm *IPTablesManager) SetupAndEnsureMasqRules(ctx context.Context, flanne
 	}
 	if !flannelIPv6Net.Empty() {
 		// recycle iptables rules only when network configured or subnet leased is not equal to current one.
-		if prevIPv6Network != flannelIPv6Net && prevIPv6Subnet != currentlease.IPv6Subnet {
+		if !(flannelIPv6Net.Equal(prevIPv6Network) && prevIPv6Subnet.Equal(currentlease.IPv6Subnet)) {
```
manuelbuil, line 129: Same

main.go
```diff
@@ -560,12 +562,15 @@ func ReadIP6CIDRsFromSubnetFile(path string, CIDRKey string) []ip.IP6Net {
 }
 
 func newTrafficManager(useNftables bool) trafficmngr.TrafficManager {
+	if runtime.GOOS == "windows" {
+		log.Info("Starting flannel in windows mode...")
+		return &windows.WindowsManager{}
+	}
```
manuelbuil, line 568: Instead of doing this and creating a new file, why don't you just increase the verbosity required for the `ErrUnimplemented` error to be logged and return nil? I think this is too much just to make the logs clearer

main.go
thomasferrandiz: it's just to follow the convention when you put a log message before an operation that can take some time.

pkg/trafficmngr/iptables/iptables.go
thomasferrandiz: that's what it should do because there's the negation ! before

I think before it was wrong.
It can be either `! (a && b)` or `!a || !b`.
I can change it to the second option if you find it more readable.

main.go
thomasferrandiz: My reasoning is that in this way I never display the "Starting flannel in iptables mode" message which was confusing users.
Also I hoped that it would allow me to delete the `iptables_windows.go` file but it's still needed to compile on windows.

I was also thinking since we have more requests for flannel on windows there could at some point be something to do in the windows traffic manager even it does't use iptables or nftables.

pkg/trafficmngr/iptables/iptables.go
manuelbuil: you are right, sorry I thought it was negating each!

main.go
manuelbuil: My main problem is that I find it confusing having so many windows files around. Why can't we remove `iptables-_windows.go` and `nftables_windows.go`?

I don't think we will ever need Windows doing anything around traffic manager because the forwarding is always there and the natting is also default when creating the overlay network.

main.go
manuelbuil: I did not know about this convention, thanks!

main.go
thomasferrandiz: If we remove those files then the windows builds fail because all the other files in the package have this constraint:
```
//go:build !windows
// +build !windows
```

I have another idea though: instead of having a dedicated windows package which does nothing, I can move the log line inside the traffic manager Init methds so  that the log can be different based on the OS.

manuelbuil: good idea!

manuelbuil (approved):

PR #2019, opened 2024-07-17
docs: Fix Kubernetes integration link

rajiv-k:
The old link was returning 404.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


rbrtbnfgl:
Hi thanks for the PR. But I think that it's not the right docs page where to look at. The older page was removed I'll update this integration page with different Projects.

PR #2179, opened 2025-02-05
Refactor: Rename SetUpNetlinkTest to TestSetUpNetlink for consistency

pingora-gateway:
In pkg/ns/ns.go, which is a test file, the function SetUpNetlinkTest is actually a test case. To adhere to Go's testing framework conventions and improve code readability, we rename it to TestSetUpNetlink.

The previous name did not follow the standard naming convention for test functions, which should start with 'Test' to be recognized by the testing framework.

Changes:
- Renamed SetUpNetlinkTest to TestSetUpNetlink.
- Updated the function signature to match Go's testing framework requirements (no return value).

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/backend/route_network_test.go
+++ b/pkg/backend/route_network_test.go
@@ -22,13 +22,10 @@ import (
 
 	"github.com/flannel-io/flannel/pkg/ip"
 	"github.com/flannel-io/flannel/pkg/lease"
-	"github.com/flannel-io/flannel/pkg/ns"
 	"github.com/vishvananda/netlink"
 )
 
 func TestRouteCache(t *testing.T) {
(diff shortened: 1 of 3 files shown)


manuelbuil:
Are you sure about this? I understood that Go naming convention states that files with test functions should end with "_test.go". Example: https://go.dev/doc/tutorial/add-a-test

mahmut-Abi:
Thanks for pointing that out! According to Go's naming conventions, test files should indeed end with _test.go. I'll double-check and ensure all test files follow this convention. If there are any discrepancies, I'll fix them promptly.

PR #2191, opened 2025-02-27
enable setting CNI directory paths in helm chart

killianmuldoon:
enable setting the location of the CNI directories in the helm chart.

```release-note
None required
```


rbrtbnfgl (approved):
LGTM

PR #2200, opened 2025-03-10
Use peer's wireguard port, not our own

euank:
## Description

Before this change, the wireguard code constructed a peer endpoint via "PeerIP + n.dev.listenPort", i.e. we used the peer's IP, but our port.

This works fine if every k8s node has the same ListenPort, which is admittedly the common setup...

However, some people may desire to use different ports for some cases, and in that case, we should respect that different port.

I've manually tested that this works in my cluster, where 1 node has different ports for wireguard from all the others.

After deploying this diff to each node, I get a working pod network.

Note: I believe this is a bugfix which shouldn't break any properly configured setups, so I didn't add an opt-in or opt-out to it. I'm happy to add an opt-in flag if we can think of any reasonable setup this would break.

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/backend/wireguard/wireguard_network.go
+++ b/pkg/backend/wireguard/wireguard_network.go
@@ -101,10 +101,10 @@ func (n *network) Run(ctx context.Context) {
 
 type wireguardLeaseAttrs struct {
 	PublicKey string
+	Port      uint16
 }
 
-// Select the endpoint address that is most likely to allow for a successful
-// connection.
+// Select the mode that is most likely to allow for a successful connection.
 // If both ipv4 and ipv6 addresses are provided:
 //   - Prefer ipv4 if the remote endpoint has a public ipv4 address
 //     and the external iface has an ipv4 address as well. Anything with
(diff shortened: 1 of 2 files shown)


pkg/backend/wireguard/wireguard_network.go
```diff
@@ -209,14 +212,17 @@ func (n *network) handleSubnetEvents(ctx context.Context, batch []lease.Event) {
 				}
 			} else {
 				var publicEndpoint string
-				if n.mode == Ipv4 {
-					publicEndpoint = fmt.Sprintf("%s:%d", event.Lease.Attrs.PublicIP.String(), n.dev.attrs.listenPort)
-				} else if n.mode == Ipv6 {
-					publicEndpoint = fmt.Sprintf("[%s]:%d", event.Lease.Attrs.PublicIPv6.String(), n.dev.attrs.listenPort)
```
euank, line 215: This line had an additional bug of using the ipv6 IP + ipv4 port, so I think this probably wouldn't generally work.

euank:
I've pushed an update which should fix the e2e failures I was getting before.

I was dereferencing `n.v6Dev` without a `nil` check before, even when `EnableIPv6` wasn't set... which of course meant flannel paniced in the e2e test.

My local cluster is dual-stack (ipv4+6 on all nodes), so I didn't get that in my own local usage, and there aren't any unit tests in the wireguard package to catch it.

I think that's all sorted now! The e2e tests pass locally, and I've updated my cluster to the current commit, and all my pod networking is still happy.

PR #2239, opened 2025-05-28
fix: clean-up rules when starting instead of shutting down

thomasferrandiz:
## Description
This avoids traffic interruption while the flannel pods restart and still allows us to clean-up when switching between iptables and nftables.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


rbrtbnfgl (approved):

pkg/trafficmngr/iptables/iptables.go
```diff
@@ -47,59 +46,48 @@ type IPTablesManager struct {
...
-			log.Errorf("Failed to tear down IPTables: %v", err)
-		}
+	err = ipt.ClearAndDeleteChain("nat", "FLANNEL-POSTRTG")
+	if err != nil {
+		log.V(2).Infof("could not clean-up FLANNEL-POSTRTG (IPv4): %v", err)
+	}
+	err = ipt.ClearAndDeleteChain("nat", "FLANNEL-FWD")
+	if err != nil {
+		log.V(2).Infof("could not clean-up FLANNEL-FWD (IPv4): %v", err)
+	}
```
manuelbuil, line 73: If we can't clean the rules up, we are only logging an error but Flannel will still start, right? Was this the behavior we had before forcing the rules to be deleted?

manuelbuil (changes requested):
We used to give a `cleanUpDeadline`. Now there is none, I guess it's fine and it will not take ages to delete?

thomasferrandiz:
> We used to give a `cleanUpDeadline`. Now there is none, I guess it's fine and it will not take ages to delete?

I didn't use much time when I tested.
I just added a context with deadline before because I couldn't use the main context while shutting down.

pkg/trafficmngr/iptables/iptables.go
thomasferrandiz: we have to continue because we could be trying to clean-up iptables rules on a system that has only nftables so flannel would fail to start

manuelbuil (approved):

PR #2248, opened 2025-06-12
capture close() err on subnet file save

ahmetb:
## Description

The OS may accept write()s but may fail on close(), which is currently ignored.
Because it's ignored, it won't be logged anywhere either. (I'm sending this PR because I'm seeing missing subnet.env file on disk, but there's no `Failed to write subnet file`).

## Todos
- [ ] Tests: N/A
- [ ] Documentation: N/A
- [ ] Release note: N/A

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/subnet/subnet.go
+++ b/pkg/subnet/subnet.go
@@ -15,6 +15,7 @@
 package subnet
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


manuelbuil (approved):

pkg/subnet/subnet.go
```diff
@@ -94,10 +94,12 @@ func WriteSubnetFile(path string, config *Config, ipMasq bool, sn ip.IP4Net, ipv
 
 	fmt.Fprintf(f, "FLANNEL_MTU=%d\n", mtu)
 	_, err = fmt.Fprintf(f, "FLANNEL_IPMASQ=%v\n", ipMasq)
-	f.Close()
 	if err != nil {
 		return err
 	}
+	if err := f.Close(); err != nil {
+		return er
```
manuelbuil, line 101: it should be err

pkg/subnet/subnet.go
```diff
@@ -94,10 +94,12 @@ func WriteSubnetFile(path string, config *Config, ipMasq bool, sn ip.IP4Net, ipv
 
 	fmt.Fprintf(f, "FLANNEL_MTU=%d\n", mtu)
 	_, err = fmt.Fprintf(f, "FLANNEL_IPMASQ=%v\n", ipMasq)
-	f.Close()
 	if err != nil {
 		return err
 	}
+	if err := f.Close(); err != nil {
```
manuelbuil, line 100: If you move `f.Close()` after the return, it is possible that we never reach f.Close() and we leave the file open. In this case, I think the best is to create a function that we can "defer" where the file is closed and we assess if there is an error while closing the file

manuelbuil (changes requested):
Couple of things

pkg/subnet/subnet.go
ahmetb: I think each of these `fmt.Fprintf(`s deserve a err handler, instead I'm thinking of using a buffer to write this in one shot.

I'll add a close() whose err is not handled to capture the leaks.

manuelbuil (approved):

PR #2251, opened 2025-06-19
Fix deadlock in startup for large clusters

sudheerv:
## Description

Issue: #2252 

Channel event queue is a sized buffer (default=5000) and ends up blocking if there are events that are higher than the buffer size. This can happen on a large cluster with 5K nodes. Flannel sets up the informers during startup and while the informer callbacks occur asynchronously, NewSubnetManager uses a wait.PollUntilContextTimeout() to check for informer sync completion (with a timeout of 10 min). The problem is AddEvent blocks when the channel buffer is full holding the DeltaFIFO lock that is also needed by cache controller's HasSynced() which is checked in the callback for wait.PollUntilContextTimeout(). This results in a deadlock where the main thread is indefinitely blocked.

Below is the summary from stacktrace on a stuck flanneld due to this issue.

| Goroutine | Waiting On                                          | Why                                 |
| --------: | --------------------------------------------------- | ----------------------------------- |
|    **81** | Sending to a **channel** in `handleAddLeaseEvent()` | Channel is full or receiver is gone |
|         1 | `.HasSynced()` → needs lock                         | Main goroutine, can't proceed       |
|        88 | `.Resync()` → needs lock                            | Reflector, also blocked             |
|       116 | `.Update()` → needs lock                            | Reflector, blocked too              |

This PR makes the AddLeaseEvent and UpdateLeaseEvent non-blocking with async adds with a fixed concurrency using goroutines when exceeding the buffer size. Any events that can not fit into the channel buffer are queued for subsequent processing.

Additionally, when this happens, since wait.PollUntilContextTimeout() results in a timeout, relaxes the fatal condition on main for timeout by ignoring a timeout error to allow for the backend consumer to be setup which results in draining the channel event queue and thus frees up space to process the queued events

Other reference PRs:
https://github.com/flannel-io/flannel/issues/719
https://github.com/flannel-io/flannel/pull/729
https://github.com/flannel-io/flannel/pull/2248

Note that the mechanism implemented in the PR to back off and retry excess events is applicable even outside of the startup to handle back pressure from the backends and filling up of the channel event buffer during runtime.

## Todos
- [x] Tests

Deadlock logs:
------------
```
queue size : 50
cluster size: 196

I0618 21:13:39.194940       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.97.128/25]
I0618 21:13:39.194947       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.127.0/25]
I0618 21:13:39.194954       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.105.128/25]
I0618 21:13:39.194963       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.130.128/25]
I0618 21:13:39.194969       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.124.128/25]
I0618 21:13:39.194977       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.126.128/25]
I0618 21:13:39.194985       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.15.0/25]
I0618 21:13:39.194992       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.125.0/25]
I0618 21:13:39.194998       1 kube.go:511] Creating the node lease for IPv4. This is the n.Spec.PodCIDRs: [100.96.125.128/25]
I0618 21:18:39.195302       1 reflector.go:281] github.com/flannel-io/flannel/pkg/subnet/kube/kube.go:491: forcing resync
```

Logs with Fix:
```
queue size : 50
cluster size: 196
E0619 00:39:09.459563       1 kube.go:157] Node controller sync not completed yet context deadline exceeded
E0619 00:39:09.459617       1 main.go:229] Timed out waiting for node controller sync. Continuing anyway.
I0619 00:39:09.459623       1 main.go:236] Created subnet manager: Kubernetes Subnet Manager - ......
I0619 00:39:09.459627       1 main.go:239] Installing signal handlers
I0619 00:39:09.459689       1 main.go:484] Found network config - Backend type: vxlan
I0619 00:39:09.462607       1 kube.go:714] List of node annotations: .....
I0619 00:39:09.462642       1 match.go:211] Determining IP address of default interface
I0619 00:39:09.463503       1 match.go:264] Using interface with name bond0 and address 10.116.9.204
I0619 00:39:09.463518       1 match.go:286] Defaulting external address to interface address (10.116.9.204)
I0619 00:39:09.463555       1 vxlan.go:141] VXLAN config: VNI=1 Port=0 GBP=false Learning=false DirectRouting=false
I0619 00:39:09.465517       1 kube.go:681] List of node annotations: ....
I0619 00:39:09.465546       1 vxlan.go:155] Interface flannel.1 mac address set to: 8e:51:89:39:41:0d

I0619 00:37:48.697825       1 kube.go:250] Channel buffer full, add event asynchronously
E0619 00:39:48.602228       1 kube.go:157] Node controller sync not completed yet context deadline exceeded
E0619 00:39:48.602274       1 main.go:229] Timed out waiting for node controller sync. Continuing anyway.
I0619 00:39:48.602281       1 main.go:236] Created subnet manager: Kubernetes Subnet Manager - lor1-0005164.int.linkedin.com
I0619 00:39:48.602284       1 main.go:239] Installing signal handlers
I0619 00:39:48.602357       1 main.go:484] Found network config - Backend type: vxlan
I0619 00:39:48.605857       1 kube.go:714] List of node annotations: ..........
I0619 00:39:48.605909       1 match.go:211] Determining IP address of default interface
I0619 00:39:48.606908       1 match.go:264] Using interface with name bond0 and address 10.190.231.115
I0619 00:39:48.606924       1 match.go:286] Defaulting external address to interface address (10.190.231.115)


```

### DIFF
--- a/main.go
+++ b/main.go
@@ -225,7 +225,8 @@ func main() {
 
 	sm, err := newSubnetManager(ctx)
 	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
+		contCacheNotReady := os.Getenv("CONT_WHEN_CACHE_NOT_READY")
+		if contCacheNotReady == "true" && errors.Is(err, context.DeadlineExceeded) {
 			log.Error("Timed out waiting for node controller sync. Continuing anyway.")
 			// Don't exit — continue with startup
 		} else {
--- a/main.go
+++ b/main.go
@@ -225,8 +225,13 @@ func main() {
 
 	sm, err := newSubnetManager(ctx)
 	if err != nil {
-		log.Error("Failed to create SubnetManager: ", err)
-		os.Exit(1)
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Error("Timed out waiting for node controller sync. Continuing anyway.")
+			// Don't exit — continue with startup
+		} else {
+			log.Error("Failed to create SubnetManager: ", err)
+			os.Exit(1)
+		}
 	}
 	log.Infof("Created subnet manager: %s", sm.Name())
 
--- a/pkg/subnet/kube/kube.go
+++ b/pkg/subnet/kube/kube.go
@@ -29,6 +29,7 @@ import (
 	"github.com/flannel-io/flannel/pkg/ip"
 	"github.com/flannel-io/flannel/pkg/lease"
 	"github.com/flannel-io/flannel/pkg/subnet"
+	"golang.org/x/sync/semaphore"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -72,7 +73,7 @@ type kubeSubnetManager struct {
 	nodeController            cache.Controller
 	subnetConf                *subnet.Config
 	events                    chan lease.Event
-	maxAsyncSend              chan struct{}
+	asyncSendSemaphore        *semaphore.Weighted
 	clusterCIDRController     cache.Controller
 	setNodeNetworkUnavailable bool
 	disableNodeInformer       bool
@@ -194,7 +195,7 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
 		}
 	}
 	ksm.events = make(chan lease.Event, scale)
-	ksm.maxAsyncSend = make(chan struct{}, 100)
+	ksm.asyncSendSemaphore = semaphore.NewWeighted(100)
 	// when backend type is alloc, someone else (e.g. cloud-controller-managers) is taking care of the routing, thus we do not need informer
 	// See https://github.com/flannel-io/flannel/issues/1617
 	if sc.BackendType == "alloc" {
@@ -255,10 +256,14 @@ func (ksm *kubeSubnetManager) enqueueLeaseEvent(evt lease.Event, nodeName string
 	}
 
 	// Instead of select with default, *block* until a slot is free
-	ksm.maxAsyncSend <- struct{}{} // acquire a slot, block if none available
+	// Use a context with no timeout to block until a slot is available
+	if err := ksm.asyncSendSemaphore.Acquire(context.Background(), 1); err != nil {
+		log.Errorf("error in acquiring semaphore for async event send, dropping event: %v", err)
+		return
+	}
 
 	go func() {
-		defer func() { <-ksm.maxAsyncSend }() // release slot when done
+		defer func() { ksm.asyncSendSemaphore.Release(1) }() // release slot when done
 
 		ticker := time.NewTicker(100 * time.Millisecond)
 		defer ticker.Stop()
--- a/pkg/subnet/kube/kube.go
+++ b/pkg/subnet/kube/kube.go
@@ -154,7 +154,7 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
 		})
 
 		if err != nil {
-			log.Error("Node controller sync not completed yet ", err)
+			log.Errorf("Node controller sync not completed within 1s: %v", err)
 			return sm, fmt.Errorf("error waiting for nodeController to sync state: %w", err)
 		}
 		log.Infof("Node controller sync successful")
--- a/pkg/subnet/kube/kube.go
+++ b/pkg/subnet/kube/kube.go
@@ -215,9 +215,11 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
 			resyncPeriod,
 			cache.ResourceEventHandlerFuncs{
 				AddFunc: func(obj interface{}) {
-					ksm.handleAddLeaseEvent(lease.EventAdded, obj)
+					ksm.handleAddLeaseEvent(ctx, lease.EventAdded, obj)
 				},
-				UpdateFunc: ksm.handleUpdateLeaseEvent,
+				UpdateFunc: func(oldObj, newObj interface{}) {
+					ksm.handleUpdateLeaseEvent(ctx, oldObj, newObj)
+				},
 				DeleteFunc: func(obj interface{}) {
 					_, isNode := obj.(*v1.Node)
 					// We can get DeletedFinalStateUnknown instead of *api.Node here and we need to handle that correctly.
@@ -234,7 +236,7 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
 						}
 						obj = node
 					}
-					ksm.handleAddLeaseEvent(lease.EventRemoved, obj)
+					ksm.handleAddLeaseEvent(ctx, lease.EventRemoved, obj)
 				},
 			},
 			cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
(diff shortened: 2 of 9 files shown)


pkg/subnet/kube/kube.go
```diff
@@ -72,6 +72,7 @@ type kubeSubnetManager struct {
 	nodeController            cache.Controller
 	subnetConf                *subnet.Config
 	events                    chan lease.Event
+	maxAsyncSend            chan struct{}
```
ahmetb, line 75: nit: gofmt

pkg/subnet/kube/kube.go
```diff
@@ -138,10 +139,23 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
 
 		log.Infof("Waiting %s for node controller to sync", nodeControllerSyncTimeout)
 		err = wait.PollUntilContextTimeout(ctx, time.Second, nodeControllerSyncTimeout, true, func(context.Context) (bool, error) {
-			return sm.nodeController.HasSynced(), nil
+			ch := make(chan bool, 1)
```
ahmetb, line 142: Does this need to be a buffered chan? You send from a goroutine right away and recv in the main thread so both ends have an actor, no need to buffer?

pkg/subnet/kube/kube.go
```diff
@@ -138,10 +139,23 @@ func NewSubnetManager(ctx context.Context, apiUrl, kubeconfig, prefix, netConfPa
...
+			case <-ctx.Done():
+				return false, ctx.Err()
+			case <-time.After(time.Second):
+				return false, nil // or error to break
+			}
 		})
+
 		if err != nil {
-			return nil, fmt.Errorf("error waiting for nodeController to sync state: %v", err)
+			log.Error("Node controller sync not completed yet ", err)
```
ahmetb, line 157:
```suggestion
			log.Errorf("Node controller sync not completed within 1s: %v", err)
```

pkg/subnet/kube/kube.go
```diff
@@ -226,6 +241,36 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
...
+	// Try to send immediately
+	select {
+	case ksm.events <- evt:
+		return
+	default:
+		log.Infof("Channel buffer full, add event asynchronously")
+	}
+
+	// Instead of select with default, *block* until a slot is free
+	ksm.maxAsyncSend <- struct{}{} // acquire a slot, block if none available
```
ahmetb, line 254: Could use high level library https://pkg.go.dev/golang.org/x/sync/semaphore instead

sudheerv:
[flannel.stacktrace.hanging.txt](https://github.com/user-attachments/files/20808571/flannel.stacktrace.hanging.txt)

pkg/subnet/kube/kube.go
sudheerv: added comments to explain the reason for using a buffered channel send

pkg/subnet/kube/kube.go
sudheerv: Ack - I was originally trying to avoid adding new ext library dependencies, but makes sense in this case for readability. Updated.

main.go
```diff
@@ -225,8 +225,13 @@ func main() {
 
 	sm, err := newSubnetManager(ctx)
 	if err != nil {
-		log.Error("Failed to create SubnetManager: ", err)
-		os.Exit(1)
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Error("Timed out waiting for node controller sync. Continuing anyway.")
```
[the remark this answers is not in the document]
sudheerv: Yes, we will still need this because the backend consumers are only started after subnetmanager/cache is initialized. If we fail the initialization, consumers are never started and the channel queue will remain stuck at max depth (even after subsequent restarts) for large clusters.

I do agree with the concern that this is a "behavior" change, so I added an ENV with the default behavior to be to fail as today (and only large clusters that need this behavior will set it to "true"). 

Let me know what you think.

sudheerv:
@manuelbuil Thank you for the careful review. Updated the PR per your suggestions and made the startup behavior change when cache not ready configurable (defaulting to current behavior of failing startup).

sudheerv:
> Thanks a lot for the detailed description and testing Flannel in such challenging environment! Does this happen while starting up the 5k env? Or have you experienced it even after the cluster was up and running?
> 
> Anyway, I have a couple of comments but in general I think your design is a very good solution for this problem. Thanks for taking the time!


re: `Does this happen while starting up the 5k env? Or have you experienced it even after the cluster was up and running?`

Yeah, we have only seen this during startup for clusters that are larger than 5K nodes. Once the start up is completed, and the backend consumers are setup, they should be draining the channel event queue and the chances for hitting this issue becomes very low, if not entirely non-zero (depending on the backpressure from the backends such as locked iptables). 

PS: independent of the deadlock problem, another factor at play is the order of initialization between the producer and consumers - that is why, I had to add an "escape" switch to let the startup continue even after fixing the deadlock to allow for consumers to be started and not fatal fail the startup.

pkg/subnet/kube/kube.go
```diff
@@ -226,6 +246,48 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
...
+	}
+
+	// Instead of select with default, *block* until a slot is free
+	// Use a context with no timeout to block until a slot is available
+	if err := ksm.asyncSendSemaphore.Acquire(context.Background(), 1); err != nil {
+		log.Errorf("error in acquiring semaphore for async event send, dropping event: %v", err)
+		return
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
sudheerv: Ah yes, Good point - Updated as suggested!

sudheerv:
> Thanks for addressing my comments! One final comment to help with graceful shutdowns in case the main program dies

Thank you! Good catch - Updated as suggested.

pkg/subnet/kube/kube.go
```diff
@@ -226,6 +246,48 @@ func newKubeSubnetManager(ctx context.Context, c clientset.Interface, sc *subnet
...
+		defer func() { ksm.asyncSendSemaphore.Release(1) }() // release slot when done
+
+		backoff := 100 * time.Millsecond
```
thomasferrandiz: This should be time.Millisecond

sudheerv: yeah, it was a typo. fixed it. Thanks!

thomasferrandiz (approved):

thomasferrandiz:
@sudheerv Thanks for the PR.
Could you rebase it to solve the go.mod conflict?
Then we're good to merge it.

sudheerv:
> @sudheerv Thanks for the PR. Could you rebase it to solve the go.mod conflict? Then we're good to merge it.

Thank you @thomasferrandiz . Did a rebase - Could you pls help merge it (not sure if it needs another review/approval stamp)?

Issue #2267, opened 2025-07-25
k8s 1.32: cannot read all nodes, only its own Node object


defo89:
Hello,

Kubernetes version: v1.32.6
Flannel version: v0.27.2

After upgrading k8s apiserver/kubelet to `v1.32.6`, following error is logged:
```shell
2025-07-25T10:19:22.303615971Z I0725 10:19:22.303489       1 main.go:213] CLI flags config: {etcdEndpoints:http://127.0.0.1:4001,http://127.0.0.1:2379 etcdPrefix:/coreos.com/network etcdKeyfile: etcdCertfile: etcdCAFile: etcdUsername: etcdPassword: version:false kubeSubnetMgr:true kubeApiUrl: kubeAnnotationPrefix:flannel.alpha.coreos.com kubeConfigFile:/etc/kubernetes/kubelet/kubeconfig iface:[10.1.2.3] ifaceRegex:[] ipMasq:true ifaceCanReach: subnetFile:/run/flannel/subnet.env publicIP: publicIPv6: subnetLeaseRenewMargin:60 healthzIP:0.0.0.0 healthzPort:30670 iptablesResyncSeconds:5 iptablesForwardRules:true blackholeRoute:false netConfPath:/etc/kube-flannel/net-conf.json setNodeNetworkUnavailable:true}
2025-07-25T10:19:22.305633923Z I0725 10:19:22.305587       1 kube.go:139] Waiting 10m0s for node controller to sync
2025-07-25T10:19:22.305645907Z I0725 10:19:22.305627       1 kube.go:537] Starting kube subnet manager
2025-07-25T10:19:22.312332575Z W0725 10:19:22.312273       1 reflector.go:561] pkg/mod/k8s.io/client-go@v0.31.11/tools/cache/reflector.go:243: failed to list *v1.Node: nodes is forbidden: User "system:node:node4.domain.com" cannot list resource "nodes" in API group "" at the cluster scope: node 'node4.domain.com' cannot read all nodes, only its own Node object
[...]

PR #2270, opened 2025-07-30
fix NodeAuthorization: watch only own node to avoid listing all nodes

defo89:
## Description

Fixes https://github.com/flannel-io/flannel/issues/2267

With `AuthorizeNodeWithSelectors` enabled by default starting from 1.32, the call to list all nodes is denied. Instead, we are now watching own node only. /thanks @Nuckal777

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fixed NodeAuthorization for k8s 1.32: watch only own node
```


thomasferrandiz:
Thanks for the PR but flannel needs to watch all the nodes.

Each instance of flannel watches the other nodes so that it can create iptables rules to route traffic towards the pods running on these nodes.

PR #2438, opened 2026-04-30
fix: reap orphaned zombie child processes when running as PID 1

pratikjagrut:
## Description

**Problem:** flanneld doesn't reap orphaned child processes when running as PID 1 in a container. The iptables-wrapper v2 shell script's nft/legacy auto-detection spawns a `timeout` command whose internal helper process gets orphaned and reparented to flanneld. Since Go's runtime only reaps processes it directly started, this orphan becomes a permanent zombie (`[timeout] <defunct>`).

This results in exactly one zombie process per node that:
- Persists for the lifetime of the container
- Reappears after DaemonSet restart
- Cannot be killed (it's already dead, just not reaped)
- Gets flagged by security audits and monitoring tools

**Fix:** Update iptables-wrapper from v2 to v3. The v3 wrapper is rewritten as a Go binary that uses `exec` instead of a shell script with `timeout`, so no child process is forked and no zombie is created.

**Components affected:**
- `images/Dockerfile` — updates iptables-wrapper to v3 (`bfef9e5`), removes `iptables-wrapper-installer.sh`, uses new install command

**Testing:**
- Manual: deployed patched flanneld to RKE2 cluster, verified `ps aux | grep defunct` returns no results
- Manual: restarted Flannel DaemonSet, confirmed no zombie reappears
- Build: `go build` and `go vet` pass cleanly

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

Update iptables-wrapper from v2 to v3 to fix zombie child process (`[timeout] <defunct>`) created during nft/legacy auto-detection when flanneld runs as PID 1.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -28,7 +28,7 @@ RUN export GOOS=$(xx-info os) &&\
     make dist/flanneld
 RUN git clone https://github.com/kubernetes-sigs/iptables-wrappers.git /iptables-wrapper
 WORKDIR /iptables-wrapper
-RUN git checkout 5792812d9e5a5bb7f22d79d557bbfeece253343d
+RUN git checkout bfef9e5087a198b50a4124bb9ce9d2c7c99025dd
 RUN export GOOS=$(xx-info os) &&\
     export GOARCH=$(xx-info arch) &&\
     export ARCH=$(xx-info arch) &&\
@@ -40,12 +40,11 @@ RUN apk add --no-cache iproute2 ca-certificates nftables iptables strongswan ipt
 RUN apk add wireguard-tools --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/community
 COPY --from=flannel-builder /build/dist/flanneld /opt/bin/flanneld
 COPY dist/mk-docker-opts.sh /opt/bin/
-COPY --from=flannel-builder /iptables-wrapper/iptables-wrapper-installer.sh /
-COPY --from=flannel-builder /iptables-wrapper/bin/iptables-wrapper /
+COPY --from=flannel-builder /iptables-wrapper/bin/iptables-wrapper /usr/sbin
 # check manually that iptables-legacy and iptables-nft are present since
 # iptables-wrapper-installer.sh sanity check doesn't work for multi-arch build
 RUN which iptables-legacy && which iptables-nft
-RUN /iptables-wrapper-installer.sh --no-sanity-check
+RUN /usr/sbin/iptables-wrapper install
 
 
 ENTRYPOINT ["/opt/bin/flanneld"]


thomasferrandiz (dismissed):

pratikjagrut:
@thomasferrandiz Thank you for approving the PR but I just squashed the commits so can you please run the CI and re-approve the PR if needed.

manuelbuil:
It feels to me that the real problem is in the "iptables-wrapper's nft/legacy auto-detection" script rather than in flannel. Have you investigated if it is possible to fix it there? I am also wondering if that script is needed at all now that all OS use nft iptables

main.go
```diff
@@ -264,6 +264,7 @@ func main() {
 		wg.Done()
 	}()
 
+
```
rbrtbnfgl: Could you clean the new line here?

pratikjagrut: Removed.

rbrtbnfgl:
I'm agree with Manuel but I don't know if we can remove the `iptables-wrapper` for now.

rbrtbnfgl:
anyway I did a quick test updating the used iptables-wrapper from v2 to v3 and it seems that they updated the script execution with `exec` from my local test it seems that the orphaned child is not created anymore so it's enough to update the Dockerfile.
```diff --git a/images/Dockerfile b/images/Dockerfile
index 393781a5..015b39d0 100644
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -28,7 +28,7 @@ RUN export GOOS=$(xx-info os) &&\
     make dist/flanneld
 RUN git clone https://github.com/kubernetes-sigs/iptables-wrappers.git /iptables-wrapper
 WORKDIR /iptables-wrapper
-RUN git checkout 5792812d9e5a5bb7f22d79d557bbfeece253343d
+RUN git checkout bfef9e5087a198b50a4124bb9ce9d2c7c99025dd
 RUN export GOOS=$(xx-info os) &&\
     export GOARCH=$(xx-info arch) &&\
     export ARCH=$(xx-info arch) &&\
@@ -40,12 +40,11 @@ RUN apk add --no-cache iproute2 ca-certificates nftables iptables strongswan ipt
 RUN apk add wireguard-tools --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/community
 COPY --from=flannel-builder /build/dist/flanneld /opt/bin/flanneld
 COPY dist/mk-docker-opts.sh /opt/bin/
-COPY --from=flannel-builder /iptables-wrapper/iptables-wrapper-installer.sh /
 COPY --from=flannel-builder /iptables-wrapper/bin/iptables-wrapper /
 # check manually that iptables-legacy and iptables-nft are present since
 # iptables-wrapper-installer.sh sanity check doesn't work for multi-arch build
 RUN which iptables-legacy && which iptables-nft
-RUN /iptables-wrapper-installer.sh --no-sanity-check
+RUN /iptables-wrapper install
 
 
 ENTRYPOINT ["/opt/bin/flanneld"]```

**Comment by pratikjagrut** [comment]
@rbrtbnfgl Thanks for the suggestion. I updated the dockerfile to use iptables-wrapper v3. It's working fine. PTAL.

**Review by rbrtbnfgl — APPROVED** [review]

File: images/Dockerfile [hunk 2]
```diff
@@ -40,12 +40,11 @@ RUN apk add --no-cache iproute2 ca-certificates nftables iptables strongswan ipt
 RUN apk add wireguard-tools --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/community
 COPY --from=flannel-builder /build/dist/flanneld /opt/bin/flanneld
 COPY dist/mk-docker-opts.sh /opt/bin/
-COPY --from=flannel-builder /iptables-wrapper/iptables-wrapper-installer.sh /
 COPY --from=flannel-builder /iptables-wrapper/bin/iptables-wrapper /
```
rbrtbnfgl:
```suggestion
COPY --from=flannel-builder /iptables-wrapper/bin/iptables-wrapper /usr/sbin
```

images/Dockerfile
```diff
@@ -40,12 +40,11 @@ RUN apk add --no-cache iproute2 ca-certificates nftables iptables strongswan ipt
...
 # check manually that iptables-legacy and iptables-nft are present since
 # iptables-wrapper-installer.sh sanity check doesn't work for multi-arch build
 RUN which iptables-legacy && which iptables-nft
-RUN /iptables-wrapper-installer.sh --no-sanity-check
+RUN /iptables-wrapper install
```
rbrtbnfgl:
```suggestion
RUN /usr/sbin/iptables-wrapper install
```

rbrtbnfgl (changes requested):
These fixes should solve the test issue

images/Dockerfile
pratikjagrut: Done. Thanks.

images/Dockerfile
pratikjagrut: Done. Thanks.

rbrtbnfgl (approved):

PR #2466, opened 2026-06-26
feat: new install_conf cmd to install flannel's config file

thomasferrandiz:
## Description
The PR adds a static go binary that can be used to  copy files atomically.
In this way we won't need cp for the init container and we can use a minimal container
esp. downstream where we use k3s-root to get the iptables binary.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/cmd/install_conf/main.go
+++ b/cmd/install_conf/main.go
@@ -1,0 +1,41 @@
+// Copyright (c) 2026 Flannel Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


rbrtbnfgl (dismissed):

cmd/install_conf/main.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	src := os.Args[1]
+	dest := os.Args[2]
+	destDir := filepath.Dir(dest)
+	destFileName := filepath.Base(dest)
+
+	if err := utils.CopyFileAtomic(src, destDir, "."+destFileName+".tmp", destFileName); err != nil {
+		fmt.Fprintf(os.Stderr, "error installing %q: %v\n", src, err)
+		os.Exit(1)
```
manuelbuil, line 39: We were wondering if there might be a well known library which implements this function. We could import it instead of reinventing the wheel. But it does not seem to be any reliable and maintained one

manuelbuil (approved):

PR #2513, opened 2026-07-27
Migrate kind e2e tests from bash_unit to Ginkgo

thomasferrandiz:
## Description

This is a test-infrastructure change (no product code). It replaces the bash_unit-driven kind e2e suite (`e2e/run-e2e-tests.sh` and its helpers) with a native Ginkgo v2 + Gomega suite written in Go.

**Why:** the previous suite was a large, hard-to-maintain bash script that shelled out to `kubectl`, `kind`, `yq` and `jq`, with ad hoc polling and string assertions. Moving it to Go lets us drive the cluster and Kubernetes through first-class libraries, reuse the module toolchain, and get structured, readable specs and failure output.

**Approach:**
- New `e2e/` Go package gated behind the `e2e` build tag, so it never runs during `make unit-test`.
- Cluster lifecycle uses the `sigs.k8s.io/kind` library (create, load images via `docker save` -> node import, delete, and node exec for rule inspection). `kind` is now a library dependency rather than a CLI tool.
- Kubernetes operations use `client-go` (pod create, IPs, podCIDR, `remotecommand` exec, logs) and all readiness waits (DaemonSet rollout, nodes, coredns, pods, ping reachability, and `/run/flannel/subnet.env` on each node).
- The flannel manifest is rendered from `Documentation/kube-flannel.yml` as a template (image, `net-conf.json`, udp privileged flag) and applied/deleted via the dynamic client, replacing the yq + kubectl flow.
- Ports the full prior coverage: per-backend connectivity ping tests, iperf3 perf tests, iptables/nftables golden-rule assertions, and the on-failure debug dump.

**Scope note:** only the kind suite is migrated. The etcd/docker `dist/functional-test*.sh` suites intentionally stay on bash_unit.

**Components affected:** e2e test suite, `make kind-e2e-test`, and the `kind-e2eTests` GitHub Actions workflow (drops the kind CLI install since the library handles clusters now). Removes 5 obsolete bash scripts and keeps `e2e/kind-config.yaml`.

**Testing:** `go vet -tags e2e ./e2e/` is clean, the test binary compiles, `gofmt` is clean, and `go mod tidy` is stable. A `--ginkgo.dry-run` enumerates 11 specs (6 connectivity + 5 perf), matching the original coverage (udp and its perf variant remain amd64-only). The full run exercises real kind clusters and is covered by the `kind-e2eTests` workflow on CI.

## Todos
- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


manuelbuil:
Could we do a follow-up to separate framework/helpers from specs for maintainability? Or maybe in this PR itself? The same directory includes helper-functions (e.g. in cluster.go) and the tests themselves. I would prefer having two different subdirectories: one for infrahelpers and one for tests

thomasferrandiz:
I agree the separation can help once the suite grows, but in Go the current boundary is already the idiomatic one: helpers live in non-_test.go files and specs in _test.go files within the same package. Moving them into subdirectories here would mean a separate package and a fairly noisy export/refactor of the helper surface, so I would prefer to keep this PR as-is and revisit that as a follow-up if the e2e suite grows enough to justify it.

rbrtbnfgl (approved):

Issue #2470, opened 2026-06-30
etcd subnet watch never recovers from compaction (hot-loops, fills disk)


phinze:
## Expected Behavior

When the etcd subnet-lease watch falls behind etcd's compaction horizon (`mvcc: required revision has been compacted`), flanneld should recover on its own: re-list current leases at a fresh revision and resume watching from there, the same catch-up it does at startup. A transient compaction shouldn't require a restart.

## Current Behavior

Once the watch's start revision has been compacted away, `etcdSubnetRegistry.watchSubnets` reconnects at the *same* revision indefinitely. The revision is never advanced, so every reconnect fails immediately with the same compaction error and the watch re-establishes in a tight loop, emitting a WARN+INFO pair per cycle (backoff caps at 5s). It never recovers without a restart. Captured loop (note the revision is always the original `rev 1`):

```
registry.go:293] registry: watching subnets starting from rev 1
registry.go:314] etcd watch channel for /coreos.com/network/subnets closed with error etcdserver: mvcc: required revision has been compacted, reconnecting...
registry.go:293] registry: watching subnets starting from rev 1
registry.go:314] etcd watch channel for /coreos.com/network/subnets closed with error etcdserver: mvcc: required revision has been compacted, reconnecting...
```

Root cause: `watchSubnets` (and `watchSubnet`) take a `since` revision and never update it across reconnects. Recovery logic does exist (`leasesWatchReset`, via the `isIndexTooSmall` case), but it lives inside the per-event parse switch, which a compaction never reaches: compaction arrives as a watch-response error (`wresp.Err()` / `CompactRevision`), handled by the earlier channel-closed branch that simply reconnects at the stale `since`.

## Possible Solution

In `watchSubnets`/`watchSubnet`:

1. On a watch-response error, detect compaction (`CompactRevision != 0` / `rpctypes.ErrCompacted`) and recover by re-listing (`leasesWatchReset`) at a current revision and resuming from there, instead of reconnecting at the stale revision.
2. Advance the resume revision as events arrive (`since = resp.Header.Revision + 1`) so an ordinary reconnect resumes forward rather than replaying from the original start rev (which is what lets it drift below the compaction horizon in the first place).

I have a patch plus a regression test and will open a PR referencing this issue.

## Steps to Reproduce (for bugs)

1. Add the test below to `pkg/subnet/etcd` (it uses the existing `integration` harness).
[...]

PR #2524, opened 2026-08-05
subnet/etcd: recover single-subnet watch from compaction

phinze:
## Description

Follow-up to #2471, which fixed the compaction hot-loop in `watchSubnets` but
deliberately left the single-subnet `watchSubnet` alone pending a design call on
the recovery semantics (see my comment on #2470). No preference came back, and
we've now hit the bug in production ourselves, so this proposes an answer.

`watchSubnet` has the identical defect: it takes a `since` revision that is never
advanced, so once etcd compacts past it every reconnect fails with `mvcc:
required revision has been compacted` at the 5s backoff cap indefinitely. The
watch never delivers another event and only a restart clears it.

The open question was what recovery should do when the re-read finds the lease
gone: resume, or terminate? This picks resume, which keeps the behaviour the
watch already has after an ordinary delete. The one addition is that recovery
synthesizes the `EventRemoved` the missed delete would have produced, inferred
from the lease's absence, since the delete event itself sits below the
compaction horizon and can't be recovered from the watch stream. Synthesizing
it rather than acting on it keeps the revoke policy with the caller, where
`CompleteLease` already implements it as "lease revoked, shut down"; deciding
that inside the registry would bake one consumer's policy into the shared layer.

Also advances the resume revision as events arrive, matching #2471, so an
ordinary reconnect moves forward rather than replaying from the original start
rev, which is what lets it drift below the horizon to begin with. The reconnect
warnings now include the revision being resumed from; not having that is a large
part of why this took a while to identify in the field.

Worth flagging for a maintainer's call: `watchSubnet` and `watchSubnets` are now
~76% identical line-for-line, and this bug existing at all is a direct
consequence of that. I've kept them separate here to keep the fix reviewable and
to avoid churning the working plural path, but I'm happy to send a follow-up
unifying the reconnect/backoff scaffolding behind one driver if you'd prefer that
over two copies drifting again.

Testing: three new tests, all using the existing `integration` harness.
`TestWatchSubnetRecoversFromCompaction` and
`TestWatchSubnetReportsRemovalAfterCompaction` cover the two recovery outcomes,
lease still present and lease gone. `TestWatchSubnetSurvivesCompactionAfterReconnect`
reproduces the sequence actually seen in the field: a watch that is established
and healthy, drifts below the compaction horizon while still connected, and only
wedges when etcd is bounced underneath it. That drift is why this goes unnoticed
for so long, and it is the part the other two skip by starting already
compacted.

Worth knowing for review: recovery surfaces a re-read snapshot rather than
replaying the events missed while disconnected, since those events are below the
horizon and no longer exist. The third test pins that ordering explicitly, and a
change made before recovery lands is simply absorbed into the snapshot.

Verified all three fail on master and pass with the fix; the third reproduces the
hot-loop with its backoff ramp to the 5s cap, and was run repeatedly to check it
isn't timing-sensitive. Full `pkg/subnet/etcd` suite passes (17s, up from 14s);
gofmt, `go vet`, and golangci-lint v2.7.2 (the version CI pins) are all clean.
The patched build has not yet run on our own hardware.

Components: `pkg/subnet/etcd`. Affects the etcd subnet manager's per-lease watch,
which every backend reaches via `CompleteLease`; the kube subnet manager is
unaffected.

## Todos
- [x] Tests
- [ ] Documentation (n/a)
- [x] Release note

## Release Note
```release-note
Fix the etcd subnet manager's single-subnet watch hot-looping forever after an etcd compaction. The watch now re-reads the lease at a current revision and resumes, and reports the lease as removed if it was deleted while the watch was disconnected.
```

### DIFF
--- a/pkg/subnet/etcd/registry.go
+++ b/pkg/subnet/etcd/registry.go
@@ -431,17 +431,39 @@ func (esr *etcdSubnetRegistry) watchSubnet(ctx context.Context, since int64, sn 
 			case wresp, ok := <-rch:
 				err := wresp.Err()
 				if !ok || err != nil {
-					if err != nil {
-						log.Warningf("etcd watch channel for %s closed with error %v, reconnecting...", key, err)
-					} else {
-						log.Warningf("etcd watch channel for %s closed, reconnecting...", key)
-					}
 					cancel()
+					// If the watch fell behind etcd's compaction horizon, reconnecting
+					// at the same revision fails identically forever (mvcc: required
+					// revision has been compacted). Re-read the lease at a current
+					// revision and resume from there instead of hot-looping.
+					if isCompacted(wresp) {
+						log.Warningf("etcd watch for %s fell behind compaction horizon (compact rev %d), re-reading and resuming", key, wresp.CompactRevision)
+						next, rerr := esr.resyncWatchSubnet(ctx, sn, sn6, leaseWatchChan)
+						if rerr != nil {
+							log.Errorf("failed to re-read subnet lease after compaction: %v", rerr)
+							time.Sleep(exponentialBackoff)
+							exponentialBackoff = min(exponentialBackoff*2, maxBackoff)
+							break innerLoop
+						}
+						since = next
+						exponentialBackoff = initialBackoff
+						break innerLoop
+					}
+					if err != nil {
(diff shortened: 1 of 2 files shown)


thomasferrandiz (approved):

thomasferrandiz:
Thanks for the fix!
If you have time to do a follow-up and unify `watchSubnet` and `watchSubnets` that would be great.

PR #2560, opened 2026-09-11
vxlan: recover from closed netlink subscription

### LINKED ISSUE #2549: bug: watchVXLANDevice panics (nil pointer) and CPU-spins when netlink subscription closes on ENOBUFS
### Affected version
flannel v0.27.4+ (introduced by commit 73af8ea / PR #2272)

### Description
`watchVXLANDevice` has two bugs triggered when the netlink socket buffer overflows (ENOBUFS):

**Bug 1 — nil pointer panic (k3s/flannel crashes)**

The internal goroutine spawned by `netlink.LinkSubscribe` has `defer close(ch)` and returns on any `s.Receive()` error, including `ENOBUFS`. When the channel closes, `watchVXLANDevice`'s select receives a zero-value `LinkUpdate{Link: nil}` without detecting the close. Calling `update.Attrs()` dispatches through a nil `Link` interface → SIGSEGV.

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18]
goroutine N [running]:
github.com/flannel-io/flannel/pkg/backend/vxlan.(*network).watchVXLANDevice(...)
pkg/backend/vxlan/vxlan_network.go:138

**Bug 2 — 100% CPU spin (even without the crash)**

If the nil is caught but the code does `continue`, a closed channel is always immediately ready in a Go `select` — the goroutine busy-loops forever at full CPU.

### Root cause in current code

```go
// pkg/backend/vxlan/vxlan_network.go
case update := <-updates:          // no ok-pattern — can't detect channel close
    if update.Attrs() == nil {     // update.Link is nil → Attrs() panics
        continue                   // if nil guarded: closed channel fires every iteration → CPU spin
    }

### PR DESCRIPTION
**Pull request by dlanov**
## Description

Bug fix for #2549.

`watchVXLANDevice` currently assumes the netlink update channel remains open. When `netlink.LinkSubscribe` exits after a receive error such as `ENOBUFS`, the channel is closed. Reading from the closed channel produces a zero-value `LinkUpdate`, which can panic when `Attrs()` is called. Continuing to read the closed channel would also cause a CPU spin.

This change:

* detects a closed netlink update channel and safely handles nil `Link` updates;
* cleans up a failed subscription and re-establishes it after a context-aware retry delay;
* preserves the existing fatal behavior for the initial subscription failure;
* preserves the existing `RTM_DELLINK` behavior;
* keeps VXLAN monitoring active across recoverable subscription failures.

Automated testing:

* `go test ./pkg/backend/vxlan -count=1`
* `go test -race ./pkg/backend/vxlan -count=1`
* `go vet ./pkg/backend/vxlan`
* `git diff --check`

Affected component: VXLAN backend.

Fixes #2549

## Todos

* [x] Tests
* [ ] Documentation
* [ ] Release note

## Release Note

```release-note
None required
```


thomasferrandiz (approved):

manuelbuil (approved):
@dlanov were you able to reproduce this issue?

dlanov:
I was able to reproduce the failure path in a unit test by simulating the netlink subscription channel being closed, which triggers the same behavior as the reported receive failure. I have not reproduced the original `ENOBUFS` condition itself on a live system, but the regression tests cover the closed-channel case and verify that the watcher recovers and resubscribes instead of panicking or spinning.

manuelbuil:
> I was able to reproduce the failure path in a unit test by simulating the netlink subscription channel being closed, which triggers the same behavior as the reported receive failure. I have not reproduced the original `ENOBUFS` condition itself on a live system, but the regression tests cover the closed-channel case and verify that the watcher recovers and resubscribes instead of panicking or spinning.

Thanks. It feels like the reporter of the issue found the bugs but was also not able to reproduce them in a live system. I am trying to understand how likely it is to find such ENOBUF situation but I'd say it's not that probable (fortunately!)

